PR #31, opened 2015-06-23
III-316: Add unavailable and past search properties and default them …

bramcordie:
…to true


cyberwolf:
Functional requirements have changed, this won't make it in.

PR #38, opened 2015-07-29
Backfill migrations to create clean migration path

willaerk:
In order to create a clean migration path from an empty database
to the latest schema version, a number of missing migrations have
been added here. This makes the database population through migrations
equivalent to the method via the DatabaseSchemaInstaller class.

Database migrations have been numbered starting from epoch (01/01/1970)
for lack of availability of a more accurate timestamp. This will
ensure they're processed first.

### DIFF
--- a/src/Migrations/Version19700101000000.php
+++ b/src/Migrations/Version19700101000000.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace CultuurNet\UDB3\Silex\Migrations;
+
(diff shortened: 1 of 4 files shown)


cyberwolf:
Looks good!

I compared a schema after a clean install with a schema initialised by running the migrations with mysqldbcompare, and only the 2 already known issues showed up as differences:

- the doctrine_migration_versions table itself is missing in a clean install
- the event_variation_search_index table is missing in a clean install

PR #152, opened 2016-08-04
III-1267: Use updated udb3-jwt branch

larrybolt:


bramcordie:
After merging the changes to udb3-jwt and updating the dependencies of the silex app this change is unnecessary.

PR #188, opened 2016-10-27
III-1574 Minimum label length is now 2 characters.

LucWollants:


bramcordie:
Already updated the service with the latest dependencies!

PR #229, opened 2017-02-28
III-1916 Added feature toggle for history.

LucWollants:


bertramakers:
Closing because it was decided not to go this way after all (see comments in ticket).

PR #300, opened 2018-09-12
Feature/iii 2738

stijnswaanen:


web/index.php
```diff
@@ -45,6 +45,7 @@ function () {
             ->with(new RequestMatcher('^/contexts/.*', null, 'GET'))
             ->with(new RequestMatcher('^/(event|place|label)/' . $app['id_pattern'] . '$', null, 'GET'))
             ->with(new RequestMatcher('^/(events|places)/' . $app['id_pattern'] . '$', null, 'GET'))
+            ->with(new RequestMatcher('^/(events|places)/' . $app['id_pattern'] . '/calSum+$', null, 'GET'))
```
cyberwolf: The + at the end of the path should be dropped, now you can use calSum or calSumm or calSummmmmmm and these would still match :)

src/Place/PlaceControllerProvider.php
```diff
@@ -62,6 +63,7 @@ function (Application $app) {
         $controllers->put('/{itemId}/images/{mediaObjectId}', 'place_editing_controller:updateImage');
 
         $controllers->get('/', 'place_controller:getByPostalCode');
+        $controllers->get('/{cdbid}/calSum', 'place_controller:getCalendarSummary');
```
cyberwolf: I recommend using calsum with all lowercase characters.

src/Event/EventControllerProvider.php
```diff
@@ -59,6 +60,8 @@ function (Application $app) {
         $controllers->delete('/{itemId}/images/{mediaObjectId}', 'event_editing_controller:removeImage');
         $controllers->put('/{itemId}/images/{mediaObjectId}', 'event_editing_controller:updateImage');
 
+        $controllers->get('/{cdbid}/calSum', 'event_controller:getCalendarSummary');
```
cyberwolf: I recommend using calsum with all lowercase characters.

PR #304, opened 2018-10-19
III-2638: Cli command to validate place JSON-LD for import API

cyberwolf:


bertramakers:
Closing because it's a duplicate of the newer #328 (did not see this one before I made the new one).

PR #320, opened 2019-04-05
Remove install command

bertramakers:
~~⚠️  Do not merge until https://github.com/cultuurnet/udb3-silex/pull/319 has been merged AND the setup script(s) in udb3-vagrant have been updated to run the migrations instead of this custom install command.~~

Should be okay to merge if no problems pop up in review.


bertramakers:
@jdrieghe can you take a look at this instead please when you have some time? We recently started using the migrations for new installations, so that's why I've added this PR to remove the old install command. 

The actual schemas that this command uses are in `udb3-php` but I haven't removed those yet because there's some sqlite tests in there that still rely on them. We should also have those use the migrations once we can merge the code relevant to those tests into udb3-silex (or move the migrations to udb3-php if the relevant repository classes are shared among applications but I doubt that). But that's for later.

jdrieghe:
Looks good, none of this seems necessary anymore 👏.
Can you take a quick look at the conflict, @bertramakers ?

PR #327, opened 2019-05-23
Merge the udb3-export library into udb3-silex

bertramakers:
### Changed

- Moved the separate udb3-export library to udb3-silex itself for easier development

----------------------

This library is only used by udb3-silex according to packagist. Moving it to a subdirectory in udb3-silex and having composer symlink it as a dependency is the first step into merging it with the actual code in udb3-silex and removing alltogether it as a composer dependency.


bertramakers:
I've archived https://github.com/cultuurnet/udb3-export now that this has been merged.

PR #329, opened 2019-06-06
III-2943 Remove dashboard and place lookup endpoints and services

bertramakers:
The `/user/offers/` endpoint that is also deleted with this PR was not in use in the front-end app and was just a duplicate of `/dashboard/items`. It was implemented to clean up the endpoint names but never actually used. See https://jira.uitdatabank.be/browse/III-2177


bertramakers:
Marked as `DNM` until we've tested the switch to SAPI3 and removed the feature toggles from the front-end.

PR #332, opened 2019-07-10
Consume NewsArticleAboutEventAdded messages

bertramakers:
~~Still requires actual config and bootstrapping~~.


src/Curators/NewsArticleProcessManager.php
```diff
@@ -0,0 +1,54 @@
...
+            call_user_func([$this, $handlerMethodName], $payload);
+        }
+    }
+
+    private function handleNewsArticleAboutEventAdded(NewsArticleAboutEventAdded $newsArticleAboutEventAdded): void
+    {
+        $this->commandBus->dispatch(
+            new AddLabel(
+                $newsArticleAboutEventAdded->getEventId()->toNative(),
+                new Label('curatoren', false)
```
jdrieghe: Should we make factory methods for these labels? `Label::curators()` or something?

bertramakers: - [ ] Move to `private` constant

tests/Curators/Events/NewsArticleAboutEventAddedJSONDeserializerTest.php
```diff
@@ -0,0 +1,61 @@
...
+use ValueObjects\StringLiteral\StringLiteral;
+
+final class NewsArticleAboutEventAddedJSONDeserializerTest extends TestCase
+{
+    /**
+     * @var NewsArticleAboutEventAddedJSONDeserializer
+     */
+    private $deserializer;
+
+    public function setUp()
```
jdrieghe: Nitpicking here, but I think the setUp method in `TestCase` is `protected`. There's no point in loosening the access level.

PR #335, opened 2019-07-17
III-2973 - set calendarinfo on reports to small view

brampauwelyn:
### Changed
- Changed calendar view on reports to small, instead of large


bertramakers:
Hey @brampauwelyn! Is there a reason the `composer.lock` file was updated with this change? I'm assuming this was an accident. Can you revert this please? Thanks!

bertramakers:
@brampauwelyn thanks for taking the time to revert the `composer.lock` file. I noticed in the build output that some tests are failing now, but for some reason the build is still green. Can you fix these tests (should be pretty straightforward) while I check why the build was green with failing tests?

bertramakers:
> while I check why the build was green with failing tests?

Should be fixed in this PR: https://github.com/cultuurnet/udb3-silex/pull/343

jdrieghe:
Since it was fairly trivial, I quickly fixed the test here. Good to merge now 👍

PR #337, opened 2019-07-24
Fix permission check when dispatching commands

bertramakers:
### Fixed

- Permissions are now checked again when dispatching a command

### Changed

- Unauthorized commands now return a `401` status code instead of a `400`


bertramakers:
One of the 2 Travis checks was green, the other was stuck on "booting virtual machine".

PR #340, opened 2019-07-25
Add UiTPAS label to event when organizer has UiTPAS label with different casing

bertramakers:


lib/udb3-uitpas-bridge/src/Event/EventProcessManager.php
```diff
@@ -166,7 +166,15 @@ function (Label $label) {
      */
     private function addIntersectingLabelsToEvent($eventId, $labels1, $labels2, $visible)
     {
-        $matchingLabelsAsStrings = array_intersect($labels1, $labels2);
+        $matchingLabelsAsStrings = [];
+        foreach ($labels1 as $label1) {
+            foreach ($labels2 as $label2) {
+                if (strcmp(mb_strtoupper($label1), mb_strtoupper($label2)) === 0) {
```
jdrieghe:
```suggestion
                if (strcasecmp($label1, $label2) === 0) {
```

jdrieghe: Would this be enough? Or do we need the `mb_strtoupper`?

bertramakers: I just checked and apparently we can't use multi-byte characters in labels right now (via the API). Is there a downside to using it though (to future-proof it)?

jdrieghe: The only downside is that the comparison is slightly more complicated to understand. Maybe it would be even better if the labels here were passed as value objects so the comparison could be encapsulated and unit tested there.

bertramakers: Ah true. They were passed as strings because we used `array_intersect` before, but that is always case-sensitive. Since we have to manually check each label now, we could use value objects indeed. Feel free to make that change if you have some time today. Otherwise we can pick it up tomorrow.

bertramakers: Note that there's already a `CultuurNet\UDB3\Label` value object (also used in this file), but it's located in `cultuurnet/udb3-php` so if you need to make changes to it you'll need to create another PR in that library first.

jdrieghe: Will do

jdrieghe: No need. The `Label` VO already has an `equals` method that handles multi-byte safe, case-insensitive string comparison

PR #338, opened 2019-07-25
Move and internalize udb3-uitpas-bridge to udb3-silex

bertramakers:
Do not merge this PR until the PR(s) for the following tickets have also been merged into it:
- [x] https://jira.uitdatabank.be/browse/III-2668 => #340 
- [x] https://jira.uitdatabank.be/browse/III-2753 => #353 

---
~Merge https://github.com/cultuurnet/udb3-silex/pull/348 first and change base to `master`~

PR #342, opened 2019-08-01
[DNM] Add single-use packages

bertramakers:
### Added

- cultuurnet/udb3-udb2-bridge package
- cultuurnet/udb3-models-import package
- cultuurnet/udb3-symfony package (controllers)

------------------------------

As with #338 , this should not be merged until @willaerk is available to fix the deploy process, because these composer dependencies will become a symlink and this requires a one-time manual fix. Ideally it's merged / deployed at the same time as #338 .


bertramakers:
I considered renaming udb3-symfony to udb3-http for clarity, but the namespace would still be `Cultuurnet\UDB3\Symfony` which would be even more confusing so I left it as-is. Will be no issue anymore once we move to Slim or something else and use PSR-7 compatible controllers, as these ones from udb3-symfony will become obsolete then.

PR #344, opened 2019-08-07
[III-2944] Internalize udb3-export lib

jdrieghe:
### Added
- Added all the code from `udb3-export` under `src/EventExport`

### Changed
- Composer dependencies are now always sorted alphabetically
- Curators code is now autoloaded through `src/` directory and PSR-4 instead of through the `Silex` subnamespace

### Removed
- Composer dependency on `udb3-export` lib through a symbolic link
- Code style requirement to suffix interfaces with `Interface`

### Fixed
- Code styles in newly copied library code

---
Ticket: https://jira.uitdatabank.be/browse/III-2944

---
This is a first refactor to internalize the "libraries" we imported from single use packages. This will make it easier to internalize a service from `udb3-php` afterwards. We need that service to solve the validation issue that came up while testing `III-2944`.


src/Curators/Events/NewsArticleAboutEventAddedJSONDeserializer.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace CultuurNet\UDB3\Silex\Curators\Events;
+namespace CultuurNet\UDB3\Curators\Events;
```
bertramakers: The usage of this class (and others changed in this commits) should also be updated in https://github.com/cultuurnet/udb3-silex/blob/master/app/Curators/CuratorsServiceProvider.php

src/EventExport/Format/HTML/HTMLEventFormatter.php
```diff
@@ -230,7 +230,9 @@ private function formatTaaliconen($event, &$formattedEvent)
 
         foreach ($this->taalicoonSpecs as $name => $spec) {
             $i++;
-            /** @var EventSpecificationInterface $spec */
+            /**
+ * @var EventSpecificationInterface $spec 
+*/
```
bertramakers: This seems incorrectly formatted?

src/EventExport/Format/HTML/TransformingIteratorIterator.php
```diff
@@ -13,7 +13,9 @@
  */
 class TransformingIteratorIterator extends IteratorIterator
 {
-    /** @var callable */
+    /**
+ * @var callable 
+*/
```
bertramakers: Again here. Seems like phpcbf has an issue with docblocks in classes?

bertramakers: There's more, but not going to point each one out. 😉

PR #346, opened 2019-08-07
[III-2944] Validate location id on event creation

jdrieghe:
### Fixed
- Creating an event now correctly checks that the location exists. It will return a `400 BAD REQUEST` when the location is invalid.

---
Ticket: https://jira.uitdatabank.be/browse/III-2944

---
⚠️ Merge https://github.com/cultuurnet/udb3-silex/pull/345 first and change base to `master`


bertramakers:
I noticed you've copied the service and updated the controller to use the copy, but the other usages haven't been updated and the original service has not been removed from udb3-php. I'm okay with merging this to get the fix deployed asap so Stan can continue deploying, but can we pick this cleanup up as next priorities so we don't get stuck with even more services? (As well as integrating the remaining libraries.)

jdrieghe:
@bertramakers that was the plan ;)

PR #353, opened 2019-08-09
Apply UiTPAS labels to events based on active card systems

bertramakers:
### Changed

- UiTPAS labels are now applied to events based on the active card systems of the event, and no longer based on the UiTPAS labels on the organizer.
- Removing irrelevant UiTPAS labels is now a bit smarter, the process manager no longer removes an UiTPAS label to re-add it later again.

---
Ticket: https://jira.uitdatabank.be/browse/III-2668

---

Note that the list of labels in the [`uitpas-service` configuration](https://github.com/cultuurnet/udb3-uitpas-service/blob/master/config.dist.yml#L4-L14) will need to be updated to add the correct card system ids instead of those (currently unused) label keys. Will also mention this in the ticket.

### DIFF
--- a/src/UiTPAS/Event/EventProcessManager.php
+++ b/src/UiTPAS/Event/EventProcessManager.php
@@ -76,6 +76,7 @@ class EventProcessManager implements EventListenerInterface
 
         $uitPasLabels = $this->uitpasLabelsRepository->loadAll();
 
+        // Create a list of UiTPAS labels that are supposed to be on the event.
         $expectedUitPasLabelsForEvent = array_map(
             function (CardSystem $cardSystem) use ($uitPasLabels, $eventId) {
                 $id = $cardSystem->getId()->toNative();
@@ -93,6 +94,7 @@ class EventProcessManager implements EventListenerInterface
         );
         $expectedUitPasLabelsForEvent = array_filter($expectedUitPasLabelsForEvent);
 
+        // Create a list of UiTPAS labels that are not supposed to be on the event.
         $uitPasLabelsToRemoveIfApplied = [];
         foreach ($uitPasLabels as $uitPasLabel) {
             $remove = true;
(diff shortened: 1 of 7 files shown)


src/UiTPAS/Event/EventProcessManager.php
```diff
@@ -82,21 +74,51 @@ private function handleEventCardSystemsUpdated(EventCardSystemsUpdated $eventCar
 
         $this->logger->info('Handling updated card systems message for event ' . $eventId);
 
-        $uitpasLabels = $this->uitpasLabelsRepository->loadAll();
+        $uitPasLabels = $this->uitpasLabelsRepository->loadAll();
+
+        $expectedUitPasLabelsForEvent = array_map(
```
jdrieghe: I was a bit confused when I saw `$expected` in a non-test context. Made me wonder if we should rethink some of the variables used here.
e.g.:
`$uitPasLabels` -> `$availableUitpasLabels`
`$expectedUitPasLabelsForEvent` -> `$requestedUitPasLabelsForEvent` (not convinced yet about this one)

src/UiTPAS/Event/EventProcessManager.php
```diff
@@ -76,6 +76,7 @@ private function handleEventCardSystemsUpdated(EventCardSystemsUpdated $eventCar
 
         $uitPasLabels = $this->uitpasLabelsRepository->loadAll();
 
+        // Create a list of UiTPAS labels that are supposed to be on the event.
```
jdrieghe: I wonder if these comments can be replaced by a descriptive private method 🤔

src/UiTPAS/Event/EventProcessManager.php
bertramakers: I see your point. My original thought process was that it's a list of labels that should be on the event based on the card systems, but some may already be on the event from before so `labelsToAdd` or something like that didn't seem correct. Since we kind of create a diff and dispatch various commands to get to that state, "expected" seemed better because it described the outcome rather then the action. But it can be confusing because of the convention to use `expected` in tests.

Would prefer something along the lines of `$applicableUitPasLabelsForEvent` over `$requestedUitPasLabelsForEvent` though since technically nothing is being requested (a domain message was sent from UiTPAS notifying is of the update of card systems on the event, but UiTPAS itself does not care about the labels AFAIK).

If you agree feel free to pick this up whenever you have some time, or I can pick it up tomorrow afternoon or  Wednesday.

src/UiTPAS/Event/EventProcessManager.php
bertramakers: Good idea, I'll pick it up tomorrow / Wednesday but if you have some time beforehand feel free to do it yourself if you like.

bertramakers:
Picked it up today anyway since they're actually pretty minor changes.

PR #356, opened 2019-08-14
Remove explicit dependency on 2dotstwice/collection

jdrieghe:
### Changed
- `CardSystems` collection class is now a simple array of `CardSystem`s

### Removed
- Dependency on `2dotstwice/collection`


src/UiTPAS/Event/Event/EventCardSystemsUpdatedDeserializer.php
```diff
@@ -34,7 +33,7 @@ public function deserialize(StringLiteral $data)
             throw new \InvalidArgumentException('Expected cardSystems property to be an array.');
         }
 
-        $cardSystems = new CardSystems();
+        $cardSystems = [];
         foreach ($dto->cardSystems as $cardSystemDTO) {
```
bertramakers: This `foreach` can now be changed to `array_map` if I'm not mistaken?

jdrieghe: `array_map` to produce an array indexed by specific keys looks pretty messy in my opinion. I'd rather leave it as a `foreach` in that case.

jdrieghe:
@bertramakers I briefly looked into the usages in other packages and I think it can be removed entirely without too much effort.

src/UiTPAS/Event/Event/EventCardSystemsUpdatedDeserializer.php
bertramakers: Ah I did not notice the keys by id. In that case `foreach` is fine yeah.

PR #360, opened 2019-08-21
III-3049 - Add bookable offer property to place read model

jdrieghe:
### Added
- Enricher for event read model to add `isDummyPlaceForEducationEvents` when the place is a dummy place

---
Ticket: https://jira.uitdatabank.be/browse/III-3049

### DIFF
--- a/app/Place/PlaceJSONLDServiceProvider.php
+++ b/app/Place/PlaceJSONLDServiceProvider.php
@@ -68,11 +68,15 @@ class PlaceJSONLDServiceProvider implements ServiceProviderInterface
 
         $app[self::JSONLD_REPOSITORY] = $app->share(
             function ($app) {
+                $dummyPlaceIds = [];
+                if (isset($app['config']['bookable_event']['dummy_place_ids'])) {
+                    $dummyPlaceIds = $app['config']['bookable_event']['dummy_place_ids'];
+                }
                 $repository = new DummyPlaceProjectionEnricher(
                     new CacheDocumentRepository(
                         $app['place_jsonld_cache']
                     ),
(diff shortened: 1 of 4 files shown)


app/Place/PlaceJSONLDServiceProvider.php
```diff
@@ -67,8 +68,13 @@ function ($app) {
...
             function ($app) {
-                $repository = new CacheDocumentRepository(
-                    $app['place_jsonld_cache']
+                $repository = new DummyPlaceProjectionEnricher(
+                    new CacheDocumentRepository(
+                        $app['place_jsonld_cache']
+                    ),
+                    [
+                        $app['config']['bookable_event']['dummy_location_ids'],
+                    ]
```
bertramakers: Isn't this a nested array now? 🤔

src/Place/DummyPlaceProjectionEnricher.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->repository = $repository;
+        $this->dummyLocationIds = $dummyLocationIds;
+    }
+
+    public function get($id)
+    {
+        $readModel = $this->repository->get($id);
+        foreach ($this->dummyLocationIds as $dummyLocationId) {
+            $body = $readModel->getBody();
+            if (strpos($body->place->{'@id'}, $dummyLocationId) !== false) {
```
bertramakers: This is indeed a URL so good call using strpos instead of an equals comparison here! ✅

app/Place/PlaceJSONLDServiceProvider.php
jdrieghe: Whoops

PR #362, opened 2019-08-22
Only show location name for dummy locations in PDF exports

bertramakers:
### Changed

- When exporting an event with a dummy location, it will only show the location name and not the full address.

---
Ticket: https://jira.uitdatabank.be/browse/III-3012

### DIFF
--- a/src/EventExport/Format/HTML/templates/export.map.activities.html.twig
+++ b/src/EventExport/Format/HTML/templates/export.map.activities.html.twig
@@ -70,7 +70,11 @@
                             <i class="fa fa-map-marker"></i>&nbsp;
                         </td>
                         <td>
-                            {{ event.address.name }}, {{ event.address.street }}, {{ event.address.postcode }} {{ event.address.municipality }}
+                            {{ event.address.name }}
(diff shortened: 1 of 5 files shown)


src/EventExport/Format/HTML/templates/export.map.activities.html.twig
```diff
@@ -70,7 +70,11 @@
                             <i class="fa fa-map-marker"></i>&nbsp;
                         </td>
                         <td>
-                            {{ event.address.name }}, {{ event.address.street }}, {{ event.address.postcode }} {{ event.address.municipality }}
+                            {{ event.address.name }}
+                            {%-  if not event.address.isDummyAddress -%}
```
jdrieghe: Curious what the reason behind these custom twig delimiters is 🤔

bertramakers: It's to remove the whitespace in front and/or behind the opening/closing braces.
In this case it's not that important but in `export.tips.html.twig` I had to use them to keep the structure of the outputted HTML the same as before since there's some unit tests that compare it to HTML files in the tests folder.

PR #363, opened 2019-08-22
III-3053 - Statically set the list of dummy place ids during bootstrapping

jdrieghe:
### Added
- Bootstrapping to set a list of configured dummy ids on `LocationId` value object

---
Ticket: https://jira.uitdatabank.be/browse/III-3053

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1319,7 +1319,7 @@ $app->register(new \CultuurNet\UDB3\Silex\AuditTrailServiceProvider());
 $app->register(new CuratorsServiceProvider());
 
 if (isset($app['config']['bookable_event']['dummy_place_ids'])) {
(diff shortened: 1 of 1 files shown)


bootstrap.php
```diff
@@ -1317,4 +1318,8 @@ function () {
 
 $app->register(new CuratorsServiceProvider());
 
+if (isset($app['config']['bookable_event']['dummy_place_ids'])) {
+    LocationId::setDummyPlaceForEducationIds($app['config']);
```
bertramakers:
```suggestion
    LocationId::setDummyPlaceForEducationIds($app['config']['bookable_event']['dummy_place_ids']);
```

jdrieghe: 😅 Looks like I committed a bit too quickly

PR #365, opened 2019-08-26
III-3045 - Add CLI command to mark place as duplicate

jdrieghe:
### Added
- CLI command to mark place as duplicate
- Configured `MarkAsDuplicateCommandHandler` on the command bus.
- Configured `LocationMarkedAsDuplicateProcessManager` on the event bus

### Changed
- Updated `cultuurnet/udb3` to latest version

### Removed
- A non-existing class that was imported in a service provider

---
Ticket: https://jira.uitdatabank.be/browse/III-3045

### DIFF
--- a/app/Place/PlaceEditingServiceProvider.php
+++ b/app/Place/PlaceEditingServiceProvider.php
@@ -3,6 +3,7 @@
 namespace CultuurNet\UDB3\Silex\Place;
 
 use Broadway\UuidGenerator\Rfc4122\Version4Generator;
+use CultuurNet\UDB3\Event\LocationMarkedAsDuplicateProcessManager;
 use CultuurNet\UDB3\Offer\OfferEditingServiceWithLabelMemory;
 use CultuurNet\UDB3\Place\Commands\PlaceCommandFactory;
 use CultuurNet\UDB3\Place\DefaultPlaceEditingService;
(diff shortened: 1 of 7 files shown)


app/Place/PlaceEditingServiceProvider.php
```diff
@@ -38,6 +38,15 @@ function ($app) {
                 );
             }
         );
+
+        $app[LocationMarkedAsDuplicateProcessManager::class] = $app->share(
```
bertramakers: Only suggestion I'd have is that this might be more related to events so we might want to put it in the `EventEditingServiceProvider` instead, but it's a grey area so I'm fine with either one.

jdrieghe: I'll move it tomorrow morning ;)

tests/Http/Jobs/ReadRestControllerTest.php
```diff
@@ -44,7 +44,7 @@ public function it_returns_a_json_response_for_existing_job()
 
         $expectedResponse = new JsonResponse($jobStatus->toNative());
 
-        $this->assertEquals($expectedResponse, $response);
+        $this->assertEquals($expectedResponse->getContent(), $response->getContent());
```
bertramakers: Yep this used to happen a lot in the past. We always fixed it like this but I guess some hadn't been changed yet.

PR #366, opened 2019-08-26
III-3042 - only show a small calendar format when the calendarType is…

brampauwelyn:
### Changed
- Show **small** calendar format when the `calendarType` is multiple, otherwise show **large** calendar format

---
Ticket: [https://jira.uitdatabank.be/browse/III-3042](https://jira.uitdatabank.be/browse/III-3042)

### DIFF
--- a/src/EventExport/Format/HTML/HTMLEventFormatter.php
+++ b/src/EventExport/Format/HTML/HTMLEventFormatter.php
@@ -2,6 +2,7 @@
 
 namespace CultuurNet\UDB3\EventExport\Format\HTML;
 
+use CultuurNet\UDB3\CalendarType;
 use CultuurNet\UDB3\Event\EventType;
 use CultuurNet\UDB3\Event\ReadModel\JSONLD\Specifications\EventSpecificationInterface;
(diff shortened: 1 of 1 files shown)


src/EventExport/Format/HTML/HTMLEventFormatter.php
```diff
@@ -199,7 +199,9 @@ private function addCalendarInfo($eventId, stdClass $event, array &$formattedEve
     {
         if ($this->calendarSummaryRepository) {
             try {
-                $calendarSummary = $this->calendarSummaryRepository->get($eventId, ContentType::HTML(), Format::SMALL());
+                $calendarType = $event->calendarType;
+                $calendarSummaryFormat = ($calendarType == "multiple" ? Format::SMALL() : Format::LARGE());
```
jdrieghe:
```suggestion
                $calendarSummaryFormat = $calendarType === "multiple" ? Format::SMALL() : Format::LARGE();
```

bertramakers: Maybe also use the `CalendarType` enum instead of hardcoding `multiple` as a string here.
So something like `CalendarType::fromNative($calendarType)->is(CalendarType::MULTIPLE())`.

PR #369, opened 2019-08-29
Remove MyOrganizers services & controllers

bertramakers:
### Removed

* Removed service provider for MyOrganizers projectors and lookup service
* Removed controller provider for MyOrganizers controller
* Removed MyOrganizers controller

---
Ticket: https://jira.uitdatabank.be/browse/III-3066


bertramakers:
Still need to add the migration to drop the related database table, will do so later today.

PR #391, opened 2019-10-31
III-3055 Update udb3-php dependency

bertramakers:
### Changed

- Updated cultuurnet/udb3 dependency so we can deploy III-3055

---
Ticket: https://jira.uitdatabank.be/browse/III-3055

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "cultuurnet/silex-amqp": "~0.1",
     "cultuurnet/silex-service-provider-jwt": "~0.1",
     "cultuurnet/udb2-domain-events": "~0.1",
-    "cultuurnet/udb3": "dev-master#7b233a909118737b2cafec49fed1adde3b8d5ba0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -30,7 +30,7 @@
     "cultuurnet/silex-amqp": "~0.1",
     "cultuurnet/silex-service-provider-jwt": "~0.1",
     "cultuurnet/udb2-domain-events": "~0.1",
-    "cultuurnet/udb3": "dev-master#7b233a909118737b2cafec49fed1adde3b8d5ba0",
+    "cultuurnet/udb3": "~0.1",
```
bertramakers: I was able to remove the fixed commit hash here this time. Composer didn't want to update to this commit in the past, but this time it updated to the latest commit without problem by using `~0.1`. So moving forward this is better because then we can just run `composer up cultuurnet/udb3` without having to change the commit hash.

PR #392, opened 2019-11-04
Feature/III-3039

konrad126:
### Added
- regression test for exceding the maximum number of export items
- config value for max export items
---
[Jira](https://jira.uitdatabank.be/browse/III-3039)

### DIFF
--- a/src/EventExport/EventExportService.php
+++ b/src/EventExport/EventExportService.php
@@ -3,6 +3,7 @@
 namespace CultuurNet\UDB3\EventExport;
 
 use Broadway\UuidGenerator\UuidGeneratorInterface;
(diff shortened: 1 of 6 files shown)


src/EventExport/EventExportService.php
```diff
@@ -115,6 +127,10 @@ public function exportEvents(
             throw ($e);
         }
 
+        if ($totalItemCount > $this->maxAmountOfItems) {
+            throw new MaximumNumberOfExportItemsExceeded();
```
bertramakers: Wouldn't this also throw if the `$totalItemCount` is more than the max amount, but there is a selection which is smaller? Eg. total count is 5000, max count is 2000, but selection is 10. In that case it shouldn't throw because we're just exporting 10 items. This whole "pre-query" thing actually seems unnecessary even when we're exporting a selection.

PR #395, opened 2019-11-07
Only allow god users to access an event's history endpoint

bertramakers:
### Changed

- Only god users can now access an event's history endpoint

---
Ticket: https://jira.uitdatabank.be/browse/III-3115


app/Event/DeprecatedEventControllerProvider.php
```diff
@@ -15,25 +15,6 @@ class DeprecatedEventControllerProvider implements ControllerProviderInterface
...
-        $app['event_editing_controller'] = $app->share(
-            function (Application $app) {
-                return new EditEventRestController(
-                    $app['event_editor'],
-                    $app['media_manager'],
-                    $app['event_iri_generator']
-                );
-            }
-        );
-
```
bertramakers: These services are immediately overwritten in `EventControllerProvider.php`, and more importantly broken because of missing arguments.

PR #397, opened 2019-11-13
Add api key to event metadata and history projection

bertramakers:
### Changed

- Refactored the way context / metadata is determined and set.

### Fixed

- The api key is now added to all events' metadata instead of only in some edge cases.

### Removed

- Removed `auth_jwt` and `uitid_token_credentials` from the metadata that gets stored as they were cast to `{}` anyway and are too sensitive to store.

---
Ticket: https://jira.uitdatabank.be/browse/III-3113


bertramakers:
The `Impersonator` class is used by workers that execute commands asynchronously. That's why the metadata still has to contain the JWT and UiTID token credentials, so those commands can be authorized correctly. However once we persist the metadata they are filtered out by `ContextFactory::prepareForLogging()`.

PR #398, opened 2019-11-14
[WIP] III-3114 Add api name to metadata

bertramakers:
### Added

- Added api name to context and event metadata

---
Ticket: https://jira.uitdatabank.be/browse/III-3114

---

Not mentioned in the ticket were the UiTPAS and Curators AMQP listeners that can automatically add/remove labels based on messages from those services. I've added those as well.


bertramakers:
Still needs an update of `cultuurnet/udb3` to actually update the history projection with the api name.

PR #405, opened 2019-11-19
III-3148 Move organizer code

bertramakers:
### Added

- Added the Organizer code removed in udb3-php in https://github.com/cultuurnet/udb3-php/pull/444

---
Ticket: https://jira.uitdatabank.be/browse/III-3148


bertramakers:
> Lots of moving:)

Yep, if we keep this up we will have a lot less overhead of working in udb3-php and having to run `composer up` and creating multiple pull requests.

I'll document how to figure out what can be moved and what not on Confluence so you can also give it a shot whenever you're working on something that's in udb3-php.

PR #407, opened 2019-11-21
Feature: III 3137

konrad126:
### Added
- History projector for Places
- REST controller for Places history

---
Ticket:
[Jira](https://jira.uitdatabank.be/browse/III-3137)


bootstrap.php
```diff
@@ -459,6 +459,31 @@ function ($app) {
...
+            $app['places_history_repository']
+        );
+        return $projector;
+    }
+);
+
+$app['places_history_repository'] = $app->share(
+    function ($app) {
+        return new \CultuurNet\UDB3\Doctrine\ReadModel\CacheDocumentRepository(
+            $app['place_history_cache']
```
bertramakers: Nit: I'm not sure we need to create a separate service in the service container for this.
Could just be `$app['cache']('place_history');`

bootstrap.php
```diff
@@ -459,6 +459,31 @@ function ($app) {
     }
 );
 
+$app['place_history_projector'] = $app->share(
```
bertramakers: Nit: Ideally these services would be moved to a service provider. Would also make it a bit easier to get a better overview when porting away from Silex.

src/Http/Place/HistoryPlaceRestController.php
```diff
@@ -0,0 +1,80 @@
...
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+class HistoryPlaceRestController
+{
+    use ApiProblemJsonResponseTrait;
+
+    private const HISTORY_ERROR_NOT_FOUND = 'An error occurred while getting the history of the event with id %s!';
+    private const HISTORY_ERROR_GONE = 'An error occurred while getting the history of the event with id %s which was removed!';
+    private const HISTORY_ERROR_FORBIDDEN = 'Forbidden to access event history.';
```
bertramakers: These error messages still use the name `event` but it should be `place`

src/Place/ReadModel/Enum/EventDescription.php
```diff
@@ -0,0 +1,21 @@
...
+class EventDescription
+{
+    const CREATED = 'Created';
+    const DELETED = 'Deleted';
+    const LABEL_ADDED = 'label-added';
+    const LABEL_REMOVED = 'label-removed';
+    const DESCRIPTION_TRANSLATED = 'description-translated';
+    const TITLE_TRANSLATED = 'title-translated';
+    const PLACE_IMPORTED_FROM_UDB2 = 'place-imported-from-udb2';
+    const PLACE_UPDATED_FROM_UDB2 = 'place-updated-from-udb2';
```
bertramakers: This class is not used anymore right?

src/Place/ReadModel/History/HistoryProjector.php
```diff
@@ -0,0 +1,218 @@
...
+
+namespace CultuurNet\UDB3\Place\ReadModel\History;
+
+use Broadway\Domain\DateTime as BroadwayDateTime;
+use Broadway\Domain\DomainMessage;
+use Broadway\Domain\Metadata;
+use Broadway\EventHandling\EventListenerInterface;
+use CultuurNet\UDB3\Cdb\EventItemFactory;
+use CultuurNet\UDB3\Event\ReadModel\DocumentRepositoryInterface;
+use CultuurNet\UDB3\Event\ReadModel\History\Log;
```
bertramakers: This `Log` class is part of the `Event` namespace. Can you move it so it's in for example `CultuurNet\UDB3\History`?

src/Place/ReadModel/History/Log.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace CultuurNet\UDB3\Place\ReadModel\History;
+
+use DateTime;
+use JsonSerializable;
+
+class Log implements JsonSerializable
```
bertramakers: This class is not actually used by the projector. Personally I'd prefer not to duplicate it and move the one from event to a shared namespace instead, since it's important that both the event and place history have the same JSON structure, and we'd be duplicating that.

src/Place/ReadModel/History/HistoryProjector.php
```diff
@@ -0,0 +1,218 @@
...
+     * @var DocumentRepositoryInterface
+     */
+    private $documentRepository;
+
+    public function __construct(DocumentRepositoryInterface $documentRepository)
+    {
+        $this->documentRepository = $documentRepository;
+    }
+
+    public function handle(DomainMessage $domainMessage)
```
bertramakers: The way the `handle` method is implemented is very different from https://github.com/cultuurnet/udb3-silex/blob/master/src/Event/ReadModel/History/HistoryProjector.php#L42 which may be a bit confusing in the long run.

Additionally most events could be handled by the `createOfferEventHandler` in that history projector, so maybe we could extract some of that shared logic to a base class / trait / separate class we can inject?

It would also make it easier to add more data from the metadata in the future, as we will now have to do that twice with the `get...FromMetadata` methods in both classes.

src/Place/ReadModel/Enum/EventDescription.php
konrad126: Yes, but just check if the description for a deleted event is ok now

src/Place/ReadModel/History/HistoryProjector.php
konrad126: I have extracted a `BaseHIstoryProjector` and put some shared functionality there. 

As for the `createOfferEventHandler` method. We would have to modify it to be used with Place events since they have `getPlaceId()` method instead of `getEventId()`.

The approach I used in "my" handler does require more typing since there is one method for each event and that method is just a proxy method to the `writeHistory` method. But, here I guess we're entering the highly opinionated waters, to me, it feels more explicit when coming to code to see what is happening.

So, for now, I haven't extracted the `createOfferEventHandler` but we can discuss it some more ofc :))

bertramakers: > We would have to modify it to be used with Place events since they have getPlaceId() method instead of getEventId().

`createOfferEventHandler` uses `getItemId()` which is also what all of your handlers use except for `projectPlaceCreated` 😉 

Most of the events handled in both history projectors will be inheriting from `CultuurNet\UDB3\Offer\Events\AbstractEvent`, which uses `getItemId()`.

Only events specific to places or events will be extending from other classes which use `getPlaceId()` or `getEventId()`. But that's a minority.

src/Place/ReadModel/History/HistoryProjector.php
```diff
@@ -0,0 +1,135 @@
...
+            $event->getPlaceId(),
+            $this->createGenericLog($domainMessage, 'Aangemaakt in UiTdatabank')
+        );
+    }
+
+    private function projectPlaceDeleted(PlaceDeleted $event, DomainMessage $domainMessage): void
+    {
+        $this->writeHistory(
+            $event->getItemId(),
+            $this->createGenericLog($domainMessage, 'Aangemaakt verwijderd')
```
bertramakers: This message should be `Place verwijderd`. `Aangemaakt verwijderd` means `Created deleted`. 😉 
I'm not sure we had to add this event handler already, as the history projector for events doesn't handle `EventDeleted` either, but we can leave it in now.

PR #411, opened 2019-12-04
Feature/iii 3144

bertramakers:
### Changed

- Changed how history logs are stored, so previous logs are overwritten when replaying

### Fixed

- Fixed PlaceImportedFromUDB2 logs

---
Ticket: https://jira.uitdatabank.be/browse/III-3144


src/History/Log.php
```diff
@@ -60,9 +62,30 @@ public function __construct(
...
-    public function getId(): string
+    public function getUniqueKey(): string
     {
-        return $this->id;
+        return $this->id . '_' . $this->date->format('c');
+    }
+
+    public function withoutAuthor(): Log
+    {
+        $c = clone $this;
```
konrad126: I know it's nitpicky but even though this is a 3-line function I would still like if the variable has a name like `log` or `clone` than just `c` :)

tests/Place/ReadModel/History/HistoryProjectorTest.php
```diff
@@ -209,14 +211,20 @@ public function it_projects_PlaceUpdatedFromUDB2_event()
...
-        $this->assertEquals(
-            $history,
-            $document->getBody()
-        );
+        $body = array_values((array) $document->getBody());
+
+        foreach ($history as $log) {
+            // Do not use assertContains() here, as it doesn't work the same as in_array for some reason.
+            if (!in_array($log, $body)) {
+                var_dump($log);
```
konrad126: Are the var_dumps some leftover?

bertramakers: Yes, I forgot to remove it indeed.

bertramakers:
Btw this is ready for review again @konrad126

PR #412, opened 2019-12-04
Fix history for places imported from udb2

bertramakers:
### Fixed

- Fixed history projection for places imported from udb2


bertramakers:
I split this off from #411 because it might take a while until we can deploy #411 and this is quite urgent to fix.

PR #413, opened 2019-12-05
III-3153

konrad126:
### Added
- logger to Results Generator
- logger to Export Command Handler

---
Ticket:
[jira](https://jira.uitdatabank.be/browse/III-3153)

### DIFF
--- a/app/Export/ExportServiceProvider.php
+++ b/app/Export/ExportServiceProvider.php
@@ -102,7 +102,7 @@ class ExportServiceProvider implements ServiceProviderInterface
 
         $app['event_export_log_handler'] = $app->share(
             function () {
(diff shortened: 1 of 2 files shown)


app/Export/ExportServiceProvider.php
```diff
@@ -95,6 +99,20 @@ function (CommandBusInterface $commandBus, Application $app) {
                 return $commandBus;
             }
         );
+
+        $app['event_export_log_handler'] = $app->share(
+            function () {
+                return new StreamHandler(__DIR__ . '/../log/export.log');
```
bertramakers: Are you sure this path is correct? Wouldn't this resolve to `/app/log/export.log` instead of `/log/export.log`?
Same for the other logger.

PR #426, opened 2020-01-21
III-3203

bertramakers:
### Added

- Added feature toggle to switch user query endpoints from UiTID v1 to Auth0

### Removed

- Removed unused / obsolete UiTID service classes and providers

---
Ticket: https://jira.uitdatabank.be/browse/III-3203


src/User/Auth0UserIdentityResolver.php
```diff
@@ -20,20 +20,28 @@ public function __construct(Management $auth0)
...
+        
+        $user = array_shift($users);
 
-        var_dump($user);
-        exit;
-
-        return null;
+        return new UserIdentityDetails(
+            new StringLiteral($user['user_id']),
+            new StringLiteral($user['name']),
```
bertramakers: I think this should be `nickname`?
See the JSON example in https://madewithlove.slack.com/archives/CSLQN091S/p1579707751114000

src/User/Auth0UserIdentityResolver.php
```diff
@@ -46,12 +46,21 @@ public function getUserById(StringLiteral $userId): ?UserIdentityDetails
...
+
+        if (empty($users)) {
+            return null;
+        }
+
+        $user = array_shift($users);
+
+        return new UserIdentityDetails(
+            new StringLiteral($user['user_id']),
+            new StringLiteral($user['name']),
```
bertramakers: Same comment, should probably also be `nickname`.
Maybe consider moving this to a private method to convert the `array` to `UserIdentityDetails`?

bertramakers: Ah okay you've already added a normalize method in a later commit. 👍

src/User/Auth0UserIdentityResolver.php
bertramakers: Same for the query to fetch users by username.

app/Auth0/Auth0ServiceProvider.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $app[Management::class] = $app::share(
+            function (Application $app) {
+                // @TODO make domain configurable
+                // @TODO the regular JWT is not valid for the management API
+                // @see https://auth0.com/docs/api/management/v2/tokens
+                return new Management(
+                    (string) $app['jwt']->jwtToken(),
+                    'publiq-acc.eu.auth0.com'
+                );
```
bertramakers: @konrad126 these `@TODO`s should also still be done

config.yml.dist
```diff
@@ -137,7 +137,8 @@ jwt:
         file: public-auth0.pem
     validation:
       iss: https://publiq-acc.eu.auth0.com/
-
+    domain: publiq-acc.eu.auth0.com
+
```
bertramakers: @konrad126 can you create an appconfig PR for this please? https://bitbucket.org/cultuurnet/appconfig/

As well as a PR in https://bitbucket.org/cultuurnet/udb3-vagrant/src for local installations.
You can find an example here: https://bitbucket.org/cultuurnet/udb3-vagrant/pull-requests/75

config.yml.dist
```diff
@@ -138,7 +138,9 @@ jwt:
     validation:
       iss: https://publiq-acc.eu.auth0.com/
     domain: publiq-acc.eu.auth0.com
-    
+    client_id:
+    client_secret:
```
bertramakers: Same comment as before, we need appconfig and udb3-vagrant PRs for this change as well.

app/Auth0/Auth0ServiceProvider.php
```diff
@@ -5,30 +5,42 @@
...
-                    (string) $app['jwt']->jwtToken(),
-                    $app['config']['jwt']['auth0']['domain']
+                $provider = new Auth0ManagementTokenProvider(
+                    new Auth0ManagementTokenGenerator(
+                        new Client(),
+                        $app['config']['jwt']['auth0']['client_id'],
+                        $app['config']['jwt']['auth0']['domain'],
+                        $app['config']['jwt']['auth0']['client_secret']
+                    ),
+                    new InMemoryRepository(),
```
bertramakers: @konrad126 This was fine for getting it up and running quickly, but we should use a Redis cache instead on production. Can you add another implementation for this repository using Redis? We already have examples in other places. Just check for usages of `$app['cache']`. For example if you call `$app['cache']('auth0-management-token')` you will get an instance of `Doctrine\Common\Cache\PredisCache` to store and retreive the token.

PR #433, opened 2020-02-27
Bug: III-3243

konrad126:
### Fixed
- emails with `+` sing would not return correct result

### Added
- urlencode emails before sending them as queries to Auth0 SDK

---
Ticket:
[Jira](https://jira.uitdatabank.be/browse/III-3243)

### DIFF
--- a/src/User/Auth0UserIdentityResolver.php
+++ b/src/User/Auth0UserIdentityResolver.php
@@ -32,12 +32,12 @@ final class Auth0UserIdentityResolver implements UserIdentityResolverInterface
 
     public function getUserByEmail(EmailAddress $email): ?UserIdentityDetails
     {
-        return $this->fetchUser('email:"' . $email . '"');
+        return $this->fetchUser('email:"' . urlencode($email->toNative()) . '"');
     }
 
     public function getUserByNick(StringLiteral $nick): ?UserIdentityDetails
     {
-        return $this->fetchUser('email:"' . $nick . '" OR nickname:"' . $nick . '"');
+        return $this->fetchUser('email:"' . urlencode($nick->toNative()) . '" OR nickname:"' . urlencode($nick->toNative()) . '"');
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/User/Auth0UserIdentityResolver.php
```diff
@@ -32,12 +32,12 @@ public function getUserById(StringLiteral $userId): ?UserIdentityDetails
...
     public function getUserByEmail(EmailAddress $email): ?UserIdentityDetails
     {
-        return $this->fetchUser('email:"' . $email . '"');
+        return $this->fetchUser('email:"' . urlencode($email->toNative()) . '"');
     }
 
     public function getUserByNick(StringLiteral $nick): ?UserIdentityDetails
     {
-        return $this->fetchUser('email:"' . $nick . '" OR nickname:"' . $nick . '"');
+        return $this->fetchUser('email:"' . urlencode($nick->toNative()) . '" OR nickname:"' . urlencode($nick->toNative()) . '"');
```
bertramakers: Good idea doing this for nicknames as well 👍

tests/User/Auth0UserIdentityResolverTest.php
```diff
@@ -87,7 +87,7 @@ public function it_gets_user_by_email()
 
         $users->expects($this->atLeast(1))
             ->method('getAll')
-            ->with(['q' => 'email:"ivo@hdz.com"'])
+            ->with(['q' => 'email:"'.urlencode("ivo@hdz.com").'"'])
```
bertramakers: Nit: In tests, I like hardcoding the expected value instead of running it through the same conversion functions. Then the test is separated from the implementation and you know for sure the output is what you'd expect it to be. Of course in this case `urlencode` should work fine as it's a built-in function and its usage is straightforward, so we can leave this as-is if you want. It's more of an issue with more complicated functions where we might actually be using the wrong options / flags / arguments and we repeat that mistake in the test.

konrad126: Valid point, I will update it :)

PR #434, opened 2020-03-24
Feature: III-3227

konrad126:
### Added
- AddressRemoved handling to OrganizerLDProjector
- RemoveAddress command/handler

### Changed
- bumped udb3-php dependencyBump udb3-php dependency

---
Ticket:
[Jira](https://jira.uitdatabank.be/browse/III-3227)

### DIFF
--- a/src/Organizer/DefaultOrganizerEditingService.php
+++ b/src/Organizer/DefaultOrganizerEditingService.php
@@ -13,6 +13,7 @@ use CultuurNet\UDB3\Language;
 use CultuurNet\UDB3\Organizer\Commands\AddLabel;
 use CultuurNet\UDB3\ContactPoint;
 use CultuurNet\UDB3\Organizer\Commands\DeleteOrganizer;
+use CultuurNet\UDB3\Organizer\Commands\RemoveAddress;
 use CultuurNet\UDB3\Organizer\Commands\RemoveLabel;
 use CultuurNet\UDB3\Organizer\Commands\UpdateAddress;
 use CultuurNet\UDB3\Organizer\Commands\UpdateContactPoint;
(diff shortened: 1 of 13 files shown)


src/Organizer/OrganizerLDProjector.php
```diff
@@ -307,6 +308,19 @@ private function applyAddressTranslated(AddressTranslated $addressTranslated)
         return $this->applyAddress($addressTranslated, $addressTranslated->getLanguage());
     }
 
+    public function applyAddressRemoved(AddressRemoved $addressRemoved)
+    {
+        $organizerId = $addressRemoved->getOrganizerId();
+        $document = $this->repository->get($organizerId);
+        $jsonLD = $document->getBody();
+
+        if (!isset($jsonLD->address)) {
```
bertramakers: Do we need this check?

src/Organizer/DefaultOrganizerEditingService.php
```diff
@@ -150,4 +151,11 @@ public function delete(string $id): void
             new DeleteOrganizer($id)
         );
     }
+
+    public function removeAddress(string $organizerId): void
```
bertramakers: Okay for now, but I'd suggest you and @jdrieghe try to remove these services in the long run. In most cases they don't do anything extra aside from dispatching a command, and in the other rare cases they do something extra it might be able to move it to the command handler instead. In the case of the `create` method, a `CreateOrganizer` command could be added. But this is not required for this PR, just a general comment for a possible future improvement.

bertramakers:
Also, remember to document the new endpoint in https://github.com/cultuurnet/udb3-schema

PR #437, opened 2020-04-09
Feature: III-3227

konrad126:
### Added
- dispatch RemoveAddres command if Adress data is not present in import

---
Ticket:
[Jira](https://jira.uitdatabank.be/browse/III-3227)


bertramakers:
Remember to also add @jdrieghe as potential reviewer

PR #445, opened 2020-04-30
Clean up rolled out uitid-api-key-required feature flag

jdrieghe:
### Removed
- Already rolled out feature flag: `uitid-api-key-required`

---
Overview: https://docs.google.com/spreadsheets/d/1CKZEbIRsh7rAP5H1SEAwxR7T1XR6VLDu_F6GZLQGrUw/edit#gid=0


bertramakers:
The config bypass looks good from a technical standpoint @jdrieghe 👍 
The only nit I'd have is that the name is "bypass authentication" but it only bypasses the API key check, not the JWT check. The logic is consistent with the old feature flag, but the name is maybe a bit too broad.
But again at this point it's a nice to have so I don't want to hold this up needlessly if it prevents us from focusing on other work.

PR #449, opened 2020-05-07
Add events code from udb3 php

jdrieghe:
### Added
- Commands from `Events` namespace of udb3-php

---
Related PR in udb3-php: https://github.com/cultuurnet/udb3-php/pull/458

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "cultuurnet/search-v3": "dev-master",
     "cultuurnet/silex-amqp": "~0.1",
     "cultuurnet/udb2-domain-events": "~0.1",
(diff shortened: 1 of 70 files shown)


composer.json
```diff
@@ -30,7 +30,7 @@
     "cultuurnet/search-v3": "dev-master",
     "cultuurnet/silex-amqp": "~0.1",
     "cultuurnet/udb2-domain-events": "~0.1",
-    "cultuurnet/udb3": "~0.1",
+    "cultuurnet/udb3": "dev-move-event-related-code-to-udb3-silex as 0.1",
```
konrad126, line 33: Should we switch this back?

jdrieghe: Only after we merge the other PR in my opinion. I changed it here to prove the CI would succeed after the code was removed from the dependency.

PR #450, opened 2020-05-12
Internalize more event related code from udb3 php

jdrieghe:
### Added
- A bunch of code from https://github.com/cultuurnet/udb3-php that was only used in this application

---
Related PR: https://github.com/cultuurnet/udb3-php/pull/459


bertramakers:
Seems like there's some coding standard errors in between the warnings, causing the Travis check to fail. Not many though, should be easy to fix I think.

jdrieghe:
@bertramakers linting issues resolved

PR #456, opened 2020-05-26
Feature: III-3278

konrad126:
### Added
- Cache (wrapper) class
- EventStreamBuilder
- SearchRepositoryFactory

### Changed
- inject dependencies into command instead of directly accessing service container
- remove dependency/inheritance of KNP command

### Removed
- unused EventCdbXmlCommand 
- unused UpdateCdbXMLCommand
- PermissionCommand
- ImportSavedSearchesCommand
- ImportRoleConstraintsCommand
- Support for filesystem cache during replays

---
Ticket:
[Jira](https://jira.uitdatabank.be/browse/III-3278)


app/SavedSearches/SearchRepositoryFactory.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function createForVersion(string $version): UDB3SavedSearchRepository
+    {
+        if (!in_array($version, [SapiVersion::V2, SapiVersion::V3])) {
+            throw new \InvalidArgumentException('Invalid version value: ' . $version);
+        }
+
+        if ($version === SapiVersion::V3) {
+            $version = new StringLiteral('saved_searches_sapi3');
```
jdrieghe, line 46: I'm not a big fan of re-using the variable here. As a class gets more complicated, these are often sources of hard to spot bugs.

app/Console/ImportSavedSearchesCommand.php
```diff
@@ -86,7 +97,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
             $output->writeln('Importing query with name: ' . $record['NAME']);
 
-            $this->getUDB3SavedSearchesRepository($input)->write(
+            $this->repositoryFactory->createForVersion($input->getOption(self::SAPI_VERSION))->write(
```
jdrieghe, line 100: Do we still use SAPI version 2 here? @bertramakers ?

app/Console/ReplayCommand.php
```diff
@@ -38,12 +38,17 @@ class ReplayCommand extends AbstractCommand
      * @var EventBusInterface
      */
     private $eventBus;
+    /**
+     * @var array
+     */
+    private $config;
 
-    public function __construct(CommandBusInterface $commandBus, EventStreamBuilder $eventStreamBuilder, EventBusInterface $eventBus)
+    public function __construct(CommandBusInterface $commandBus, EventStreamBuilder $eventStreamBuilder, EventBusInterface $eventBus, array $config)
```
jdrieghe: I don't like injecting the entire config array. It's usually a better idea to only inject the configuration variables you need.

app/Console/ImportSavedSearchesCommand.php
bertramakers: There is still a table for saved searches for sapi2, but it should be removed in the near future.
There's even a ticket to remove all dependencies on sapi2 from udb3 in the next up column: https://jira.uitdatabank.be/browse/III-3097
I've added a comment to also remove the code for sapi2 saved searches and role constraints from the backend.

So I think you could directly inject the sapi3 repository instead and remove the `sapi_version` option.

Thinking a bit further though, I'm not even sure if this command is still needed. Since most are only run manually by a select few people, we could check with @willaerk and @corneelwille which ones we actually still need and remove the obsolete ones.

bertramakers:
Do you think it's possible to update to `symfony/console 3.x` so we can use a custom command loader like we did [here](https://github.com/cultuurnet/udb3-search-service/blob/master/app/CommandServiceProvider.php#L70)? That way we don't have to instantiate all the dependencies of each command when we're just running one.

bertramakers:
You'll also have to update the `ConsumeCommand` in this package to use dependency injection. https://github.com/cultuurnet/silex-amqp/blob/master/src/Console/ConsumeCommand.php

However that package is also used by other apps like `udb3-cdbxml-service` and `udb3-uitpas-service`. For `udb3-search-service`, we copy-pasted it here: https://github.com/cultuurnet/udb3-search-service/blob/master/app/Console/ConsumeCommand.php

Since the package name literally has `silex` in the name, I'd suggest you just copy-paste the one from `ubd3-search-service` into `udb3-silex` for now. It's not that much code and hasn't been updated in 2+ years, so it's pretty stable and safe to duplicate across the codebases instead of putting it in a shared package I think.

bertramakers:
> Removed
> * unused EventCdbXmlCommand
> * unused UpdateCdbXMLCommand

For future reference if anyone sees this later: They were removed because they were already broken and the necessary classes to make them work didn't exist anymore. Their functionality was also outdated ever since udb2 was shut down.

app/Cache.php
```diff
@@ -0,0 +1,37 @@
...
+     * @var Application
+     */
+    private $application;
+
+    public function __construct(Application $application)
+    {
+        $this->application = $application;
+    }
+
+    public function useCache(string $name)
```
bertramakers, line 27: Can you add a return type here?

app/Console/ReplayCommand.php
jdrieghe: Hmm, it looks like we're using this config so we can write to it. Not optimal, but probably not something we should fix right now.

bertramakers: Shouldn't it be passed by reference then?

jdrieghe: Good point. Will dive into it with @konrad126 tomorrow.

jdrieghe: You can see here how it got re-assigned to the global `$app['config']` after setting the properties. https://github.com/cultuurnet/udb3-silex/pull/456/files#diff-b729b0cc33328f7ab4412b9adc2d2a6fL238

That means there's indeed a bug here.
I would suggest to keep that part of the code as it was and not inject the configuration (not even by reference). When we refactor this towards a new container, chances are big we'll have a `Config` component that can override properties which we can then inject here.

jdrieghe: On the other hand, we can't remove `->getSilexApplication()` in that case. So maybe passing by reference is cleaner for now.

jdrieghe: Should be fixed

jdrieghe: Passing by reference didn't work:
```
PHP Notice:  Indirect modification of overloaded element of Silex\Application has no effect in /vagrant/source/udb3-silex/bin/udb3.php on line 79
```

Solved it by wrapping it in a `ConfigWriter` class for now

bootstrap.php
```diff
@@ -295,36 +295,12 @@ function (Application $app) {
...
-            $cache = new \Doctrine\Common\Cache\FilesystemCache($cacheDirectory);
-
-            return $cache;
-        };
-    }
-);
-
-$app['cache'] = $app->share(
-    function (Application $app) {
-        $activeCacheType = $app['config']['cache']['active'] ?: 'filesystem';
```
bertramakers, line 321: I'm wondering what would be a good way to have a reminder to also remove this from the `appconfig` once this change has been deployed. Any suggestions how we can keep track of that?

bertramakers: It's technically not _necessary_ to remove it, but if we don't we end up with a lot of obsolete config which will get confusing in the long run.

bertramakers: And if we remove it from the `appconfig` right now, it will cause a deploy of the config and suddenly revert it to `filesystem` until this change is also deployed.

jdrieghe: We can keep a simple todo list with items to process "after next deploy"?

jdrieghe:
@bertramakers I'll look into the Symfony update

bertramakers:
> @bertramakers I'll look into the Symfony update

Maybe leave it for now, as discussed in the call we need to keep some time on the budget for these consolidations in case there are problems after the deploy and testing. We can do the update and improvement in the Silex EOL project once we start that.

PR #463, opened 2020-06-04
Productions - Part 1

jdrieghe:
### Added
- Productions entity and table
- Command to group multiple events as a new production

---
Ticket:

### DIFF
--- a/app/Migrations/Version20200602134547.php
+++ b/app/Migrations/Version20200602134547.php
@@ -44,6 +44,7 @@ class Version20200602134547 extends AbstractMigration
 
         $table->setPrimaryKey(['event_id']);
         $table->addIndex(['production_id']);
+        $table->addIndex(['name']);
     }
 
     public function down(Schema $schema)
--- a/app/Migrations/Version20200602134547.php
+++ b/app/Migrations/Version20200602134547.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace CultuurNet\UDB3\Silex\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20200602134547 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $table = $schema->createTable('productions');
+        $table->addColumn(
+            'event_id',
+            'guid',
+            [
+                'length' => 36,
+                'notnull' => true,
+            ]
+        );
+        $table->addColumn(
+            'production_id',
+            'guid',
+            [
+                'length' => 36,
+                'notnull' => true,
+            ]
+        );
+        $table->addColumn(
+            'name',
+            'string',
+            [
+                'length' => 32,
+                'notnull' => true,
+            ]
+        );
+        $table->addColumn(
+            'added_at',
+            'date_immutable',
+            [
+                'notnull' => true,
+            ]
+        );
+
+        $table->setPrimaryKey(['event_id']);
+        $table->addIndex(['production_id']);
+    }
+
+    public function down(Schema $schema)
+    {
+        $schema->dropTable('productions');
+    }
+}
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -3,6 +3,8 @@
 use Broadway\CommandHandling\CommandBusInterface;
 use Broadway\EventHandling\EventBusInterface;
 use CultuurNet\Broadway\EventHandling\ReplayFlaggingEventBus;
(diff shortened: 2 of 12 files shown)


app/Migrations/Version20200602134547.php
```diff
@@ -0,0 +1,53 @@
...
+        $table->addColumn(
+            'added_at',
+            'date_immutable',
+            [
+                'notnull' => true,
+            ]
+        );
+
+        $table->setPrimaryKey(['event_id']);
+        $table->addIndex(['production_id']);
```
bertramakers, line 46: Thinking a bit ahead about the search functionality on name, should we also add an index on `name`?
The `MATCH()` function looks interesting: https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html

bertramakers: Or if you want to add that as another migration in a later PR, that's also fine 👍

bootstrap.php
```diff
@@ -650,6 +652,7 @@ function ($app) {
         $commandBus->subscribe($app['place_geocoordinates_command_handler']);
         $commandBus->subscribe($app['event_geocoordinates_command_handler']);
         $commandBus->subscribe($app['organizer_geocoordinates_command_handler']);
+        $commandBus->subscribe($app[ProductionCommandHandler::class]);
```
bertramakers, line 655: Good thinking of using class constants as service definitions already 👍 
It's something we discussed multiple times in the past I think, but it's easy to forget once you're actually working in these files with string constants.

app/Migrations/Version20200602134547.php
jdrieghe: > Full-text indexes can be used only with InnoDB or MyISAM tables, and can be created only for CHAR, VARCHAR, or TEXT columns.

Do you know what engine we are using?

bertramakers:
A couple of high-level comments/questions:

- Personally, I see the `Production` class as a DTO that lives on the infrastructure level together with the `Repository` interface and class. So I would suggest to not use it in the command, and instead just inject the required arguments in the command directly. Otherwise all future commands (AddEventToProduction etc) could just be reduced to `SaveProduction` with a `Production` object in the end?

- Continuing with the example of `AddEventToProduction`, how would that repository method look like? Would it also require a `Production` DTO? I don't think we need to fetch the existing events to build that DTO just to add a new row for the new event. So it would probably be something like `addEvent($productionId, $productionName, $eventId)`? Or would you still use the `Production` DTO but without the other event ids inside? 

I guess what I'm trying to say is, it's a bit confusing at the moment what the `Production` DTO represents exactly. Is it the complete production with all of the events, only the events that need to be added, ...? It might become even more confusing for future operations like deleting an event from a production.

Based on these two questions, maybe it's clearer to leave out the `Production` DTO and just pass the required data directly into the command and later into the repository methods? So something like `groupEventsAsProduction($productionId, $productionName, $events)`, `addEventToProduction($productionId, $productionName, $event)`, `deleteEventFromProduction($productionId, $event)`, ... (the names are just examples here to illustrate the example)

Aside from this feedback, I think it's good work so far. 👍 This is more about the last 2% to get the code as clear and understandable as possible.

app/Migrations/Version20200602134547.php
bertramakers: Local, acceptance and test all use InnoDB. Production I'm not 100% sure, but I think we can safely assume it should be the same as acceptance and test.

jdrieghe:
@bertramakers I think you make a good point about keeping the entity out of the command. I went back and forth on that and eventually decided on the approach with the least lines of code. I'll go back and change it.

I do think we should keep the productions entity, especially for reading (e.g. listing existing productions https://jira.uitdatabank.be/browse/III-3298)

Concerning your question about adding events to an existing production, I see two solutions, depending on how much we want to invest in a pure entity storage model:

1. Work through the entity and let the repository decide which lines to add, either through querying the existing items first or by keeping track of a "dirty state" on the entity (much like "uncommitted events" on an aggregate root)

2. Have a simple `->addEventToProduction()` method on the repository that bypasses the entity while writing

bertramakers:
@jdrieghe Sounds good! Let's keep the `Production` DTO on the repository level, but try to limit it to the methods for creating a new one and reading. Then it's clear that the events inside the `Production` object always represent the complete collection and not a partial state. For adding/deleting events we can just pass the individual event id to the repository methods. For merging we only need the production ids. I think that covers all cases?

jdrieghe:
@bertramakers yep, seems complete to me. If not, there is room for some refactoring ;)

PR #464, opened 2020-06-04
Add and remove events from productions

jdrieghe:
### Added
- Command to add event to an existing production
- Command to remove event from an existing production

---
Ticket: https://jira.uitdatabank.be/browse/III-3324

### DIFF
--- a/tests/Event/Productions/ProductionCommandHandlerTest.php
+++ b/tests/Event/Productions/ProductionCommandHandlerTest.php
@@ -77,4 +77,24 @@ class ProductionCommandHandlerTest extends TestCase
         }
         $this->assertTrue($production->containsEvent($eventToAdd));
     }
+
+    /**
+     * @test
+     */
+    public function itCannotAddAnEventThatAlreadyBelongsToAnotherProduction():void
+    {
+        $eventBelongingToFirstProduction = Uuid::uuid4()->toString();
+        $name = "A Midsummer Night's Scream 2";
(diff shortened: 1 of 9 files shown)


tests/Event/Productions/ProductionCommandHandlerTest.php
```diff
@@ -97,4 +97,30 @@ public function itCannotAddAnEventThatAlreadyBelongsToAnotherProduction():void
             new AddEventToProduction($eventBelongingToFirstProduction, $secondProductionCommand->getProductionId())
         );
     }
+
+    /**
+     * @test
+     */
+    public function itCanRemoveAnEventFromAProduction(): void
```
konrad126, line 104: Could you switch it to snake_case to be consistent with other tests (or is there a reason for using camelCase here specifically?)

jdrieghe: No specific reason other than muscle memory. I'll snake case the tests here

jdrieghe: You know what, I'll do it in the follow-up PR because that one adds another test.

konrad126: Yeah ofc, it's no biggie :)

jdrieghe: Fixed here: https://github.com/cultuurnet/udb3-silex/pull/465/commits/72912f14550753adf5cdc2168ebc2a4c52c3be82

PR #466, opened 2020-06-05
Productions write api

jdrieghe:
### Added
- `/productions/` endpoint to group multiple events as a new production

---
### Todo
- [ ] Manually verify in the vagrant box that this works and respects the necessary permission requirements

### DIFF
--- a/src/Http/Productions/CreateProductionValidator.php
+++ b/src/Http/Productions/CreateProductionValidator.php
@@ -24,13 +24,13 @@ class CreateProductionValidator implements DataValidatorInterface
     {
         if (!isset($data['name'])) {
             return [
-                ['name' => 'Required but could not be found'],
+                'name' => 'Required but could not be found',
             ];
         }
 
         if (empty(trim($data['name']))) {
             return [
-                ['name' => 'Cannot be empty'],
+                'name' => 'Cannot be empty',
             ];
         }
 
@@ -41,13 +41,13 @@ class CreateProductionValidator implements DataValidatorInterface
     {
         if (!isset($data['eventIds'])) {
             return [
-                ['eventIds' => 'Required but could not be found'],
+                'eventIds' => 'Required but could not be found',
             ];
         }
 
         if (count($data['eventIds']) < 2) {
             return [
-                ['eventIds' => 'At least two events should be provided'],
+                'eventIds' => 'At least two events should be provided',
             ];
         }
 
--- a/src/Http/Productions/CreateProductionValidator.php
+++ b/src/Http/Productions/CreateProductionValidator.php
@@ -7,8 +7,42 @@ use CultuurNet\UDB3\Http\Deserializer\DataValidator\DataValidatorInterface;
 
 class CreateProductionValidator implements DataValidatorInterface
 {
-    public function validate(array $data)
+    public function validate(array $data): void
     {
-        throw new DataValidationException();
+        $messages = [];
+        $messages = array_merge($messages, $this->validateName($data));
+        $messages = array_merge($messages, $this->validateEvents($data));
+
+        if (!empty($messages)) {
+            $exception = new DataValidationException();
+            $exception->setValidationMessages($messages);
+            throw $exception;
+        }
+    }
+
+    private function validateName(array $data): array
+    {
+        if (!isset($data['name'])) {
+            return ["Property 'name' is required"];
+        }
+
+        if (empty(trim($data['name']))) {
+            return ["Property 'name' cannot be empty"];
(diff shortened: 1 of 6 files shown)


src/Http/Productions/ProductionsWriteController.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->commandBus = $commandBus;
+    }
+
+    public function create(Request $request): Response
+    {
+        $data = json_decode($request->getContent(), true);
+
+        $command = new GroupEventsAsProduction(
+            $data['eventIds'],
+            $data['name']
```
bertramakers, line 28: We should add some validation for this deserialization of JSON.

Here's a good example of how we do that when creating a new organizer:
- A class for the payload as an object: https://github.com/cultuurnet/udb3-silex/blob/master/src/Http/Deserializer/Organizer/OrganizerCreationPayload.php
- A data validator class: https://github.com/cultuurnet/udb3-silex/blob/master/src/Http/Deserializer/Organizer/OrganizerCreationPayloadDataValidator.php (This is something custom right now. We should start using something like respect instead in the future, but let's maybe stick to this approach for now so as not to widen the scope of this ticket too much and to maintain some consistency)
- A deserializer to convert the json to the payload class: https://github.com/cultuurnet/udb3-silex/blob/master/src/Http/Deserializer/Organizer/OrganizerCreationPayloadJSONDeserializer.php (which uses the data validator internally)

Additionally, this JSON payload is basically the JSON representation of the command itself. So what you could do is _not_ have a class for the payload, and just have the deserializer return the command instead.

If you do that, you could either dispatch the command here manually if you like, or use https://github.com/cultuurnet/udb3-silex/blob/master/src/Http/CommandDeserializerController.php instead. Here's an example how we use those for the exports: https://github.com/cultuurnet/udb3-silex/blob/master/app/Export/ExportControllerProvider.php#L14

Although I'm also a fan of having it a bit more explicit, so having a `ProductionsWriteController::create()` which deserializes the JSON to a command and dispatches it instead of re-using `CommandDeserializerController` is also totally okay for me. Maybe check how often we actually use `CommandDeserializerController` and if it's very little, do keep it explicit like you do right now.

Even having a deserializer is no strict requirement for me in this scenario. It's nice for payloads that share a lot of complex properties like calendar info etc but in this case, it's a pretty straightforward 1-1 conversion.

Data validation is the only must-have I think, which you can also achieve by injecting the validator directly.

bertramakers:
Aside from adding validation this looks good to me 👍 
I like that you had small PRs for each repository method, and can now do small PRs for the HTTP layer.

src/Http/Productions/CreateProductionValidator.php
```diff
@@ -0,0 +1,56 @@
...
+            throw $exception;
+        }
+    }
+
+    private function validateName(array $data): array
+    {
+        if (!isset($data['name'])) {
+            return [
+                ['name' => 'Required but could not be found'],
+            ];
```
bertramakers, line 28: Shouldn't this be `return ['name' => 'Required but could not be found'];`?
So the errors would be printed in JSON as:
```
{
  name: "Required but could not be found",
  eventIds: "Required but could not be found"
}
```
Instead of
```
[
  {
    name: "Required but could not be found",
  },
  {
    eventIds: "Required but could not be found"
  }
]
```

tests/Http/Productions/CreateProductionValidatorTest.php
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * @test
+     * @dataProvider invalidData
+     */
+    public function it_fails_on_invalid_data(array $data): void
+    {
+        $this->expectException(DataValidationException::class);
+        $this->validator->validate($data);
```
bertramakers, line 45: Maybe assert the messages in the `DataValidationException` as well to make sure they're correctly structured, as it seems there's an issue with them.

src/Http/Productions/CreateProductionValidator.php
bertramakers: Here's an example from another validator inside a test: https://github.com/cultuurnet/udb3-silex/blob/master/tests/Http/Deserializer/Address/AddressJSONDeserializerTest.php#L34-L41

bertramakers:
Aside from a minor issue with the validation messages the last changes look promising 👍

src/Http/Productions/CreateProductionValidator.php
jdrieghe: Woah, good spot !

PR #468, opened 2020-06-15
Add route to remove an event from an existing production

jdrieghe:
### Added
- Route to remove an event from an existing production: `DELETE /{productionId}/events/{eventId}`

### DIFF
--- a/src/Http/Productions/ProductionsWriteController.php
+++ b/src/Http/Productions/ProductionsWriteController.php
@@ -57,7 +57,7 @@ class ProductionsWriteController
 
         $this->commandBus->dispatch($command);
 
-        return new Response('', 201);
+        return new Response('', 200);
     }
 
     public function removeEventFromProduction(
@@ -71,6 +71,6 @@ class ProductionsWriteController
 
         $this->commandBus->dispatch($command);
 
-        return new Response('', 201);
+        return new Response('', 200);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Http/Productions/ProductionsWriteController.php
```diff
@@ -56,6 +57,20 @@ public function addEventToProduction(
 
         $this->commandBus->dispatch($command);
 
-        return new Response('', 201);
+        return new Response('', 200);
```
bertramakers, line 60: Should it actually be `204` (`No content`)?

jdrieghe: Yes, good point. I'll fix it for all routes in the next PR to avoid unnecessary merges

tests/Http/Productions/ProductionsWriteControllerTest.php
```diff
@@ -101,6 +102,22 @@ function (AddEventToProduction $command) use ($productionId, $eventId) {
...
     }
 
+    /**
+     * @test
+     */
+    public function it_can_remove_an_event_from_an_existing_production(): void
+    {
+        $productionId = ProductionId::generate();
+        $eventId = Uuid::uuid4();
+        $this->commandBus->expects($this->once())->method('dispatch')->willReturnCallback(
```
bertramakers, line 112: Btw Broadway has a `TraceableCommandBus` for testing purposes. You have to set it up as:
```
$this->commandBus = new Broadway\CommandHandling\Testing\TraceableCommandBus();
$this->commandBus->record();
```
And then in your test you can just do `$this->commandBus->getRecordedCommands()`.

bertramakers: My bad for not mentioning this in earlier PRs 😉

jdrieghe: That's probably going to make the tests a lot more readable. Will change it in the follow-up PR.

bertramakers:
Some minor suggestions but no big issues so I'm approving already.

PR #469, opened 2020-06-16
Endpoint to merge existing productions

jdrieghe:
### Added
- Endpoint to merge existing productions: `/productions/{toProductionId}/merge/{fromProductionId}`

---
~Merge https://github.com/cultuurnet/udb3-silex/pull/468 first and change base to `master`~

### DIFF
--- a/tests/Event/Productions/ProductionCommandHandlerTest.php
+++ b/tests/Event/Productions/ProductionCommandHandlerTest.php
@@ -42,7 +42,7 @@ class ProductionCommandHandlerTest extends TestCase
             Uuid::uuid4()->toString(),
         ];
 
-        $command = new GroupEventsAsProduction($events, $name);
+        $command = GroupEventsAsProduction::withProductionName($events, $name);
         $this->commandHandler->handle($command);
 
         $createdProduction = $this->productionRepository->find($command->getProductionId());
@@ -64,7 +64,7 @@ class ProductionCommandHandlerTest extends TestCase
             Uuid::uuid4()->toString(),
         ];
 
-        $command = new GroupEventsAsProduction($events, $name);
+        $command = GroupEventsAsProduction::withProductionName($events, $name);
         $this->commandHandler->handle($command);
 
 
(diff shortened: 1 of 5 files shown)


tests/Http/Productions/ProductionsWriteControllerTest.php
```diff
@@ -59,14 +59,13 @@ public function it_can_create_production(): void
...
-                $this->assertEquals($expectedEvents, $command->getEventIds());
-            }
-        );
+        $this->commandBus->record();
         $this->controller->create($request);
+
+        $recordedCommand = $this->commandBus->getRecordedCommands()[0];
+        $this->assertInstanceOf(GroupEventsAsProduction::class, $recordedCommand);
+        $this->assertEquals($name, $recordedCommand->getName());
+        $this->assertEquals([$eventId1, $eventId2], $recordedCommand->getEventIds());
```
bertramakers, line 68: Nit: Should've mentioned it in the previous review, but I was always taught it was better to re-create the object you'd expect and then compare it using `assertEquals()` whenever possible, without getters. So you're effectively comparing the inner state.

Also if you do it like:
```
$expectedCommands = [new GroupEventsAsProduction(...)]; // Note the array
$this->assertEquals($expectedCommands, $recordedCommands);
```
You can verify that no extra commands were dispatched.

But it's definitely nitpicking and good enough as-is, so I'm approving either way.

bertramakers: As a compromise, you could also assert the count of the `$recordedCommands` being `1` of course. I think the biggest advantage of building an `$expected` array/object is just readability and less cognitive load.

tests/Event/Productions/ProductionCommandHandlerTest.php
```diff
@@ -132,12 +132,12 @@ public function it_will_not_remove_events_from_another_production()
...
         $eventBelongingToFirstProduction = Uuid::uuid4()->toString();
         $name = "A Midsummer Night's Scream 2";
-        $firstProductionCommand = new GroupEventsAsProduction([$eventBelongingToFirstProduction], $name);
+        $firstProductionCommand = GroupEventsAsProduction::withProductionName([$eventBelongingToFirstProduction], $name);
         $this->commandHandler->handle($firstProductionCommand);
 
         $eventBelongingToSecondProduction = Uuid::uuid4()->toString();
         $name = "A Midsummer Night's Scream 3";
-        $secondProductionCommand = new GroupEventsAsProduction([$eventBelongingToSecondProduction], $name);
+        $secondProductionCommand = GroupEventsAsProduction::withProductionName(h[$eventBelongingToSecondProduction], $name);
```
bertramakers, line 140: Typo? (`withProductionName(h[`)

PR #474, opened 2020-07-02
Skip suggested similar events

konrad126:
### Added
- endpoint for skipping events

### Changed
- some minor bugs fixed


src/Event/Productions/Tuple.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Event\Productions;
+
+class Tuple
```
jdrieghe, line 5: Should we call this `EventPair`? 
`Tuple` is a structural name, similar to `String` or `Array`. You wouldn't call a VO `string` without giving it more domain meaning, so I wouldn't use `tuple` either.

src/Event/Productions/ProductionRepository.php
```diff
@@ -151,4 +151,33 @@ public function findProductionForEventId(string $eventId): Production
...
     }
+
+    /**
+     * @param string $forEventId
+     * @param ProductionId $inProductionId
+     *
+     * @return Tuple[]
+     * @throws EntityNotFoundException
+     */
+    public function findTuples(string $forEventId, ProductionId $inProductionId): array
```
jdrieghe, line 162: Similar to the remark here (https://github.com/cultuurnet/udb3-silex/pull/474/commits/6ff814af96371417e6be6c0d500cbc11fb095b2f#r449185330), I would rename this method to `findEventPairs()`

src/Event/Productions/ProductionRepository.php
```diff
@@ -151,4 +151,33 @@ public function findProductionForEventId(string $eventId): Production
...
+    {
+        $results = $this->getConnection()->fetchAll(
+            'SELECT * FROM productions WHERE production_id = :productionId AND event_id <> :eventId',
```
jdrieghe, line 165: Implementation detail: I prefer the `!=` syntax over `<>`, especially for non-numerical values like uuids

app/Event/ProductionControllerProvider.php
```diff
@@ -40,6 +40,8 @@ function (Application $app) {
         $controllers->delete('/{productionId}/events/{eventId}', ProductionsWriteController::class . ':removeEventFromProduction');
         $controllers->post('/{productionId}/merge/{fromProductionId}', ProductionsWriteController::class . ':mergeProductions');
 
+        $controllers->post('/skipp', ProductionsWriteController::class . ':skipEvents');
```
jdrieghe, line 43:
```suggestion
        $controllers->post('/skip', ProductionsWriteController::class . ':skipEvents');
```

src/Event/Productions/ProductionRepository.php
jdrieghe: We should also add a test for this repository method.

src/Event/Productions/SimilaritiesClient.php
```diff
@@ -0,0 +1,58 @@
...
+    private $key;
+
+    public function __construct(Client $client, string $uri, string $key)
+    {
+        $this->client = $client;
+        $this->uri = $uri;
+        $this->key = $key;
+    }
+
+    public function markAsLinked(array $tuples): void
```
jdrieghe, line 31: Since this is the api client for GCP, we should probably use their terminology in our methods.
The way I think about this, we perform a link action which results in multiple event pairs being linked in a production. As a side effect, we call `->excludeTemporarily()` on the Similarities API.

Similarly, when we perform a `skip` action, an event pair is skipped and we should call an `->excludePermanently()` on the Similarities API.

src/Event/Productions/ProductionRepository.php
konrad126: Done (and I've caught a bug that I missed, testing FTW)

tests/Event/Productions/ProductionRepositoryTest.php
```diff
@@ -155,4 +189,10 @@ private function givenThereIsAProduction(string $name = 'foo'): Production
 
         return $production;
     }
+
+
+    private function givenThereAreEventPairs(string $eventId, string $eventid)
```
jdrieghe, line 194: Unused method?

bootstrap.php
```diff
@@ -673,10 +674,19 @@ function ($app) {
         return new ProductionRepository($app['dbal_connection']);
     }
 );
+$app[SimilaritiesClient::class] = $app->share(
+    function ($app) {
+        return new SimilaritiesClient(
+            new \GuzzleHttp\Client(),
+            'https://udb-similarities-proxy-test-nu5eme4wwa-ew.a.run.app/v1/events',
```
bertramakers, line 681: Reminder to make this configurable

composer.lock
```diff
@@ -1447,12 +1447,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/cultuurnet/udb3-php.git",
-                "reference": "ac0db0f0482ce138ab6ebbfa72215453244233b8"
+                "reference": "4ecbc13064bf28c406ae6800766995f9998efec7"
```
bertramakers, line 1450: If this has only been updated _now_, do we actually even check the permissions when creating/updating productions? cc @jdrieghe ?

bertramakers: (This commit is the merge of the new permission https://github.com/cultuurnet/udb3-php/commit/4ecbc13064bf28c406ae6800766995f9998efec7)

config.yml.dist
```diff
@@ -182,3 +182,6 @@ bookable_event:
 
 # For development only
 bypass_api_key_check: false
+
+productions:
+  key:
```
bertramakers, line 187: Can you make this more descriptive? Maybe change it to something like:
```
event_similarities_api:
  base_url:
  api_key:
```

src/Event/Productions/EventPair.php
```diff
@@ -0,0 +1,39 @@
+<?php declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Event\Productions;
+
+class EventPair
```
bertramakers, line 5: It's definitely better than `Tuple` and I can live with `EventPair`, but maybe something like `SuggestedMatch` would even be clearer? Or `SimilarEventPair` to build on the "similarities" API name?

src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -52,6 +61,45 @@ public function handleRemoveEventFromProduction(RemoveEventFromProduction $comma
...
+        Production $from,
+        Production $to
+    ) {
+        $eventPairs = [];
+        foreach ($from->getEventIds() as $eventIdFrom) {
+            foreach ($to->getEventIds() as $eventIdTo) {
+                $eventPairs[] = new EventPair($eventIdFrom, $eventIdTo);
+            }
+        }
+        $this->similaritiesClient->excludeTemporarily($eventPairs);
```
bertramakers, line 103: Suggestion (also sent you this on Slack but posting it here for reference):
Maybe we should move this to a decorator?

src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -52,6 +61,45 @@ public function handleRemoveEventFromProduction(RemoveEventFromProduction $comma
...
     {
         $toProduction = $this->productionRepository->find($command->getTo());
+        $fromProduction = $this->productionRepository->find($command->getFrom());
+
         $this->productionRepository->moveEvents($command->getFrom(), $toProduction);
+
+        $this->markNewMergedPairsAsLinked(
+            $fromProduction,
+            $toProduction
+        );
```
bertramakers, line 71: Suggestion (also sent you this on Slack but posting it here for reference):
Maybe we should move this functionality of excluding the pairs to a decorator around `ProductionCommandHandler::handleMergeProductions()`?

src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -52,6 +61,45 @@ public function handleRemoveEventFromProduction(RemoveEventFromProduction $comma
...
+    }
+
+    public function handleSkipEvents(SkipEvents $command): void
+    {
+        $this->similaritiesClient->excludePermanently(EventPair::fromArray($command->getEventIds()));
```
bertramakers, line 76: I'm not sure if skipping suggestions should be a command. It's not really part of the core productions functionality inside the UDB3 backend but more of a necessity to make an external system work. Maybe this could just go directly inside the controller for the `skip` endpoint?

bertramakers: @jdrieghe what is your take on this?

src/Event/Productions/ProductionRepository.php
```diff
@@ -57,7 +57,7 @@ public function addEvent(string $eventId, Production $production)
                 'event_id' => $eventId,
                 'production_id' => $production->getProductionId()->toNative(),
                 'name' => $production->getName(),
-                'added_at' => $addedAt->format(DATE_ATOM),
+                'added_at' => $addedAt->format('Y-m-d'),
```
bertramakers, line 60: Why were these changed? We're losing some info now (exact time).

src/Event/Productions/ProductionRepository.php
```diff
@@ -151,4 +151,36 @@ public function findProductionForEventId(string $eventId): Production
...
+                            WHERE production_id = :productionId 
+                            AND event_id = :eventId)',
+            [
+                'productionId' => $inProductionId->toNative(),
+                'eventId' => $forEventId,
+            ]
+        );
+
+        if (!$results) {
+            throw new EntityNotFoundException('No Tuples found');
```
bertramakers, line 176:
```suggestion
            throw new EntityNotFoundException('No event pairs found');
```

src/Event/Productions/SimilaritiesClient.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->client = $client;
+        $this->uri = $uri;
+        $this->key = $key;
+    }
+
+    public function excludeTemporarily(array $eventPairs): void
+    {
+        $data['pairs'] = [];
+        /** @var EventPair $pair */
```
bertramakers, line 34: Nit: Would probably be useful to add a docblock above the method for this, and remove the inline comment. That way it's clear for people looking at the signature of the method that the array should contain `EventPair` objects

src/Http/Productions/ProductionsWriteController.php
```diff
@@ -22,13 +23,19 @@ class ProductionsWriteController
      * @var CreateProductionValidator
      */
     private $createProductionValidator;
+    /**
```
bertramakers, line 26:
```suggestion

    /**
```

(Missing a newline above the docblock here)

src/Http/Productions/ProductionsWriteController.php
```diff
@@ -88,4 +95,20 @@ public function mergeProductions(
...
+        $data = (array)json_decode($request->getContent(), true);
+
+        $this->skipEventsValidator->validate($data);
+
+        $this->commandBus->dispatch(
+            new SkipEvents(
+                $data['eventIds']
+            )
+        );
+
```
bertramakers, line 110:
```suggestion
```

(One newline too many here)

tests/Event/Productions/ProductionRepositoryTest.php
```diff
@@ -105,7 +105,7 @@ public function it_can_move_events_to_other_production(): void
         $this->repository->moveEvents($fromProduction->getProductionId(), $toProduction);
 
         $resultingProduction = $this->repository->find($toProduction->getProductionId());
-        foreach($eventsToMove as $eventToMove) {
+        foreach ($eventsToMove as $eventToMove) {
```
bertramakers, line 108: Nice find 👍

tests/Event/Productions/ProductionRepositoryTest.php
```diff
@@ -145,6 +178,7 @@ public function it_will_throw_if_it_cannot_find_production_for_event(): void
         $this->repository->findProductionForEventId($randomEventId);
     }
 
+
```
bertramakers, line 181:
```suggestion
```

(Unnecessary addition of a second newline)

tests/Event/Productions/ProductionRepositoryTest.php
```diff
@@ -155,4 +189,5 @@ private function givenThereIsAProduction(string $name = 'foo'): Production
 
         return $production;
     }
+
```
bertramakers, line 192:
```suggestion
```

(Don't think we add a newline before closing a class)

tests/Http/Role/ReadRoleRestControllerTest.php
```diff
@@ -296,7 +296,8 @@ public function it_returns_an_array_of_permissions_for_the_current_god_user()
             "GEBRUIKERS_BEHEREN",
             "LABELS_BEHEREN",
             "MEDIA_UPLOADEN",
-            'VOORZIENINGEN_BEWERKEN'
+            'VOORZIENINGEN_BEWERKEN',
+            'PRODUCTIES_AANMAKEN'
```
bertramakers, line 300:
```suggestion
            'PRODUCTIES_AANMAKEN',
```
(This used to be a coding standards check but it seems it was removed a while back)

src/Event/Productions/ProductionCommandHandler.php
jdrieghe: Agreed. I think I suggested earlier to have a skip endpoint that automatically returns the next pair, making the permanent exclusion from GCP a part of that implementation because otherwise, we would get the same pair again.

composer.lock
jdrieghe: We check the permissions in the command handling if I'm not mistaken, but the tests don't go through that flow.

src/Event/Productions/EventPair.php
jdrieghe: I like `SimilarEventPair`

src/Event/Productions/ProductionCommandHandler.php
jdrieghe: Or as an event listener reacting to `ProductionsWereMerged`?

src/Event/Productions/ProductionRepository.php
konrad126: Because it would break. DB would throw an error

jdrieghe: How come this didn't fail in the unit tests in that case?

jdrieghe:
@bertramakers I think it would make sense to create a small PR for the code style remarks that enforces these rules (and maybe even automatically fixes them through `phpcbf`?

src/Event/Productions/ProductionCommandHandler.php
konrad126: I don't see it as something that's not part of the core functionality of Productions. Whatever system we would use, internal or external to provide suggestions we would need a way of "rejecting" certain suggestions. 

But, regardless of that, I would still keep it in command for architectural consistency. If we are using the command pattern I would expect that all Use Cases/ Actions in an app are represented with a command

src/Event/Productions/ProductionCommandHandler.php
bertramakers: Ah, I didn't know/remember there was an event for that

konrad126: I would rather go with an event/event listener. But in any case, I would rather refactor this in later PRs because I would like to focus on implementing the rest of the API endpoints and integrating them with GCP asap

src/Event/Productions/ProductionRepository.php
konrad126: Could be because test use sqlite.
This is what I get (when pinging the endpoint so it's trying to persist to MySQL)
```
{
    "title": "An exception occurred while executing 'INSERT INTO productions (event_id, production_id, name, added_at) VALUES (?, ?, ?, ?)' with params [\"44e35903-0dc0-4cfd-8a1e-764a9f5436ba\", \"d8fcb590-79c1-49ed-afb8-81b0909eb443\", \"Marsonia\", \"2020-07-06T11:04:57+02:00\"]:\n\nSQLSTATE[22007]: Invalid datetime format: 1292 Incorrect date value: '2020-07-06T11:04:57+02:00' for column 'added_at' at row 1",
    "type": "about:blank",
    "status": 400
}
```

src/Event/Productions/ProductionCommandHandler.php
konrad126: Since there is actually not much endpoints left (one) I could also tackle this refactoring first :)

jdrieghe: Let's maybe tackle it in a follow-up so we can split up the work?

src/Event/Productions/ProductionRepository.php
jdrieghe: Makes sense. We're using the `date_immutable` type from Doctrine DBAL here. I don't think we ever wanted to have more granularity than a single day. The field is mainly here to e.g. only search for "recent" productions.

src/Event/Productions/SimilarEventPair.php
```diff
@@ -0,0 +1,49 @@
...
+    public static function fromArray(array $eventIds) : SimilarEventPair
+    {
+        if (!array_key_exists(0, $eventIds) || !array_key_exists(1, $eventIds)) {
+            throw new \InvalidArgumentException();
+        }
+
+        return new self($eventIds[0], $eventIds[1]);
+    }
+
+    public function asArray(): array
```
jdrieghe, line 32: I think we can remove this method as it's only used in a unit test. We would need to adapt the unit test maybe to verify deserialization?

src/Event/Productions/ProductionCommandHandler.php
jdrieghe: I like your suggestion of the word "reject". We should model it in the domain as `RejectSuggestedEventPair($eventPair)`

src/Event/Productions/SimilarEventPair.php
jdrieghe: I'll remove the method in the follow-up refactor

PR #476, opened 2020-07-07
Clean up and fix production command handling

jdrieghe:
### Changed
- Renamed methods that deal with a side-effect to better clarify their reactive nature. We decided not to implement a full event+listener flow due to the lack of simple event bus in UDB3 that is not forcing the Broadway event structure.
- `RejectSuggestedEventPair` now uses the `SimilarEventPair` VO instead of relying on a potentially faulty array of strings.

### Removed
- Unused method

### Fixed
- Trying to create a new production with an event that already belongs to a production now throws the correct domain exception instead of a unique constraint error on database level.

### DIFF
--- a/src/Event/Productions/EventCannotBeAddedToProduction.php
+++ b/src/Event/Productions/EventCannotBeAddedToProduction.php
@@ -9,14 +9,14 @@ final class EventCannotBeAddedToProduction extends Exception
     public static function becauseItAlreadyBelongsToAnotherProduction(string $eventId, ProductionId $productionId)
     {
         return new self(
-            'Event with id ' . $eventId . ' cannot be added to production with id ' . $productionId->toNative()
+            'Event with id ' . $eventId . ' cannot be added to production with id ' . $productionId->toNative() . ' because it already belongs to another production.'
         );
     }
 
     public static function becauseTheyAlreadyBelongToAnotherProduction(array $eventIds, ProductionId $productionId)
     {
         return new self(
-            'Events with id ' . join(',', $eventIds) . ' cannot be added to production with id ' . $productionId->toNative()
+            'Events with ids ' . join(',', $eventIds) . ' cannot be added to production with id ' . $productionId->toNative() . ' because they already belong to another production.'
         );
     }
 }
--- a/src/Event/Productions/EventCannotBeAddedToProduction.php
+++ b/src/Event/Productions/EventCannotBeAddedToProduction.php
@@ -9,7 +9,14 @@ final class EventCannotBeAddedToProduction extends Exception
     public static function becauseItAlreadyBelongsToAnotherProduction(string $eventId, ProductionId $productionId)
     {
         return new self(
-            'Event with id ' . $eventId . ' can not be added to production with id ' . $productionId->toNative()
+            'Event with id ' . $eventId . ' cannot be added to production with id ' . $productionId->toNative()
+        );
+    }
+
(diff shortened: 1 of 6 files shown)


src/Event/Productions/EventCannotBeAddedToProduction.php
```diff
@@ -9,7 +9,14 @@ final class EventCannotBeAddedToProduction extends Exception
...
         return new self(
-            'Event with id ' . $eventId . ' can not be added to production with id ' . $productionId->toNative()
+            'Event with id ' . $eventId . ' cannot be added to production with id ' . $productionId->toNative()
+        );
+    }
+
+    public static function becauseTheyAlreadyBelongToAnotherProduction(array $eventIds, ProductionId $productionId)
+    {
+        return new self(
+            'Events with id ' . join(',', $eventIds) . ' cannot be added to production with id ' . $productionId->toNative()
```
bertramakers, line 19: Nit: Should it be `Events with ids` (plural)?

Also, should we add an explanation at the end cfr the method names? E.g. `Events with ids ... cannot be added to production with id ... because they already belong to another production`?

src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -76,12 +83,7 @@ public function handleRejectSuggestedEventPair(RejectSuggestedEventPair $command
         $this->similaritiesClient->excludePermanently(SimilarEventPair::fromArray($command->getEventIds()));
     }
 
-    /** @param string $eventId
-     * @param ProductionId $productionId
-     * @throws \GuzzleHttp\Exception\GuzzleException
-     * @todo: move logic to event/event handler
-     */
-    private function markAsLinked(string $eventId, ProductionId $productionId): void
+    private function eventsWereAddedToProduction(string $eventId, ProductionId $productionId): void
```
bertramakers, line 86: Good idea to use event names for these methods as a workaround 👍

src/Event/Productions/EventCannotBeAddedToProduction.php
jdrieghe: Fixed

bertramakers: Sorry to re-visit this again, but is it because they (_both_) belong to another production, or because at least one of them already belongs to another production? If it's the latter, the method name and message might be confusing. If it's the first one it's all good 👍

jdrieghe:
Retriggered CI due to a random segfault

src/Event/Productions/EventCannotBeAddedToProduction.php
jdrieghe: It's indeed the latter, but it could be both the events. It's hard to know with the current flow which one is the problem. Do you think we should make the exception message and method name even more explicit?

bertramakers: I'd suggest making it more descriptive but not include whether it's specifically one or both, because I can see why that would be hard yes. So something like `becauseOneOrBothAlreadyBelongToAnotherProduction`? Or `becauseTheyOverlapWithAnotherProduction`? Just some ideas to give you inspiration, doesn't have to be either if you can think of something else that you like better.

But at this point it's the last 1% so if you disagree or want to move on, feel free to just merge.

jdrieghe: Was already fixing it and I think I found a clean description using `some`

bertramakers:
Last commits also look good to me 👍 
Thanks for the extra effort @jdrieghe !

PR #478, opened 2020-07-08
Feature:III-3324

konrad126:
### Added
- endpoint for productions suggestions

---
Ticket:
[Jira](https://jira.uitdatabank.be/browse/III-3324)

### DIFF
--- a/app/Event/ProductionControllerProvider.php
+++ b/app/Event/ProductionControllerProvider.php
@@ -2,7 +2,6 @@
 
 namespace CultuurNet\UDB3\Silex\Event;
 
-use CultuurNet\UDB3\Event\Productions\ProductionEnrichedEventRepository;
 use CultuurNet\UDB3\Event\Productions\ProductionRepository;
 use CultuurNet\UDB3\Event\Productions\SimilaritiesClient;
 use CultuurNet\UDB3\Http\Productions\CreateProductionValidator;
--- a/app/Event/ProductionControllerProvider.php
+++ b/app/Event/ProductionControllerProvider.php
@@ -2,6 +2,8 @@
 
 namespace CultuurNet\UDB3\Silex\Event;
 
+use CultuurNet\UDB3\Doctrine\ReadModel\CacheDocumentRepository;
+use CultuurNet\UDB3\Event\Productions\ProductionEnrichedEventRepository;
 use CultuurNet\UDB3\Event\Productions\ProductionRepository;
 use CultuurNet\UDB3\Event\Productions\RemoveEventFromProduction;
 use CultuurNet\UDB3\Event\Productions\SimilaritiesClient;
@@ -39,7 +41,8 @@ class ProductionControllerProvider implements ControllerProviderInterface
         $app[ProductionSuggestionController::class] = $app->share(
             function (Application $app) {
                 return new ProductionSuggestionController(
-                    $app[SimilaritiesClient::class]
+                    $app[SimilaritiesClient::class],
+                    $app[ProductionEnrichedEventRepository::class]
                 );
             }
         );
--- a/app/Event/ProductionControllerProvider.php
+++ b/app/Event/ProductionControllerProvider.php
@@ -2,10 +2,8 @@
 
 namespace CultuurNet\UDB3\Silex\Event;
 
-use CultuurNet\UDB3\Doctrine\ReadModel\CacheDocumentRepository;
 use CultuurNet\UDB3\Event\Productions\ProductionEnrichedEventRepository;
 use CultuurNet\UDB3\Event\Productions\ProductionRepository;
-use CultuurNet\UDB3\Event\Productions\RemoveEventFromProduction;
 use CultuurNet\UDB3\Event\Productions\SimilaritiesClient;
 use CultuurNet\UDB3\Http\Productions\CreateProductionValidator;
 use CultuurNet\UDB3\Http\Productions\ProductionsSearchController;
--- a/app/Event/ProductionControllerProvider.php
+++ b/app/Event/ProductionControllerProvider.php
@@ -40,7 +40,7 @@ class ProductionControllerProvider implements ControllerProviderInterface
             function (Application $app) {
                 return new ProductionSuggestionController(
                     $app[SimilaritiesClient::class],
-                    $app[ProductionEnrichedEventRepository::class]
+                    $app['event_jsonld_repository']
                 );
             }
         );
--- a/app/Event/ProductionControllerProvider.php
+++ b/app/Event/ProductionControllerProvider.php
@@ -4,8 +4,10 @@ namespace CultuurNet\UDB3\Silex\Event;
 
 use CultuurNet\UDB3\Event\Productions\ProductionRepository;
 use CultuurNet\UDB3\Event\Productions\RemoveEventFromProduction;
+use CultuurNet\UDB3\Event\Productions\SimilaritiesClient;
 use CultuurNet\UDB3\Http\Productions\CreateProductionValidator;
 use CultuurNet\UDB3\Http\Productions\ProductionsSearchController;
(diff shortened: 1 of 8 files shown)


bertramakers:
Oh wait I didn't notice that this also adds a new endpoint for the suggestions. I'll review that tomorrow!

app/Event/ProductionControllerProvider.php
```diff
@@ -34,6 +36,15 @@ function (Application $app) {
             }
         );
 
+        $app[ProductionSuggestionController::class] = $app->share(
+            function (Application $app) {
+                return new ProductionSuggestionController(
+                    $app[SimilaritiesClient::class],
+                    $app[ProductionEnrichedEventRepository::class]
```
bertramakers, line 43: I'd suggest to always use `event_jsonld_repository`. The various decorators around it are then more of an implementation detail. The extra `BroadcastingDocumentRepositoryDecorator` around it should be no problem, that only dispatches an event on the event bus when the JSON-LD gets _saved_. When fetching the JSON-LD it does nothing that could cause issues, and the performance overhead should be negligable.

bertramakers: Additionally, you can then move the `ProductionEnrichedEventRepository::class` definition back inside the `event_jsonld_repository`. That makes it a bit easier to read, especially with multiple decorators.

app/Event/ProductionServiceProvider.php
```diff
@@ -0,0 +1,42 @@
...
+
+use CultuurNet\UDB3\Event\Productions\ProductionCommandHandler;
+use CultuurNet\UDB3\Event\Productions\ProductionRepository;
+use CultuurNet\UDB3\Event\Productions\SimilaritiesClient;
+use Silex\Application;
+use Silex\ServiceProviderInterface;
+
+class ProductionServiceProvider implements ServiceProviderInterface
+{
+
```
bertramakers, line 13:
```suggestion
```

src/Event/Productions/SimilaritiesClient.php
```diff
@@ -62,4 +63,19 @@ public function excludePermanently(SimilarEventPair $pair)
             ['json' => $data]
         );
     }
+
+    public function nextSuggestion(\DateTime $dateFrom, int $size = 1, int $offset = 0): Suggestion
```
bertramakers, line 67: Nice, a good idea to already include a `size` parameter in case we do end up fetching multiples at once down the road depending on how the performance improvements go.

src/Event/Productions/Suggestion.php
```diff
@@ -0,0 +1,32 @@
+<?php declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Event\Productions;
+
+class Suggestion
```
bertramakers, line 5: Can't we re-use the existing `SimilarEventPair` for this?

src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->enrichedEventRepository = $enrichedEventRepository;
+    }
+
+    public function nextSuggestion(): Response
+    {
+        try {
+            $date = new \DateTime();
+            $date->setTime(0, 0, 1);
+            $suggestion = $this->similaritiesClient->nextSuggestion($date);
+            $eventTwo = $this->enrichedEventRepository->get($suggestion->getEventOne());
```
bertramakers, line 38:
```suggestion
            $eventOne = $this->enrichedEventRepository->get($suggestion->getEventOne());
```

src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -0,0 +1,52 @@
...
+            $eventOne = $this->enrichedEventRepository->get($suggestion->getEventTwo());
```
bertramakers, line 39:
```suggestion
            $eventTwo = $this->enrichedEventRepository->get($suggestion->getEventTwo());
```

src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -0,0 +1,52 @@
...
+            return new JsonResponse(
+                [
+                    'events' => [
+                        $eventTwo->getBody(),
```
bertramakers, line 43:
```suggestion
                        $eventOne->getBody(),
```

src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -0,0 +1,52 @@
...
+                        $eventOne->getBody(),
```
bertramakers, line 44:
```suggestion
                        $eventTwo->getBody(),
```

src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    private $similaritiesClient;
+    /**
+     * @var ProductionEnrichedEventRepository
+     */
+    private $enrichedEventRepository;
+
+    public function __construct(
+        SimilaritiesClient $similaritiesClient,
+        ProductionEnrichedEventRepository $enrichedEventRepository
```
bertramakers, line 26: I think it would be better to depend on an interface instead of a concrete implementation?

bootstrap.php
```diff
@@ -669,26 +669,6 @@ function ($app) {
...
-            $app['config']['event_similarities_api']['api_key']
-        );
-    }
-);
-
-$app[ProductionCommandHandler::class] = $app->share(
-    function ($app) {
-        return new ProductionCommandHandler($app[ProductionRepository::class], $app[SimilaritiesClient::class]);
-    }
-);
```
bertramakers, line 691: Thanks for moving these to a service provider 👍

src/Event/Productions/Suggestion.php
konrad126: Technically we could, but semantically they are two different things. They happen to have the same technical implementation but they are two different things so I would not opt using the SimilarEventPair here to avoid confusion.

app/Event/ProductionControllerProvider.php
jdrieghe: done

src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -34,13 +34,13 @@ public function nextSuggestion(): Response
             $date = new \DateTime();
             $date->setTime(0, 0, 1);
             $suggestion = $this->similaritiesClient->nextSuggestion($date);
-            $eventTwo = $this->enrichedEventRepository->get($suggestion->getEventOne());
             $eventOne = $this->enrichedEventRepository->get($suggestion->getEventTwo());
+            $eventTwo = $this->enrichedEventRepository->get($suggestion->getEventOne());
```
bertramakers, line 38: There's still a minor issue here, which was maybe not clear from my first comment:
The variable name is `$eventTwo` but it is based on `$suggestion->getEventOne()`. Vice-versa for `$eventOne` above.

bertramakers: The rest of the changes look good, almost there! 👌

jdrieghe: Resolved and added some extra cleanup

src/Event/Productions/SimilaritiesClient.php
```diff
@@ -64,7 +65,7 @@ public function excludePermanently(SimilarEventPair $pair)
         );
     }
 
-    public function nextSuggestion(\DateTime $dateFrom, int $size = 1, int $offset = 0): Suggestion
+    public function nextSuggestion(Date $dateFrom, int $size = 1, int $offset = 0): Suggestion
```
bertramakers, line 69: We usually use `DateTimeImmutable`, but maybe in this case `Date` is better yes since we don't store the time info anyway.

PR #480, opened 2020-07-17
Fix issue with publication date in the past breaking ETL tools

jdrieghe:
### Fixed
- Publishing only allows a future publication date or it will default to now.

---
Ticket: https://jira.uitdatabank.be/browse/III-3362

### DIFF
--- a/src/Offer/Commands/Moderation/AbstractPublish.php
+++ b/src/Offer/Commands/Moderation/AbstractPublish.php
@@ -2,6 +2,8 @@
 
 namespace CultuurNet\UDB3\Offer\Commands\Moderation;
 
+use Cake\Chronos\Chronos;
+
 abstract class AbstractPublish extends AbstractModerationCommand
 {
     /** @var  \DateTimeInterface */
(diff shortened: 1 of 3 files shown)


bertramakers:
Good idea to use `Chronos::now()` so we can mock the current date 👍 

I'd prefer for the other dates in the tests (`$lastMonth`, `$futurePublicationDate`) to be `DateTimeImmutable` objects though since that's also what we pass in the actual code in controllers etc. I know chances are abysmally low that the logic would fail with `DateTimeImmutable` instead of `Chronos` dates, but who knows and it's easy to cover.

bertramakers:
Also looks like some other tests are broken now.

jdrieghe:
@bertramakers fixed the broken test.

I'm not sure if we need to change the Chronos/DateTimeImmutable discrepancy as it's already covered by the type system. The method requires a `DateTimeInterface` to be passed, so as long as we do that we should be fine, no?
Additionally, `Chronos`'s utility methods make the tests a lot more readable in my opinion: `Chronos::now()->addMonth()`

PR #482, opened 2020-08-03
III-3403 Productions backend tweaks/fixes

bertramakers:
### Added

- Added `start` parameter to productions search endpoint
- Added `limit` parameter to productions search endpoint

### Changed

- The response from the productions search endoint is now a paged collection object
- The `name` parameter is now optional, and when omitted the search endpoint will return all productions (paginated)
- Increased the length of `name` column from 32 to 255

### Fixed

- Added missing return type in `GroupEventsAsProduction`

---
Ticket: https://jira.uitdatabank.be/browse/III-3403

### DIFF
--- a/app/Migrations/Version20200804090405.php
+++ b/app/Migrations/Version20200804090405.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace CultuurNet\UDB3\Silex\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20200804090405 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table =  $schema->getTable('productions');
+
+        // Increase the max-length of the name column from 32 to 255.
+        // "Lord of the Rings: The Fellowship of the Ring" is already 45 characters for example.
+        $table->changeColumn('name', [
+            'length' => 255,
+            'notnull' => true,
+        ]);
+    }
+
+    public function down(Schema $schema): void
+    {
+        $table =  $schema->getTable('productions');
+
+        $table->changeColumn('name', [
+            'length' => 32,
+            'notnull' => true,
+        ]);
+    }
+}
--- a/src/Http/Productions/ProductionsSearchController.php
+++ b/src/Http/Productions/ProductionsSearchController.php
@@ -27,8 +27,8 @@ class ProductionsSearchController
     public function search(Request $request): Response
     {
         $keyword = $request->get('name', '');
-        $start = $request->get('start', self::DEFAULT_START);
-        $limit = $request->get('limit', self::DEFAULT_LIMIT);
+        $start = (int) $request->get('start', self::DEFAULT_START);
+        $limit = (int) $request->get('limit', self::DEFAULT_LIMIT);
         $pageNumber = (int) ($start / $limit);
 
         $count = $this->repository->count($keyword);
--- a/src/Http/Productions/ProductionsSearchController.php
+++ b/src/Http/Productions/ProductionsSearchController.php
@@ -26,7 +26,7 @@ class ProductionsSearchController
 
     public function search(Request $request): Response
     {
-        $keyword = $request->get('name');
+        $keyword = $request->get('name', '');
         $start = $request->get('start', self::DEFAULT_START);
         $limit = $request->get('limit', self::DEFAULT_LIMIT);
         $pageNumber = (int) ($start / $limit);
--- a/src/Http/Productions/ProductionsSearchController.php
+++ b/src/Http/Productions/ProductionsSearchController.php
@@ -2,6 +2,7 @@
 
 namespace CultuurNet\UDB3\Http\Productions;
 
+use CultuurNet\Hydra\PagedCollection;
 use CultuurNet\UDB3\Event\Productions\Production;
 use CultuurNet\UDB3\Event\Productions\ProductionRepository;
 use Symfony\Component\HttpFoundation\JsonResponse;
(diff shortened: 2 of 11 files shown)


bertramakers:
Swagger changes: https://github.com/cultuurnet/udb3-swagger/pull/58

bertramakers:
Created a draft PR for now because I want to test this manually first with some local test data, but can be reviewed in the meantime.

bertramakers:
Travis webhook seems stuck, here's the (green) build https://travis-ci.org/github/cultuurnet/udb3-silex/builds/714594261

src/Http/Productions/ProductionsSearchController.php
```diff
@@ -24,17 +26,39 @@ public function __construct(ProductionRepository $repository)
 
     public function search(Request $request): Response
     {
-        $keyword = $request->get('name');
+        $keyword = $request->get('name', '');
+        $start = (int) $request->get('start', self::DEFAULT_START);
+        $limit = (int) $request->get('limit', self::DEFAULT_LIMIT);
+        $pageNumber = (int) ($start / $limit);
```
jdrieghe, line 32: Won't this result in the first page number being `0`?

app/Migrations/Version20200804090405.php
```diff
@@ -0,0 +1,31 @@
...
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20200804090405 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table =  $schema->getTable('productions');
+
+        // Increase the max-length of the name column from 32 to 255.
```
jdrieghe, line 14: Oh wow, I didn't realise 255 wasn't the default 😮

src/Http/Productions/ProductionsSearchController.php
bertramakers: Yes. I was a bit confused about it as well but I copied it from https://github.com/cultuurnet/udb3-silex/blob/108511132837398c79b488fb636a9b0adba13f4f/src/Http/Label/ReadRestController.php#L110 and assumed that we just started at 0.

However, I checked and now see that in SAPI3 we do the same but increase the value by `1` so the first page number is 1: https://github.com/cultuurnet/udb3-search-service/blob/master/src/Http/ResultTransformingPagedCollectionFactory.php#L44

So we're not consistent in whether it should be 0 or 1.

Additionally,  this info isn't even actually included in the response. For example, a collection of offers from SAPI3 on acc: https://udb-silex-acc.uitdatabank.be/offers/?disableDefaultFilters=true&apiKey=deb306a6-6f46-4c98-89ce-b03ec4fd11e2

```
{
  "@context": "http://www.w3.org/ns/hydra/context.jsonld",
  "@type": "PagedCollection",
  "itemsPerPage": 30,
  "totalItems": 428540,
  "member": [...]
}
```

It's only used if we would also inject a URL generator service, so it could generate a URL to the next and previous pages: https://github.com/cultuurnet/Hydra/blob/master/src/PagedCollection.php#L178

If you want I can change it to start from 1, but maybe it would be even better to not use this `PagedCollection` object at all and just hardcode this response object's properties to start removing a dependency on https://github.com/cultuurnet/hydra?

I don't think we're actually generating those previous and next URLs in most places (e.g. here for productions, labels search, sapi3, ...)

app/Migrations/Version20200804090405.php
bertramakers: Me neither, I only noticed it when opening the table in Sequel Pro to check that the production I `POST`ed was actually inserted.

src/Http/Productions/ProductionsSearchController.php
jdrieghe: In that case I would indeed suggest to use a custom response instead and either:
- let the UI worry about generating prev and next urls
- provider `prev` and `next` URLs from the backend in the response

app/Migrations/Version20200804090405.php
jdrieghe: Good spot then 👍

src/Http/Productions/ProductionsSearchController.php
bertramakers: I added an internal `PagedCollectionResponse` with a simplified signature.

I've also refactored the two only other usages of `CultuurNet\Hydra\PagedCollection` (one in a controller, the other in the test for that controller) and removed the `cultuurnet/hydra` package from the composer dependencies.

bertramakers:
I've set up a Github action to replace the Travis check. I also removed the webhook to Travis. However I don't seem to be able to remove the Travis check from this open PR. So I'll merge this with admin rights when you have reviewed and approved @jdrieghe

src/Http/Response/PagedCollectionResponse.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Response;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+class PagedCollectionResponse extends JsonResponse
```
jdrieghe, line 9:
```suggestion
final class PagedCollectionResponse extends JsonResponse
```

src/Http/Response/PagedCollectionResponse.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function __construct(
+        int $itemsPerPage,
+        int $totalItems,
+        array $members = [],
```
jdrieghe, line 14: I'm not a big fan of `member` here as a variable name. It is supposed to mean `items`?

jdrieghe:
@bertramakers I changed the protected branch settings to no longer require the travic check and made the github action required

src/Http/Response/PagedCollectionResponse.php
bertramakers: It's supposed to reflect the `member` key in the response object. I'm going to merge this because I'm pair programming on something else but Corneel is waiting for this PR to deploy.

PR #484, opened 2020-08-07
Validate if event exists

jdrieghe:
### Fixed
- Events that are being grouped as production or added to an existing production are now validated for their existence.
- Call to Google Cloud Api to permanently exclude an event pair now works correctly

---
Ticket: https://jira.uitdatabank.be/browse/III-3324

### DIFF
--- a/app/Event/ProductionServiceProvider.php
+++ b/app/Event/ProductionServiceProvider.php
@@ -30,7 +30,11 @@ class ProductionServiceProvider implements ServiceProviderInterface
 
         $app[ProductionCommandHandler::class] = $app->share(
             function ($app) {
-                return new ProductionCommandHandler($app[ProductionRepository::class], $app[SimilaritiesClient::class]);
+                return new ProductionCommandHandler(
+                    $app[ProductionRepository::class],
+                    $app[SimilaritiesClient::class],
+                    $app{'event_jsonld_repository'}
+                );
             }
         );
     }
--- a/app/Event/ProductionServiceProvider.php
+++ b/app/Event/ProductionServiceProvider.php
@@ -33,7 +33,7 @@ class ProductionServiceProvider implements ServiceProviderInterface
                 return new ProductionCommandHandler(
                     $app[ProductionRepository::class],
                     $app[SimilaritiesClient::class],
-                    $app{'event_jsonld_repository'}
+                    $app['event_jsonld_repository']
                 );
             }
         );
--- a/src/Event/Productions/EventCannotBeAddedToProduction.php
+++ b/src/Event/Productions/EventCannotBeAddedToProduction.php
@@ -6,17 +6,24 @@ use Exception;
 
 final class EventCannotBeAddedToProduction extends Exception
 {
-    public static function becauseItAlreadyBelongsToAnotherProduction(string $eventId, ProductionId $productionId)
+    public static function becauseItAlreadyBelongsToAnotherProduction(string $eventId, ProductionId $productionId): self
     {
         return new self(
             'Event with id ' . $eventId . ' cannot be added to production with id ' . $productionId->toNative() . ' because it already belongs to another production.'
         );
     }
 
-    public static function becauseSomeEventsBelongToAnotherProduction(array $eventIds, ProductionId $productionId)
+    public static function becauseSomeEventsBelongToAnotherProduction(array $eventIds, ProductionId $productionId): self
     {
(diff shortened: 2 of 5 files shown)


src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -31,6 +41,11 @@ public function handleGroupEventsAsProduction(GroupEventsAsProduction $command):
             $command->getName(),
             $command->getEventIds()
         );
+
+        foreach ($command->getEventIds() as $eventId) {
```
LucWollants, line 45: Maybe useful to bundle multiple exceptions when multiple events don't exist?

app/Event/ProductionServiceProvider.php
```diff
@@ -30,7 +30,11 @@ function ($app) {
 
         $app[ProductionCommandHandler::class] = $app->share(
             function ($app) {
-                return new ProductionCommandHandler($app[ProductionRepository::class], $app[SimilaritiesClient::class]);
+                return new ProductionCommandHandler(
+                    $app[ProductionRepository::class],
+                    $app[SimilaritiesClient::class],
+                    $app{'event_jsonld_repository'}
```
LucWollants, line 36: On several other places I see `$app['event_jsonld_repository']`

src/Event/Productions/EventCannotBeAddedToProduction.php
```diff
@@ -6,17 +6,24 @@
...
         return new self(
             'Event with id ' . $eventId . ' cannot be added to production with id ' . $productionId->toNative() . ' because it already belongs to another production.'
         );
     }
 
-    public static function becauseSomeEventsBelongToAnotherProduction(array $eventIds, ProductionId $productionId)
+    public static function becauseSomeEventsBelongToAnotherProduction(array $eventIds, ProductionId $productionId): self
     {
         return new self(
             'Events with ids ' . join(',', $eventIds) . ' cannot be added to production with id ' . $productionId->toNative() . ' because some events already belong to another production.'
```
LucWollants, line 19: I get a warning in PHPStorm that `join` is an alias for `implode` and it is preferred to used `implode`.

tests/Event/Productions/ProductionCommandHandlerTest.php
```diff
@@ -146,8 +185,24 @@ public function it_cannot_add_an_event_that_already_belongs_to_another_productio
     /**
      * @test
      */
-    public function it_can_remove_an_event_from_aproduction(): void
+    public function it_cannot_add_a_non_existing_event_to_a_production(): void
     {
+        $eventId = Uuid::uuid4()->toString();
+        $this->eventRepository->method('get')->with(...[$eventId])->willReturn(null);
```
LucWollants, line 191: Is it not possible to just pass `$eventId`? No need to wrap and unwrap the array.

src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -104,4 +121,12 @@ private function productionsWereMerged(
         }
         $this->similaritiesClient->excludeTemporarily($eventPairs);
     }
+
+    private function assertEventExists(string $eventId)
+    {
+        $event = $this->eventRepository->get($eventId);
```
LucWollants, line 127: I think this can throw an exception? At least the interface also show a `DocumentGoneException` exception.

LucWollants: The concrete implementation `CacheDocumentRepository` checks for `GONE` returned from the cache and in that case throws a `DocumentGoneException`.

app/Event/ProductionServiceProvider.php
jdrieghe: Wow yeah, that's a nice typo there 🙈

src/Event/Productions/ProductionCommandHandler.php
jdrieghe: The UI should only give us valid events. If there are non-existing events, it's mostly because someone tampered with the request. In those situations, I don't care too much about optimising error handling. Do you think this is a must-have?

LucWollants: Is it part of the API?

jdrieghe: Yes, but the API is only privately consumed by an internal UI used by a handful of people

PR #485, opened 2020-08-24
III-3425 - Check if event can be removed from production

jdrieghe:
### Changed
- To align with adding an event to a production, we now check if an event and production exist before we try to remove it. That way, endpoint responses will be consistent across the productions api.

---
Ticket: https://jira.uitdatabank.be/browse/III-3425

### DIFF
--- a/src/Event/Productions/ProductionCommandHandler.php
+++ b/src/Event/Productions/ProductionCommandHandler.php
@@ -145,7 +145,6 @@ class ProductionCommandHandler extends Udb3CommandHandler
             if (!$production->containsEvent($eventId)) {
                 throw EventCannotBeRemovedFromProduction::becauseItDoesNotBelongToIt($eventId, $productionId);
             }
-
         } catch (DocumentGoneException $e) {
             throw EventCannotBeRemovedFromProduction::becauseItDoesNotExist($eventId);
         }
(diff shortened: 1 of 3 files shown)


src/Event/Productions/ProductionCommandHandler.php
```diff
@@ -135,4 +136,17 @@ private function assertEventExists(string $eventId)
...
         }
     }
+
+    private function assertEventCanBeRemovedFromProduction(string $eventId, ProductionId $productionId)
+    {
+        try {
+            $this->eventRepository->get($eventId);
+            $production = $this->productionRepository->find($productionId);
+            if (!$production->containsEvent($eventId)) {
+                throw EventCannotBeRemovedFromProduction::becauseItDoesNotBelongToIt($eventId, $productionId);
```
LucWollants, line 146: I wonder if this should throw. If the event does not belong to the production it could also be treated as a `no operation command`. To explain this with an example: the command to delete an event that is already deleted is also a `no operation command` and should not throw.

jdrieghe: So we could leave in the check if the event exists but make the other a no-op? Will fix.

LucWollants: But when reading the ticket I see that is what is requested by @corneelwille

LucWollants (approved):

PR #486, opened 2020-08-31
III-3451 Use LIKE instead of MATCH for productions search

bertramakers:
### Changed

- Productions search by name now uses `LIKE` (with wildcards) instead of `MATCH` to make autocomplete/typeahead possible.

---
Ticket: https://jira.uitdatabank.be/browse/III-3451

### DIFF
--- a/src/Event/Productions/ProductionRepository.php
+++ b/src/Event/Productions/ProductionRepository.php
@@ -133,7 +133,7 @@ class ProductionRepository extends AbstractDBALRepository
             ->from($this->getTableName()->toNative());
 
         if (!empty($keyword)) {
-            $query = $query->where('name LIKE CONCAT(\'%\', :keyword, \'%\')')
+            $query = $query->where('MATCH(name) AGAINST(CONCAT(:keyword, \'*\') IN BOOLEAN MODE)')
                 ->setParameter(':keyword', $keyword);
         }
 
--- a/src/Event/Productions/ProductionRepository.php
+++ b/src/Event/Productions/ProductionRepository.php
@@ -133,7 +133,7 @@ class ProductionRepository extends AbstractDBALRepository
             ->from($this->getTableName()->toNative());
 
         if (!empty($keyword)) {
-            $query = $query->where('MATCH (name) AGAINST (:keyword)')
+            $query = $query->where('name LIKE CONCAT(\'%\', :keyword, \'%\')')
                 ->setParameter(':keyword', $keyword);
         }
 
--- a/src/Event/Productions/ProductionRepository.php
+++ b/src/Event/Productions/ProductionRepository.php
@@ -133,7 +133,8 @@ class ProductionRepository extends AbstractDBALRepository
             ->from($this->getTableName()->toNative());
 
         if (!empty($keyword)) {
-            $query = $query->where('MATCH(name) AGAINST(CONCAT(:keyword, \'*\') IN BOOLEAN MODE)')
+            $keyword .= '*';
+            $query = $query->where('MATCH(name) AGAINST(:keyword IN BOOLEAN MODE)')
                 ->setParameter(':keyword', $keyword);
         }
 


jdrieghe:
Do we still need the FULLTEXT index on that field then? Not sure if that's used for `LIKE` queries?

src/Event/Productions/ProductionRepository.php
```diff
@@ -133,7 +133,7 @@ private function createSearchQuery(string $keyword): QueryBuilder
             ->from($this->getTableName()->toNative());
 
         if (!empty($keyword)) {
-            $query = $query->where('MATCH (name) AGAINST (:keyword)')
+            $query = $query->where('name LIKE CONCAT(\'%\', :keyword, \'%\')')
```
LucWollants, line 136: I wonder if it also works like this
```
$query = $query->where('name LIKE \'%' .  :keyword . '\'%')
```

That would avoid a `CONCAT` to be called for every single row.

bertramakers:
> Do we still need the FULLTEXT index on that field then? Not sure if that's used for `LIKE` queries?

If I read https://stackoverflow.com/questions/4194485/full-text-search-vs-like correctly, we do need an index (fulltext or not), but we should drop the first wildcard for performance reasons. I think that's acceptable, most cases are someone starts to type a word and we need to find matches that also start with that word.

jdrieghe:
@bertramakers That's only for MS SQL Server.

https://stackoverflow.com/questions/6695746/mysql-like-query-is-slow-with-full-text-index
> LIKE does not use the full-text index. To make use of the fulltext index, you have to use match (as you said)

src/Event/Productions/ProductionRepository.php
bertramakers: Your proposal is not syntactically valid. MySQL always uses `CONCAT` for concatenation, no? Besides, why would `CONCAT` happen for every row but your syntax not?

bertramakers:
> @bertramakers That's only for MS SQL Server.
> 
> https://stackoverflow.com/questions/6695746/mysql-like-query-is-slow-with-full-text-index
> 
> > LIKE does not use the full-text index. To make use of the fulltext index, you have to use match (as you said)

Ah you're right. I remembered reading about that `*` suffix last Friday but couldn't find it anymore and it didn't work at first. But I see it requires `BOOLEAN MODE`. Updated it!

src/Event/Productions/ProductionRepository.php
LucWollants: Indeed I made a mistake, it should have been
```
$query = $query->where('name LIKE \'%:keyword\'%');
```

MySQL does indeed use `CONCAT` to do the string concatenation but you can make use of DBAL here to make it a fixed string in which the `$keyword` variable is already replaced. So you get a where clause as `name LIKE '%movie%'`.

It's going to be called for every row because `CONCAT` is a MySQL function and the parameter you pass can be a column value from the current row. So it's default behaviour to execute it for every row even when the input parameter is a fixed string.

bertramakers: In the meantime I switched back to `MATCH` based on @jdrieghe 's comment. https://github.com/cultuurnet/udb3-silex/pull/486/commits/f7522d3b068f4cdb1cd625ab2c0740bd517410f4

I tried to change it locally to `AGAINST(:keyword* IN BOOLEAN MODE)` but that results in:
```
An exception occurred while executing 'SELECT production_id, name, GROUP_CONCAT(event_id) as events FROM productions WHERE MATCH(name) AGAINST(?* IN BOOLEAN MODE) GROUP BY production_id' with params [\"mod\"]:\n\nSQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IN BOOLEAN MODE) GROUP BY production_id' at line 1
```

As I suspected and as is visible in the error message, DBAL doesn't replace the values itself but passes an actual prepared statement to MySQL as well as the bound values. It's MySQL who then does the replacement while executing the prepared statement. And MySQL doesn't understand `:keyword*`  because it only supports concatenation using `CONCAT`. But even if MySQL understood `:keyword*`, the concatenation would then still happen per row according to your info because it happens while executing the statement, not as some kind of `str_replace` in PHP before executing the statement.

jdrieghe (approved):
Clean!

src/Event/Productions/ProductionRepository.php
bertramakers: I was able to optimize it though with a tip from @jdrieghe to do the concatenation in PHP itself first: https://github.com/cultuurnet/udb3-silex/pull/486/commits/13081f80278e5e4cdf099ebaf0d6b0e1cfa9937d

Still not sure if it was really necessary. I'd assume MySQL would be smart enough by now to know that this particular CONCAT would always be the same for each row since it would only take a prepapered statement parameter and a hardcoded string, but with this solution we're 100% sure.

bertramakers:
Thanks for the input @jdrieghe and @LucWollants, I'm also happier with this solution than a `LIKE` with a `%` wildcard! 👌

PR #488, opened 2020-09-03
Feature/iii 3463 configurable suggestion mindate

bertramakers:
### Added

- Added `min_date` option for suggestions for easier testing on the `test` environment

### Changed

- Changed some of the response codes/messages in case of errors on the suggestions endpoint

---
Ticket: https://jira.uitdatabank.be/browse/III-3463

Related PRs:
- https://github.com/cultuurnet/udb3-swagger/pull/61
- https://github.com/cultuurnet/udb3-frontend/pull/44

### DIFF
--- a/src/Http/Productions/ProductionSuggestionController.php
+++ b/src/Http/Productions/ProductionSuggestionController.php
@@ -5,7 +5,9 @@ namespace CultuurNet\UDB3\Http\Productions;
 use Cake\Chronos\Date;
 use CultuurNet\UDB3\Event\Productions\SimilaritiesClient;
 use CultuurNet\UDB3\Event\Productions\SuggestionsNotFound;
+use CultuurNet\UDB3\Event\ReadModel\DocumentGoneException;
 use CultuurNet\UDB3\Event\ReadModel\DocumentRepositoryInterface;
+use CultuurNet\UDB3\Http\HttpFoundation\NoContent;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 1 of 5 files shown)


src/Http/Productions/ProductionSuggestionController.php
```diff
@@ -21,32 +23,54 @@ class ProductionSuggestionController
...
+        ?Date $minDate = null
     ) {
         $this->similaritiesClient = $similaritiesClient;
         $this->enrichedEventRepository = $enrichedEventRepository;
+        $this->minDate = $minDate;
     }
 
     public function nextSuggestion(): Response
     {
+        $date = $this->minDate ?: Date::now();
```
jdrieghe, line 43: Why don't we do this in the constructor? That way we can type the `minDate` variable to just `Date` instead of `?Date`.

bertramakers: True, I'm a bit too careful when it comes to `Date::now()` but this is not part of a long-running process where this method would be called multiple times over a longer time period so that's a good improvement yes.

bertramakers: Your suggestion is to do it in the `ControllerProvider` and make the `Date` in the constructor required, right?

jdrieghe: You can choose to either make the argument nullable and default to now() in the constructor or always pass a date, not much difference imho.

jdrieghe (approved):

src/Http/Productions/ProductionSuggestionController.php
bertramakers: Ah, yes I see. I was a bit confused because I misunderstood and thought you meant something like this with constructor:
Screenshot 2020-09-03 at 13 31 54

But you meant this right?
Screenshot 2020-09-03 at 13 33 56

Either way the result is the same as doing it in the ControllerProvider.

PR #494, opened 2020-09-10
III-3481

jdrieghe:
### Changed
- Trying to add an event to a production to which it already belongs will now result in an exception instead of a no-operation.

---
Ticket: https://jira.uitdatabank.be/browse/III-3481


bertramakers (approved):

bertramakers:
Did a quick manual test and works like a charm! 👌

PR #495, opened 2020-09-11
Add PHPStan at level 0

jdrieghe:
### Added
- CI now runs (and enforces) PHPStan at level 0

### Fixed
- All errors exposed by static analysis at level 0


phpstan.neon
```diff
@@ -0,0 +1,8 @@
+parameters:
+	level: 0
+	paths:
+		- app
```
bertramakers, line 4: Ah I forgot this in the cdbxml service. I'll try to add it later today.
I also just put the config directly in the phing task but I see the advantage of doing it this way 👍

src/Event/ReadModel/JSONLD/RelatedEventLDProjector.php
```diff
@@ -42,6 +47,7 @@ class RelatedEventLDProjector implements EventListenerInterface
      * @param EventServiceInterface $eventService
      * @param PlaceService $placeService
      * @param OrganizerService $organizerService
+     * @param IriOfferIdentifierFactoryInterface $iriOfferIdentifierFactory
```
bertramakers, line 50: Nit: We could just remove the whole docblock, no?

bertramakers (approved):

PR #496, opened 2020-09-11
Increase required PHPStan level to 1

jdrieghe:
### Changed
- Increased phpstan level to 1

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -5,5 +5,4 @@ parameters:
 		- src
 		- tests
 	bootstrapFiles:
-		- bootstrap.php
 		- tests/bootstrap.php
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -5,4 +5,5 @@ parameters:
 		- src
 		- tests
 	bootstrapFiles:
+		- bootstrap.php
 		- tests/bootstrap.php
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-	level: 0
+	level: 1
 	paths:
 		- app
 		- src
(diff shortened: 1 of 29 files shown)


bertramakers (approved):

bertramakers:
It seems like github actions did not run for this PR?

bertramakers (changes requested):
Getting some errors in github actions (but in another PR branched from this one)

phpstan.neon
```diff
@@ -5,5 +5,4 @@ parameters:
 		- src
 		- tests
 	bootstrapFiles:
-		- bootstrap.php
```
jdrieghe: In that case, we're going to have to move the definition of
```php
define('SYSTEM_USER_UUID', '00000000-0000-0000-0000-000000000000');
```
to another location so that we can easily load it for phpstan (or replace the usages with a static call to a class or something)

bertramakers: Alternatively we might be able to copy the `config.yml.dist` file to `config.yml` in github actions before running the tests, phpstan etc? Maybe that works even if it's just with dummy values. And then we could also use that later to configure our db connection for integration tests etc.

src/User/UserIdentityDetails.php
```diff
@@ -7,6 +7,8 @@
 
 class UserIdentityDetails implements \JsonSerializable
 {
+    public const SYSTEM_USER_UUID = '00000000-0000-0000-0000-000000000000';
```
bertramakers, line 10: Good idea to move this here, seems like a good place 👍

bertramakers (approved):

PR #499, opened 2020-09-21
III-3386 Integrate Sentry to monitor fatal errors.

LucWollants:
### Added
- Log fatal errors to Sentry

---

Ticket: https://jira.uitdatabank.be/browse/III-3386


app/SentryServiceProvider.php
```diff
@@ -0,0 +1,45 @@
...
+use Sentry\State\HubInterface;
+use Silex\Application;
+use Silex\ServiceProviderInterface;
+use Throwable;
+
+class SentryServiceProvider implements ServiceProviderInterface
+{
+    public function register(Application $app)
+    {
+        $app['sentry_hub'] = $app->share(
```
bertramakers, line 18: Nit: We could use the `HubInterface::class` constant instead. We mostly used "human-readable" names in the past becaus we didn't have the `::class` constant yet back then. We've started using it recently for service definitions in some repositories. Might as well start here as well in preparation for our move away from Silex in this app as well.

bertramakers (approved):
Elegant solution, very nice! 👌

bertramakers:
Ideally we'd also have the same integration in udb3-cdbxml-service, udb3-search-service, and udb3-uitpas-service. Could you provide the same solution there? Or if you think it will take too long, maybe add a ticket to also add it to those later?

app/SentryServiceProvider.php
LucWollants: That's indeed a good tip, I changed it! Also a good one to remember when doing some boyscouting.

LucWollants:
> Ideally we'd also have the same integration in udb3-cdbxml-service, udb3-search-service, and udb3-uitpas-service. Could you provide the same solution there? Or if you think it will take too long, maybe add a ticket to also add it to those later?

It should be straight forward to:
- Create a new Sentry project
- Add a DSN key to the config
- Initialise the Sentry Hub
- Catch all throwables

So I will add it step by step to the other services in the context of https://jira.uitdatabank.be/browse/III-3386

jdrieghe (approved):

PR #502, opened 2020-09-28
Update udb php

jdrieghe:
### Changed
- Updated dependency on `cultuurnet/udb3` to latest version
- Removed all usages of unnecessary `CalendarInterface`

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1404,12 +1404,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/cultuurnet/udb3-php.git",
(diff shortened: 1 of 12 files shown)


jdrieghe:
@bertramakers if it's equally easy to remove `CalendarInterface` usages in the other apps, we could probably get rid of it entirely

bertramakers:
> @bertramakers if it's equally easy to remove `CalendarInterface` usages in the other apps, we could probably get rid of it entirely

Should be yes, it's basically just a value object. Not sure why it has an interface. There's only one implementation anyway and it's not like it had to be de-coupled.

bertramakers (approved):

PR #507, opened 2020-10-07
III-3528 Capture user info and runtime info

LucWollants:
### Added

- Capture user information (ID, sub, uid, uitidv1id)
- Capture runtime information (api_key, api_name)

---
Ticket: https://jira.uitdatabank.be/browse/III-3528

### DIFF
--- a/app/SentryServiceProvider.php
+++ b/app/SentryServiceProvider.php
@@ -29,8 +29,9 @@ class SentryServiceProvider implements ServiceProviderInterface
             function ($app) {
                 return new SentryErrorHandler(
                     $app[HubInterface::class],
-                    $app['jwt'],
-                    $app['auth.api_key']
+                    $app['jwt'] ?? null,
+                    $app['auth.api_key'] ?? null,
+                    $app['api_name'] ?? null
                 );
             }
         );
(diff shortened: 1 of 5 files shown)


jdrieghe (approved):
Looks good to me

app/SentryServiceProvider.php
```diff
@@ -26,20 +25,19 @@ function (Application $app) {
...
-
-                return static function (Throwable $throwable) use ($sentryHub) {
-                    $sentryHub->captureException($throwable);
-                    throw $throwable;
-                };
+                return new SentryErrorHandler(
+                    $app[HubInterface::class],
+                    $app['jwt'] ?? null,
+                    $app['auth.api_key'] ?? null,
+                    $app['api_name'] ?? null
```
bertramakers, line 34: This parameter is not nullable according to https://github.com/cultuurnet/udb3-silex/pull/507/files#diff-64fd9397ae1ed3a3e08bd0a9da5c7f02R27 so the default should be an empty string instead. Or make it nullable.

bertramakers: But good idea to also include it! 👍

LucWollants: I made it consistent with other parameters and it's also nullable now, see https://github.com/cultuurnet/udb3-silex/pull/508

PR #510, opened 2020-10-12
Pass EventType to AvailableTo factory method where needed

jdrieghe:
### Changed
- We now correctly use the EventType to determine what calendar date to use for AvailableTo

---
Ticket: https://jira.uitdatabank.be/browse/III-3242

### DIFF
--- a/app/Event/EventJSONLDServiceProvider.php
+++ b/app/Event/EventJSONLDServiceProvider.php
@@ -6,6 +6,7 @@ use CommerceGuys\Intl\Currency\CurrencyRepository;
 use CommerceGuys\Intl\NumberFormat\NumberFormatRepository;
 use CultuurNet\UDB3\Cdb\PriceDescriptionParser;
 use CultuurNet\UDB3\Doctrine\ReadModel\CacheDocumentRepository;
+use CultuurNet\UDB3\Event\EventTypeResolver;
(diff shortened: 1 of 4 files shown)


bertramakers:
Code changes look good 👍 
Maybe you can also add a test for this, or change the existing `it_handles_copy_event` test in the file you changed?

Also to remind ourselves how far we've come: In the past this would have required changes to two repositories. I'm still very happy that we can make a lot of the changes in a single repository now 🤩

jdrieghe:
@bertramakers I added the missing test for the AvailableTo factory method.
I think the other part is already covered or the AvailableTo would not be properly set on the copied event.

bertramakers:
How do you know that the logic you added to the projector to determine the event type is correct and it's not always passing `null` for example?

jdrieghe:
@bertramakers Good point. Testing that will couple it to the specific use case but that's probably better than not having the test. Will add it.

jdrieghe:
I'm sorry it took a bit longer. Had quite a busy schedule combined with some time off this week.

bertramakers (approved):
No problem at all @jdrieghe , if you're busy I'd rather have you take your time to polish it a few days later and make it as good as possible (in this case with extra tests), then that you'd rush to get it done a bit sooner. 👍

PR #512, opened 2020-10-15
Remove saved searches sapi2

jdrieghe:
### Removed
- All usages of search API 2 when dealing with saved searches. We always use SapiVersion 3 for this feature.

---
Ticket: https://jira.uitdatabank.be/browse/III-3502


app/Migrations/Version20201015123012.php
```diff
@@ -0,0 +1,18 @@
...
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20201015123012 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $schema->dropTable('saved_searches_sapi2');
+    }
+
+    public function down(Schema $schema): void
```
LucWollants, line 15: I guess we no longer put in the effort to create a `down` migration? I see that was still done in the past but is rather useless.

LucWollants (approved):

app/Migrations/Version20201015123012.php
jdrieghe: I can see cases where it might be useful, but dropping a read model table is not one of them

PR #514, opened 2020-10-19
III-3530 Capture "all" exceptions

LucWollants:
### Changed

- Capture all exceptions to Sentry but discard HTTP not found exceptions [404]

---

Ticket: https://jira.uitdatabank.be/browse/III-3530

### DIFF
--- a/bin/udb3.php
+++ b/bin/udb3.php
@@ -96,6 +96,7 @@ $consoleApp->add(new DispatchMarkedAsDuplicateEventCommand($app['event_command_b
 try {
     $consoleApp->run();
 } catch (\Throwable $throwable) {
-    // The runtime exception is re-thrown to show it on the command line.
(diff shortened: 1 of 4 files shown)


bin/udb3.php
```diff
@@ -43,6 +44,7 @@
 
 /** @var \Knp\Console\Application $consoleApp */
 $consoleApp = $app['console'];
+$consoleApp->setCatchExceptions(false);
```
bertramakers, line 47: I think we'll also have to fix this in the cdbxml service: https://github.com/cultuurnet/udb3-cdbxml-service/blob/master/bin/app.php#L35

bertramakers: Fixed it in https://github.com/cultuurnet/udb3-cdbxml-service/pull/125

bertramakers:
@jdrieghe since both @LucWollants and I worked on this, maybe you can do a final check?

PR #516, opened 2020-10-22
III-3501 Create the offer popularity table

LucWollants:
### Added
- Added migration to create the offer popularity table
---
Ticket: https://jira.uitdatabank.be/browse/III-3501

### DIFF
--- a/app/Migrations/Version20201022154706.php
+++ b/app/Migrations/Version20201022154706.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace CultuurNet\UDB3\Silex\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20201022154706 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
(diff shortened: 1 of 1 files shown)


jdrieghe (approved):
Looks clean

app/Migrations/Version20201022154706.php
```diff
@@ -0,0 +1,57 @@
...
+            [
+                'notnull' => true,
+            ]
+        );
+
+        $table->addIndex(
+            [
+                'offer_id',
+            ],
+            'offer_id_index'
```
bertramakers, line 49: Maybe we should set this as the primary key? Then it will always get an index as well.

bertramakers (approved):

app/Migrations/Version20201022154706.php
LucWollants: Then we also have a unique constraint on that column, that is not an issue for the upload?

bertramakers: As noted in the meeting notes (but maybe a bit buried) the table will be truncated every time before new data is inserted.

PR #517, opened 2020-10-22
III-3584 Fix incorrect organizer name in events imported from XML

bertramakers:
### Fixed

- Fixed events imported from XML with an organizer without an id so they have the correct name format.

---
Ticket: https://jira.uitdatabank.be/browse/III-3584

Notes: 
- In the first commit I just fixed the same format in existing test data. However, this didn't break any tests. I think because the tests don't actually do anything with those specific properties in those files.
- I still have to re-add the fallback to deal with this incorrect format in SAPI3, because we cannot do a full replay right away to fix this for all existing events with an organizer like this. The fallbacks were removed in https://github.com/cultuurnet/udb3-search-service/pull/70 because we thought we had cleaned up all of these old formats.


LucWollants (approved):
Looks good! Maintaining these `json` test files brings back memories 😅

PR #518, opened 2020-10-23
III-3583 Add Popularity repository

LucWollants:
### Added
- Add Popularity value object
- Add PopularityRepository

---

### DIFF
--- a/src/Offer/Popularity/DBALPopularityRepository.php
+++ b/src/Offer/Popularity/DBALPopularityRepository.php
@@ -30,6 +30,6 @@ class DBALPopularityRepository implements PopularityRepository
             ->execute()
             ->fetchColumn();
 
-        return new Popularity((int) $popularity ?: 0);
+        return new Popularity((int) $popularity);
     }
 }
--- a/src/Offer/Popularity/DBALPopularityRepository.php
+++ b/src/Offer/Popularity/DBALPopularityRepository.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\Popularity;
+
+use Doctrine\DBAL\Connection;
+
+class DBALPopularityRepository implements PopularityRepository
+{
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    public function get(string $offerId): Popularity
+    {
+        $queryBuilder = $this->connection->createQueryBuilder();
+
+        $popularity = $queryBuilder
+            ->select('popularity')
+            ->from('offer_popularity')
+            ->where('offer_id = :offer_id')
+            ->setParameter(':offer_id', $offerId)
+            ->execute()
+            ->fetchColumn();
+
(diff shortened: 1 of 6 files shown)


src/Offer/Popularity/Popularity.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\Popularity;
+
+class Popularity
```
bertramakers, line 7: Good idea to not extend the base classes we used to use. Would be good if we can remove those at some point in the long run 👍

src/Offer/Popularity/DBALPopularityRepository.php
```diff
@@ -0,0 +1,35 @@
...
+
+        $popularity = $queryBuilder
+            ->select('popularity')
+            ->from('offer_popularity')
+            ->where('offer_id = :offer_id')
+            ->setParameter(':offer_id', $offerId)
+            ->execute()
+            ->fetchColumn();
+
+        return new Popularity((int) $popularity ?: 0);
```
bertramakers, line 33: Nit: I think `(int) $popularity` might suffice because `null` or `false` gets casted to `0` anyway by the `(int)`

bertramakers:
I think we're still missing a migration to actually add the table outside of the tests?
We only use the `SchemaConfigurator` in the SQLite tests (until we can have MySQL tests and just use the migrations)

LucWollants:
I added an extra validation on `Popularity` to be sure it's not smaller than zero.

tests/Offer/Popularity/PopularityTest.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * @test
+     * @dataProvider inValidTypeProvider
+     * @param mixed $value
+     */
+    public function it_throws_on_invalid_types($value): void
+    {
+        $this->expectException(TypeError::class);
```
bertramakers, line 41: Nit: This is of course only the case in this test file because it has `declare(strict_types=1);`
If you'd call `new Popularity()` with a float, string, etc from another place where there are no `strict_types` used it won't throw. Instead, it will automatically cast. But there's not much we can do about that. Maybe a small change we can do is change the test name to `it_throws_on_invalid_types_if_strict_types_is_enabled`

LucWollants: Another option could be to add an extra validation:
```
if (!is_int($value)) {
    throw new TypeError('Popularity should be an integer.');
}
```

bertramakers: That wouldn't do anything unless you remove the type hint in the constructor. Because the type hint in the constructor makes it cast automatically to `int` if it's called without `strict_types`.

bertramakers: But I think we should leave it like this (or at most change the test name) and move towards using `strict_types` consistently in the long run.

LucWollants: Ah yes, the typehint and implicit typecasting. I'm getting so used to `strict_types` 😀

bertramakers (approved):
Thanks @LucWollants , it's looking great 👍

PR #519, opened 2020-10-23
III-3585 Log amqp errors to sentry

bertramakers:
### Added

- Added Sentry decorator to PSR loggers, similar to https://github.com/cultuurnet/udb3-search-service/pull/79 but a bit simpler because the bootstrapping required less changes

---
Ticket: https://jira.uitdatabank.be/browse/III-3585

### DIFF
--- a/app/SentryPsrLoggerDecorator.php
+++ b/app/SentryPsrLoggerDecorator.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Silex;
+
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+/**
+ * Decorates a PSR logger to also send all errors and emergencies that contain a Throwable in their context (in an
+ * "exception" key) to Sentry.
+ * Originally written to send error/emergency logs written by CultuurNet\BroadwayAMQP\AbstractConsumer
+ * (i.e. all AMQP consumers) to Sentry.
+ */
+final class SentryPsrLoggerDecorator implements LoggerInterface
+{
+    /**
+     * @var SentryErrorHandler
+     */
+    private $sentryHandler;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    public function __construct(SentryErrorHandler $sentryHandler, LoggerInterface $logger)
+    {
+        $this->sentryHandler = $sentryHandler;
+        $this->logger = $logger;
+    }
+
+    public function error($message, array $context = array()): void
+    {
+        $this->sendThrowableToSentry($context);
+        $this->logger->error($message, $context);
+    }
+
+    public function emergency($message, array $context = array()): void
+    {
+        $this->sendThrowableToSentry($context);
+        $this->logger->emergency($message, $context);
+    }
+
+    public function alert($message, array $context = array()): void
+    {
+        $this->logger->alert($message, $context);
+    }
+
+    public function critical($message, array $context = array()): void
+    {
+        $this->logger->critical($message, $context);
+    }
+
+    public function warning($message, array $context = array()): void
+    {
+        $this->logger->warning($message, $context);
+    }
+
+    public function notice($message, array $context = array()): void
+    {
+        $this->logger->notice($message, $context);
+    }
+
+    public function info($message, array $context = array()): void
+    {
+        $this->logger->info($message, $context);
+    }
+
+    public function debug($message, array $context = array()): void
+    {
+        $this->logger->debug($message, $context);
+    }
+
+    public function log($level, $message, array $context = array()): void
+    {
+        $this->logger->log($message, $message, $context);
+    }
+
+    private function sendThrowableToSentry(array $context = array()): void
+    {
+        $throwable = $context['exception'] ?? null;
+        if (!$throwable || !($throwable instanceof Throwable)) {
+            return;
+        }
+
+        $this->sentryHandler->handle($throwable);
+    }
+}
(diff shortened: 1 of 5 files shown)


app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private function sendThrowableToSentry(array $context = array()): void
+    {
+        $throwable = $context['exception'] ?? null;
+        if (!$throwable || !($throwable instanceof Throwable)) {
+            return;
+        }
+
+        $this->sentryHandler->handle($throwable);
```
jdrieghe, line 88:
```suggestion
        $throwable = $context['exception'] ?? null;
        if ($throwable instanceof Throwable) {
            $this->sentryHandler->handle($throwable);
        }
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    private $logger;
+
+    public function __construct(SentryErrorHandler $sentryHandler, LoggerInterface $logger)
+    {
+        $this->sentryHandler = $sentryHandler;
+        $this->logger = $logger;
+    }
+
+    public function error($message, array $context = array()): void
```
jdrieghe, line 34:
```suggestion
    public function error($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->sendThrowableToSentry($context);
+        $this->logger->error($message, $context);
+    }
+
+    public function emergency($message, array $context = array()): void
```
jdrieghe, line 40:
```suggestion
    public function emergency($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->sendThrowableToSentry($context);
+        $this->logger->emergency($message, $context);
+    }
+
+    public function alert($message, array $context = array()): void
```
jdrieghe, line 46:
```suggestion
    public function alert($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->alert($message, $context);
+    }
+
+    public function critical($message, array $context = array()): void
```
jdrieghe, line 51:
```suggestion
    public function critical($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->critical($message, $context);
+    }
+
+    public function warning($message, array $context = array()): void
```
jdrieghe, line 56:
```suggestion
    public function warning($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->warning($message, $context);
+    }
+
+    public function notice($message, array $context = array()): void
```
jdrieghe, line 61:
```suggestion
    public function notice($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->notice($message, $context);
+    }
+
+    public function info($message, array $context = array()): void
```
jdrieghe, line 66:
```suggestion
    public function info($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->info($message, $context);
+    }
+
+    public function debug($message, array $context = array()): void
```
jdrieghe, line 71:
```suggestion
    public function debug($message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->debug($message, $context);
+    }
+
+    public function log($level, $message, array $context = array()): void
```
jdrieghe, line 76:
```suggestion
    public function log($level, $message, array $context = []): void
```

app/SentryPsrLoggerDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger->log($level, $message, $context);
+    }
+
+    private function sendThrowableToSentry(array $context = array()): void
```
jdrieghe, line 81:
```suggestion
    private function sendThrowableToSentry(array $context = []): void
```

bootstrap.php
```diff
@@ -1019,8 +1021,8 @@ function ($app) {
     }
 );
 
-$app['logger.amqp.event_bus_forwarder'] = $app->share(
-    function () {
+$app['logger.amqp.event_bus_forwarder'] = $app::share(
```
jdrieghe, line 1024: Why do we call `::share()` statically on an instance?

jdrieghe (changes requested):

bootstrap.php
bertramakers: Because PHPStorm added a warning that it should be called statically and it was obfuscating other warnings inside the closure.

Screenshot 2020-10-26 at 11 22 09

It says that on every `$app->share` btw:
Screenshot 2020-10-26 at 11 22 55

Might be the plugin @LucWollants recommended.

But anyway, it's a static function and it's valid syntax to call it like that, even on an instance.

LucWollants: I also noticed this "issue" with the various assert methods from PHPUnit.

jdrieghe: I disabled that rule because phpunit expects to call non-statically even though it's still a static method.

jdrieghe (approved):

LucWollants (approved):

PR #521, opened 2020-10-27
III-3583 add popularity to jsonld

bertramakers:
### Added

- Added popularity to Event & Place JSON-LD

---

### DIFF
--- a/src/Offer/Popularity/InMemoryPopularityRepository.php
+++ b/src/Offer/Popularity/InMemoryPopularityRepository.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\Popularity;
+
+final class InMemoryPopularityRepository implements PopularityRepository
+{
+    /**
+     * @var array
+     */
+    private $popularityScores = [];
+
+    public function saveScore(string $offerId, Popularity $popularity)
+    {
+        $this->popularityScores[$offerId] = $popularity;
+    }
+
+    public function get(string $offerId): Popularity
+    {
+        return $this->popularityScores[$offerId] ?? new Popularity(0);
+    }
+}
--- a/src/ReadModel/JsonDocument.php
+++ b/src/ReadModel/JsonDocument.php
@@ -8,7 +8,7 @@ namespace CultuurNet\UDB3\ReadModel;
 use Broadway\ReadModel\ReadModelInterface;
 use stdClass;
 
-class JsonDocument implements ReadModelInterface
+final class JsonDocument implements ReadModelInterface
 {
     protected $id;
     protected $body;
@@ -32,6 +32,11 @@ class JsonDocument implements ReadModelInterface
         return (object) json_decode($this->body);
     }
 
+    public function getAssocBody(): array
+    {
+        return json_decode($this->body, true);
+    }
+
     public function getRawBody()
     {
         return $this->body;
@@ -46,6 +51,11 @@ class JsonDocument implements ReadModelInterface
         return new self($this->id, json_encode($body));
     }
 
+    public function withAssocBody(array $body): JsonDocument
+    {
+        return new self($this->id, json_encode($body));
+    }
+
     /**
      * @param callable $fn
      * @return static
@@ -54,5 +64,11 @@ class JsonDocument implements ReadModelInterface
     {
         $body = $fn($this->getBody());
         return $this->withBody($body);
+    }
+
+    public function applyAssoc(callable $fn): JsonDocument
+    {
+        $body = $fn($this->getAssocBody());
+        return $this->withAssocBody($body);
     }
 }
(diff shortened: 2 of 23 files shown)


bertramakers:
@LucWollants this should be most of the work to add popularity scores to the JSON-LD of the events and places. However, we forgot to refactor the `ReadPlaceRestController` to also use the document repository directly instead of the `EntityServiceInterface`. Maybe we should refactor that first in a separate PR like we did for the event controller?

LucWollants:
Indeed, if we first refactor `ReadPlaceRestController` we have a nice step-by-step flow.

src/Offer/Popularity/InMemoryPopularityRepository.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\Popularity;
+
+final class InMemoryPopularityRepository implements PopularityRepository
+{
```
LucWollants, line 8: Just out of curiosity, what's the reason of creating a concrete class for testing purposes instead of mocking?

jdrieghe: I like the clarity compared to using mocks though. If it's only used in tests, I would suggest moving it closer to the tests, potentially even as an [anonymous class](https://www.php.net/manual/en/language.oop5.anonymous.php) that implements the interface.

LucWollants: Not sure if this clarity out-weights the extra code to maintain. I'm personally still in favour of mocking, but that can change of course 😉

Maybe it's also dependent on the size if the interface. This is a small interface so the concrete class is also very small. But with bigger interfaces that can result in more code. Although a big interface is maybe a code smell.

bertramakers: Maybe the `InMemory` implementation of the `DocumentRepository` (used in the same class) is a better example. It's a lot less code to maintain compared to having mocks in all the tests it's being used. But that's also a widely-used interface. (We already had that `InMemory` implementation long before btw.)

I personally also think it's clearer to have very simple implementations compared to mocks, and in the case of repositories it's usually simple to implement with an array.

Maybe also less chance of mocking something incorrectly? But that can be up for debate since you could also implement the InMemory implementation incorrectly.

But in this case, it also just felt like less work to implement it once compared to writing mock expectations in every test.

For more complex interfaces mocks are definitely the way to go of course.

And in the end, it's probably subjective as well. (And trivial)

I like the suggestion of moving it closer to the tests and/or using anonymous classes @jdrieghe . I was actually in doubt whether to place the call inside the `tests` or `src` folder, but ended up going with `src` because we also have other `InMemory` implementations for tests there. But that was maybe to work around a limitation when they were still in `udb3-php` and if they would be in the `tests` folder there, we couldn't load them inside the tests in `udb3-silex` because composer wouldn't autoload the classes in the `tests` folders of dependencies. So yeah we could move them to `tests` now. But maybe that's pretty minor and not necessarily worth the effort right now?

bertramakers: > Not sure if this clarity out-weights the extra code to maintain.

It's 23 lines of code and 1 line of code per test (calling the `saveScore` method), compared to ~4 lines of code per test to mock the methods:

```
$this->popularityRepository->expects($this->once())
    ->method('get')
    ->with($offerId, $includeMetadata)
    ->willReturn(...)
```

So in this case of only 3 tests, it's a little bit more code yes (`23+3*1 = 26` vs `3*4 = 12`, so 14 lines of code extra), but negligible I'd say. And in cases where there'd be more tests (like all the tests where we use `InMemoryDocumentRepository`) it would actually save a lot of code.

Plus if the interface ever changes it's easier to refactor the implementation than to refactor multiple mocks imho.

LucWollants: I'm convinced now! 😀

> It's a lot less code to maintain compared to having mocks in all the tests where it's being used
I was too focused on this pull request. But if we make more classes dependent on `PopularityRepository` that would indeed require lots of duplicated mock code which can be avoided with this concrete implementation.

tests/Offer/Popularity/PopularityEnrichedOfferRepositoryTest.php
```diff
@@ -48,15 +48,17 @@ public function it_does_not_add_popularity_score_if_include_metadata_is_false():
         $jsonLd = new JsonDocument($offerId, json_encode(['@type' => 'Event']));
         $this->decoratedRepository->save($jsonLd);
 
-        $returnedJsonLd = $this->popularityEnrichedOfferRepository->get($offerId, false);
+        $fetchJsonLd = $this->popularityEnrichedOfferRepository->get($offerId, false);
```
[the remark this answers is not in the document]
bertramakers: Yeah good spot, I'll fix it 👍

src/ReadModel/JsonDocument.php
```diff
@@ -46,6 +51,11 @@ public function withBody($body)
         return new self($this->id, json_encode($body));
     }
 
+    public function withAssocBody(array $body): JsonDocument
```
LucWollants, line 54: A very small nuance in naming, maybe use `fromAssocBody` to make it clear it's a named static constructor and not a mutation?

bertramakers: It's not a static constructor though, it uses `$this->id`. It's basically the same as this but as a one-liner:

```
$c = clone $this;
$c->body = json_encode($body);
return $c;
```

bertramakers: (And it's based on the existing `withBody()`)

LucWollants (approved):

PR #522, opened 2020-10-28
III-3583 use document repository in place read controller

bertramakers:
### Changed

- Refactored place read controller to use document repository directly instead of a service in-between
- Deprecated `DocumentRepository::get()` in favor of a new `DocumentRepository::fetch()` which has more consistent error-handling

### Added

- Added `DocumentRepository::fetch()` which always throws if a document is not found, instead of either returning `null` if it's not found or throwing if it was deleted. The difference can still be deduced based on the `isNotFound()` and `isGone()` methods on the new exception type. 

---

---

Note: The new `fetch()` method should also be useful for https://jira.uitdatabank.be/browse/III-3593 in which we want to remove a service that converts the `null` return value to an exception

### DIFF
--- a/src/Http/Place/ReadPlaceRestController.php
+++ b/src/Http/Place/ReadPlaceRestController.php
@@ -8,8 +8,10 @@ use CultuurNet\SearchV3\Serializer\SerializerInterface;
 use CultuurNet\SearchV3\ValueObjects\Place;
 use CultuurNet\UDB3\EntityNotFoundException;
 use CultuurNet\UDB3\EntityServiceInterface;
+use CultuurNet\UDB3\Event\ReadModel\DocumentGoneException;
 use CultuurNet\UDB3\Http\ApiProblemJsonResponseTrait;
 use CultuurNet\UDB3\Http\JsonLdResponse;
+use CultuurNet\UDB3\ReadModel\JsonDocument;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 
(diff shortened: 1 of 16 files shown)


bertramakers:
@LucWollants @jdrieghe this will need to be merged first before we can merge #521 
I'm already going to merge this into that PR/branch so I can test it manually that the popularity also gets added for places. I'll update #521 to point to this PR.

src/Http/Place/ReadPlaceRestController.php
```diff
@@ -78,17 +71,16 @@ public function getCalendarSummary($cdbid, Request $request)
         $place = $this->serializer->deserialize($data->getRawBody(), Place::class);
 
         if ($style !== 'html' && $style !== 'text') {
```
jdrieghe, line 71: pretty sure there's a value object hiding here somewhere

src/ReadModel/DocumentDoesNotExistException.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\ReadModel;
+
+use Exception;
+
+final class DocumentDoesNotExistException extends Exception
```
jdrieghe, line 9: Can we get rid of the `Exceptions` postfix here?

jdrieghe (approved):
Nice refactor. Would avoid adding the `Exception` postfix to new exception classes though.

src/Http/Place/ReadPlaceRestController.php
bertramakers: True but I didn't want to focus on the calendar summary methods so the scope wouldn't get too wide. It's already a lot of changes at once 😄

src/ReadModel/DocumentDoesNotExistException.php
bertramakers: Good point, I'll remove that!

jdrieghe (approved):

LucWollants (approved):
Nice refactor 👌

PR #523, opened 2020-10-29
Avoid purging of the index_readmodel table

LucWollants:
### Removed
- The table `index_readmodel` no longer exists so no purge is needed
### Added
- Added missing read models to purge

---

Ticket: https://jira.uitdatabank.be/browse/III-3500


bertramakers:
I think we're still missing `labels_json`, aside from that everything looks good!

LucWollants:
Ok, I think `labels_json` is missing then also in Confluence.

bertramakers (approved):

PR #525, opened 2020-11-24
III-3570 Add endpoint to update subEvents status(es) for single and multiple calendar types

jdrieghe:
### Added

- Added new endpoint to update the status of subEvents for events with single or multiple calendar type
- Added support for new status projection on events and places in tests

---
Ticket: https://jira.uitdatabank.be/browse/III-3570


jdrieghe:
@bertramakers we still need to fix the tests that were broken in this library by updating `udb3-php` with the latest changes

tests/Event/ReadModel/JSONLD/EventLDProjectorTest.php
```diff
@@ -545,11 +545,15 @@ public function it_handles_new_events_with_multiple_timestamps()
...
                 'endDate' => '2015-01-27T13:25:21+01:00',
+                'eventStatus' => 'EventScheduled',
+                'eventStatusReason' => [],
             ],
             (object)[
                 '@type' => 'Event',
                 'startDate' => '2015-01-28T13:25:21+01:00',
                 'endDate' => '2015-01-29T13:25:21+01:00',
+                'eventStatus' => 'EventScheduled',
+                'eventStatusReason' => [],
```
bertramakers, line 556: Hmm I think we might need to omit this if it's empty, instead of setting it to an empty array/object 🤔

bertramakers: I've omitted it as part of https://github.com/cultuurnet/udb3-php/pull/482

bertramakers (approved):
@jdrieghe Since you're the author of this PR and cannot approve I've gone ahead and marked this as approved. But can you go over my extra commits before we merge this? Especially since it's now more commits by me than you 😅 Or alternatively @LucWollants could go over everything.

It will also require a merge of https://github.com/cultuurnet/udb3-php/pull/481 first which will require an approval from you (or @LucWollants of course). And then we'll have to update the composer dependency in this PR again.

src/Http/Event/UpdateSubEventsStatusRequestHandler.php
```diff
@@ -32,7 +32,7 @@ public function __construct(
         $this->validator = $validator;
     }
 
-    public function __invoke(Request $request, string $eventId)
+    public function handle(Request $request, string $eventId)
```
jdrieghe, line 36: It was worth a try :D

jdrieghe:
Thanks for the follow-up work, @bertramakers.
This looks good and can be merged, I think.

src/Http/Event/UpdateSubEventsStatusRequestHandler.php
bertramakers: Yeah I'd also have preferred this since it was a lot cleaner in the controller provider in Silex, but I got an error along the lines of `0 arguments passed to UpdateSubEventsStatusRequestHandler, 2 expected`. So I think Silex was trying to instantiate the class itself and failing to do so correctly.

PR #527, opened 2020-12-03
Use correct table when geocoding events and places

jdrieghe:
### Changed
- We now use the correct event_store table when geocoding places and events.


bertramakers (approved):
This is a good fix to get the command running 👍 
We'll still need to look into the weird result counts though. (See comment in https://jira.uitdatabank.be/browse/III-3634)

PR #529, opened 2020-12-07
III-3634 Add debug logging for geocoding results

bertramakers:
### Added

- Added debug logging for geocoding results

---
Ticket: https://jira.uitdatabank.be/browse/III-3634

### DIFF
--- a/app/Event/EventGeoCoordinatesServiceProvider.php
+++ b/app/Event/EventGeoCoordinatesServiceProvider.php
@@ -21,12 +21,16 @@ class EventGeoCoordinatesServiceProvider implements ServiceProviderInterface
     {
         $app['event_geocoordinates_command_handler'] = $app->share(
             function (Application $app) {
-                return new GeoCoordinatesCommandHandler(
+                $handler = new GeoCoordinatesCommandHandler(
                     $app['event_repository'],
                     new DefaultAddressFormatter(),
                     new LocalityAddressFormatter(),
                     $app['geocoding_service']
                 );
+
+                $handler->setLogger($app['logger.command_bus']);
+
+                return $handler;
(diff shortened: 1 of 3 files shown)


app/Event/EventGeoCoordinatesServiceProvider.php
```diff
@@ -21,12 +21,16 @@ public function register(Application $app)
...
             function (Application $app) {
-                return new GeoCoordinatesCommandHandler(
+                $handler = new GeoCoordinatesCommandHandler(
                     $app['event_repository'],
                     new DefaultAddressFormatter(),
                     new LocalityAddressFormatter(),
                     $app['geocoding_service']
                 );
+
+                $handler->setLogger($app['logger.command_bus']);
```
bertramakers, line 31: We included it in this log because the `DefaultGeoCodingService` also uses this logger.

src/Offer/AbstractGeoCoordinatesCommandHandler.php
```diff
@@ -46,33 +51,52 @@ public function __construct(
...
                 )
             );
+
+            $coordinates = $this->geocodingService->getCoordinates($fallbackAddress);
+
+            if (!is_null($coordinates)) {
+                $this->logger->debug(
+                    "Found coordinates for fallback address '$fallbackAddress' for offer id $offerId"
+                );
+            }
```
bertramakers, line 90: We could have made this an `else` for the `if ($coordinates === null)` below, but we didn't because we only want to log this when we find coordinates for the fallback address. Not when we find coordinates for the full address, because we're not interested in those cases.

jdrieghe (approved):
Looks clean

PR #530, opened 2020-12-07
III-3634 Refactor CLI geocode commands to get event/place ids from SAPI3

bertramakers:
### Changed

- Changed the `event:geocode` and `place:geocode` commands to fetch the events/places that require geocoding from SAPI3
- Renamed `--cdbid` option to `--id`

### Added

- Added `--force` option to skip confirmation, so we can run this automatically every once in a while if necessary

### Removed

- Removed the `--all` option from the `event:geocode` and `place:geocode` commands

---
Ticket: https://jira.uitdatabank.be/browse/III-3634


app/Console/AbstractGeocodeCommand.php
```diff
@@ -15,10 +16,24 @@ abstract class AbstractGeocodeCommand extends AbstractCommand
...
+        CommandBusInterface $commandBus,
+        ResultsGeneratorInterface $searchResultsGenerator,
+        DocumentRepository $documentRepository
+    ) {
         parent::__construct($commandBus);
         $this->searchResultsGenerator = $searchResultsGenerator;
+        $this->documentRepository = $documentRepository;
+    }
+
+    protected function getDocumentRepository(): DocumentRepository
```
jdrieghe: Why not expose the property itself as protected?

jdrieghe (approved):
💅

app/Console/AbstractGeocodeCommand.php
bertramakers: I think either way is fine personally, no strong preference. I guess it's just a habit of making as much as possible private and only exposing what's needed (in this case returning the repository, but not being able to overwrite it). Anyway, it doesn't really matter since we refactored it a commit or two later to not expose the repository at all but to add a method to directly get the needed documents.

PR #531, opened 2020-12-09
III-3643 Add CLI to reindex offers with popularity

LucWollants:
### Added
- Add a new CLI command to reindex offers with a popularity score in SAPI3.

---

https://jira.uitdatabank.be/browse/III-3634

### DIFF
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -2,6 +2,7 @@
 
 namespace CultuurNet\UDB3\Silex\Console;
 
+use Doctrine\DBAL\Connection;
 use Knp\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -13,6 +14,17 @@ class ReindexOffersWithPopularityScore extends Command
         'event',
         'place',
     ];
+
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        parent::__construct();
+        $this->connection = $connection;
+    }
 
     protected function configure(): void
     {
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -59,11 +59,6 @@ class ReindexOffersWithPopularityScore extends Command
         $this
             ->setName($this->type.':reindex-offers-with-popularity')
             ->setDescription('Reindex events or places that have a popularity score.')
-            ->addArgument(
-                'type',
-                InputArgument::REQUIRED,
-                'The type of the offer, either place or event.'
-            )
             ->addOption(
                 'force',
                 null,
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -82,8 +82,8 @@ class ReindexOffersWithPopularityScore extends Command
             throw new \InvalidArgumentException('The type "' . $type . '" is not support. Use event or place.');
         }
 
-        $offers = $this->getOffers($type);
-        if (count($offers) < 1) {
+        $offerIds = $this->getOfferIds($type);
+        if (count($offerIds) < 1) {
             $output->writeln('No ' . $type . 's found with a popularity.');
             return 0;
         }
@@ -95,7 +95,7 @@ class ReindexOffersWithPopularityScore extends Command
         return 0;
     }
 
-    private function getOffers(string $type): array
+    private function getOfferIds(string $type): array
     {
         return $this->connection->createQueryBuilder()
             ->select('offer_id')
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -76,7 +76,7 @@ class ReindexOffersWithPopularityScore extends Command
     {
         $type = $input->getArgument('type');
         if (!\in_array($type, $this->allowedTypes, true)) {
-            throw new \InvalidArgumentException('The type "' . $type . '" is not support. Use event or place.');
+            throw new \InvalidArgumentException('The type "' . $type . '" is not supported. Use event or place.');
         }
 
         $offerIds = $this->getOfferIds($type);
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -3,7 +3,6 @@
 namespace CultuurNet\UDB3\Silex\Console;
 
 use Broadway\Domain\DomainEventStream;
-use Broadway\EventHandling\EventBusInterface;
 use CultuurNet\Broadway\EventHandling\ReplayModeEventBusInterface;
 use CultuurNet\UDB3\ReadModel\DocumentEventFactory;
 use Doctrine\DBAL\Connection;
@@ -28,7 +27,7 @@ class ReindexOffersWithPopularityScore extends Command
     private $connection;
 
     /**
-     * @var EventBusInterface
+     * @var ReplayModeEventBusInterface
      */
     private $eventBus;
 
@@ -44,7 +43,7 @@ class ReindexOffersWithPopularityScore extends Command
 
     public function __construct(
         Connection $connection,
-        EventBusInterface $eventBus,
+        ReplayModeEventBusInterface $eventBus,
         DocumentEventFactory $eventFactoryForEvents,
         DocumentEventFactory  $eventFactoryForPlaces
     ) {
@@ -75,10 +74,6 @@ class ReindexOffersWithPopularityScore extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        if (!$this->eventBus instanceof ReplayModeEventBusInterface) {
-            throw new \InvalidArgumentException('The event bus should implement the "ReplayModeEventBusInterface".');
-        }
-
         $type = $input->getArgument('type');
         if (!\in_array($type, $this->allowedTypes, true)) {
             throw new \InvalidArgumentException('The type "' . $type . '" is not support. Use event or place.');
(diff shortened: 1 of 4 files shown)


app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -0,0 +1,146 @@
...
+                InputOption::VALUE_NONE,
+                'Skip confirmation.'
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        if (!$this->eventBus instanceof ReplayModeEventBusInterface) {
+            throw new \InvalidArgumentException('The event bus should implement the "ReplayModeEventBusInterface".');
+        }
```
bertramakers, line 79: Could we depend on `ReplayModeEventBusInterface` in the constructor if that's required anyway?

app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -0,0 +1,146 @@
...
+
+        $type = $input->getArgument('type');
+        if (!\in_array($type, $this->allowedTypes, true)) {
+            throw new \InvalidArgumentException('The type "' . $type . '" is not support. Use event or place.');
```
bertramakers, line 83:
```suggestion
            throw new \InvalidArgumentException('The type "' . $type . '" is not supported. Use event or place.');
```

app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -0,0 +1,146 @@
...
+        $offerIds = $this->getOfferIds($type);
+        if (count($offerIds) < 1) {
+            $output->writeln('No ' . $type . 's found with a popularity.');
+            return 0;
+        }
+
+        if (!$this->askConfirmation($input, $output, $type, count($offerIds))) {
+            return 0;
+        }
+        if ($this->eventBus instanceof ReplayModeEventBusInterface) {
```
bertramakers, line 95: This `if` check can be removed since we throw an exception higher up if the event bus is not an instance of `ReplayModeEventBusInterface` (regardless of my suggestion to move it as a type check to the constructor)

app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -0,0 +1,146 @@
...
+            $this->eventBus->startReplayMode();
+        }
+
+        foreach ($offerIds as $offerId) {
+            $this->dispatchEvent($type, $offerId);
+        }
+
+        if ($this->eventBus instanceof ReplayModeEventBusInterface) {
```
bertramakers, line 103: Same as above.

app/Console/ReplayCommand.php
```diff
@@ -76,7 +76,7 @@ protected function configure()
         );
 
         $this
-            ->setName('replay')
+            ->setName('offer:replay')
```
bertramakers, line 79: Isn't it also possible to replay organizers, roles, labels etc through this? Looking at the `AggregateType` options used just above this, there are more options than just offers (events/places):
```
class AggregateType extends Enum
{
    const EVENT = 'event';
    const PLACE = 'place';
    const VARIATION = 'variation';
    const ORGANIZER = 'organizer';
    const MEDIA_OBJECT = 'media_object';
    const ROLE = 'role';
    const LABEL = 'label';
}
```

bin/udb3.php
```diff
@@ -95,6 +97,14 @@
...
 $consoleApp->add(new MarkPlaceAsDuplicateCommand($app['event_command_bus'], $app[LocationMarkedAsDuplicateProcessManager::class]));
 $consoleApp->add(new DispatchMarkedAsDuplicateEventCommand($app['event_command_bus'], $app[LocationMarkedAsDuplicateProcessManager::class], $app['event_bus']));
+$consoleApp->add(
+    new ReindexOffersWithPopularityScore(
+        $app['dbal_connection'],
+        $app['event_bus'],
+        $app[EventJSONLDServiceProvider::JSONLD_PROJECTED_EVENT_FACTORY],
+        $app[PlaceJSONLDServiceProvider::JSONLD_PROJECTED_EVENT_FACTORY]
+    )
+);
```
bertramakers, line 107: Maybe we could instantiate the same `ReindexOffersWithPopularityScore` command class twice, once for `events` and once for `places`, and instead of passing in two factories we only pass the relevant one + a string for the offer type?

So we'd end up with two commands like `place:reindex-places-with-popularity` and `event:reindex-events-with-popularity`, instead of one `offer:reindex-offers-with-popularity` with two factories and a required type factory, forcing us to do some if/else checks in the command class.

Or alternatively, we could keep it like it is but make the `type` option optional and determine which factory to use for which result by looking at the type in the database row. Then we still have the two factories and if/else checks, but it would be more flexible.

In any case this is just a suggestion for a slight improvement that popped into my head while reviewing, not a blocker. It also works as it is right now and can always be refactored later if we do need to split it up or make it more flexible for some reason.

bertramakers (changes requested):

app/Console/ReindexOffersWithPopularityScore.php
LucWollants: Good point! I injected `ReplayModeEventBusInterface` instead `EventBusInterface` and this also solve various redundant if statements.

app/Console/ReindexOffersWithPopularityScore.php
LucWollants: No longer needed after injecting `ReplayModeEventBusInterface`.

app/Console/ReindexOffersWithPopularityScore.php
LucWollants: No longer needed after injecting `ReplayModeEventBusInterface`.

app/Console/ReindexOffersWithPopularityScore.php
LucWollants: Fixed typo.

app/Console/ReplayCommand.php
LucWollants: I was a bit too drastic with this so I reverted it back to the default namespace.

bin/udb3.php
LucWollants: With the first proposal we would have 2 CLI commands which makes it possible to re-index either places or events.
With the second proposal I don't think we can decide which offer type to re-index?

I think it makes sense to re-index either places or events. Because a re-index of places could fix some events.

LucWollants: I found the first proposal combining a flexible behaviour with an elegant implementation, so I did refactor it to that proposal. I also used the `OfferType` enum.

bertramakers: Either way is fine by me, I just want to clarify something about:

> Because a re-index of places could fix some events.

Possibly, but if you run `event:reindex-events-with-popularity` afterwards it will again loop over all events in the popularity table regardless if they have already been indexed as a side effect of re-indexing the places no?

bertramakers: Also:

> With the second proposal I don't think we can decide which offer type to re-index?

The second proposal was to make the `type` option _optional_, so we could choose to either run it for both events + places (don't pass the `--type`) or for one or the other (`--type=event` or `--type=place`)

bertramakers: But anyway your implementation achieves the intended goal so it's all nitpicking. 😄 They were just some suggestions that could either make the command do less if/else checks based on the type by injecting the type (as string, for configuring the name) + the right factory (suggestion 1), or that would make the command more flexible (by allowing the user to choose to run it for either all offers or a specific type). The other comments were more important and those are resolved so this looks good to me! The rest can always be refactored if ever needed.

bertramakers: Ah I see you even actually implemented suggestion no. 2! Thanks, looks great 👌

bertramakers (approved):

app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -36,28 +37,27 @@ class ReindexOffersWithPopularityScore extends Command
...
-        parent::__construct();
+        $this->type = \strtolower($type->toNative());
         $this->connection = $connection;
         $this->eventBus = $eventBus;
         $this->eventFactoryForEvents = $eventFactoryForEvents;
-        $this->eventFactoryForPlaces = $eventFactoryForPlaces;
+
+        // It's important to call the parent constructor after setting the properties.
+        // Because the parent constructor calls the `configure` method.
+        // In this command the command name is created dynamically with the type property.
```
bertramakers, line 53: Thanks for clarifying 👍

bertramakers (approved):

app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -0,0 +1,129 @@
...
+                InputOption::VALUE_NONE,
+                'Skip confirmation.'
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $offerIds = $this->getOfferIds($this->type);
+        if (count($offerIds) < 1) {
+            $output->writeln('No ' . $this->type . 's found with a popularity.');
```
jdrieghe, line 74:
```suggestion
            $output->writeln('No ' . $this->type . s found with a popularity.');
```

app/Console/ReindexOffersWithPopularityScore.php
```diff
@@ -0,0 +1,129 @@
...
+        if ($input->getOption('force')) {
+            return true;
+        }
+
+        return $this
+            ->getHelper('question')
+            ->ask(
+                $input,
+                $output,
+                new ConfirmationQuestion('Reindex ' . $count . ' ' . $type . 's? [y/N] ', false)
```
jdrieghe, line 115:
```suggestion
                new ConfirmationQuestion('Reindex ' . $count . ' ' . $type . s? [y/N] ', false)
```

bin/udb3.php
```diff
@@ -4,7 +4,7 @@
 use Broadway\Domain\Metadata;
 use CultuurNet\SilexAMQP\Console\ConsumeCommand;
 use CultuurNet\UDB3\Event\LocationMarkedAsDuplicateProcessManager;
-use CultuurNet\UDB3\Silex\ApiName;
+use CultuurNet\UDB3\Offer\OfferType;use CultuurNet\UDB3\Silex\ApiName;
```
jdrieghe, line 7:
```suggestion
use CultuurNet\UDB3\Offer\OfferType;
use CultuurNet\UDB3\Silex\ApiName;
```

jdrieghe (approved):

app/Console/ReindexOffersWithPopularityScore.php
LucWollants: I think this will cause a syntax error because the single quote is the start of the last part of the string concatenation.

app/Console/ReindexOffersWithPopularityScore.php
LucWollants: I think this will cause a syntax error because the single quote is the start of the last part of the string concatenation.

jdrieghe: Oh wow, I need more coffee 🙈

bin/udb3.php
LucWollants: Arranged the imports.

PR #533, opened 2020-12-16
Handle the StatusUpdated event.

LucWollants:
### Added

- Added a StatusUpdatedProjector to handle the StatusUpdated event on either an event or place.

---

Ticket: https://jira.uitdatabank.be/browse/III-3656

### DIFF
--- a/src/Offer/ReadModel/JSONLD/StatusUpdatedProjector.php
+++ b/src/Offer/ReadModel/JSONLD/StatusUpdatedProjector.php
@@ -6,7 +6,7 @@ use Broadway\EventHandling\EventListenerInterface;
 use CultuurNet\UDB3\EventHandling\DelegateEventHandlingToSpecificMethodTrait;
 use CultuurNet\UDB3\Offer\Events\StatusUpdated;
 use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
-use CultuurNet\UDB3\ReadModel\DocumentRepository;
+use CultuurNet\UDB3\ReadModel\OfferDocumentRepository;
 use Psr\Log\LoggerInterface;
 
 final class StatusUpdatedProjector implements EventListenerInterface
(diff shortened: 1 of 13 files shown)


src/Offer/ReadModel/JSONLD/StatusUpdatedProjector.php
```diff
@@ -0,0 +1,79 @@
...
+        $json['status'] = $statusUpdated->getStatus()->serialize();
+
+        if (!empty($json['subEvent'])) {
+            $nrOfSubEvents = count($json['subEvent']);
+            for ($subEventIndex = 0; $subEventIndex < $nrOfSubEvents; $subEventIndex++) {
+                $json['subEvent'][$subEventIndex]['status'] = $statusUpdated->getStatus()->serialize();
+            }
+        }
+
+        $this->eventRepository->save($jsonDocument->withAssocBody($json));
```
jdrieghe, line 77: Should this use the passed `$documentRepository`?

jdrieghe (changes requested):
I think we're using the wrong repository to save the document.

In general, I'm not a big fan of the semi-abstraction used here.
I would much rather inject an `OfferRepository` and modify the offer without caring about its type. If that's not possible, I would create separate projectors. Right now, it's somewhere in the middle and can become very confusing and error-prone when another implementation of that abstraction enters the equation.

src/Offer/ReadModel/JSONLD/StatusUpdatedProjector.php
LucWollants: Indeed the wrong repository!

LucWollants:
> I would much rather inject an `OfferRepository` and modify the offer without caring about its type.
That is certainly possible, but that new `OfferRepository` would have the logic to `load` and `save` from the correct repo. So for example it would first `load` from the events repo, if nothing is found it would load from the places repo.

jdrieghe:
@LucWollants That's exactly what I mean. Now, this logic is spread across potentially multiple projectors when we could easily abstract that away to just one place. Even saving can be done based on the type in the read model.

LucWollants:
The `JsonDocument` has no real type. So when doing a `save` it can become tricky to determine the correct repo because what if it's a new event or place to save?

jdrieghe (approved):

jdrieghe:
Looks great, @LucWollants. I'm happy we can now safely refactor away all the duplicate projectors as well.

PR #536, opened 2020-12-23
Fix for modified status on parent event.

LucWollants:
### Fixed
- Fix for modified status on parent event.
---
Ticket: https://jira.uitdatabank.be/browse/III-3658


jdrieghe (approved):
Maybe add a test scenario where the reason is filled in, just to be complete?

LucWollants:
> Maybe add a test scenario where the reason is filled in, just to be complete?

Good point! I added a test for this.

PR #537, opened 2020-12-24
Add endpoints to update event status

jdrieghe:
### Added
- `/events/{id}/status` endpoint to update status of an event
- `/places/{id}/status` endpoint to update status of a place

---
Ticket: https://jira.uitdatabank.be/browse/III-3659

### DIFF
--- a/app/Event/EventControllerProvider.php
+++ b/app/Event/EventControllerProvider.php
@@ -4,6 +4,7 @@ namespace CultuurNet\UDB3\Silex\Event;
 
 use CultuurNet\UDB3\Http\Event\EditEventRestController;
 use CultuurNet\UDB3\Http\Event\ReadEventRestController;
+use CultuurNet\UDB3\Http\Event\UpdateStatusRequestHandler;
 use CultuurNet\UDB3\Http\Event\UpdateSubEventsStatusRequestHandler;
 use CultuurNet\UDB3\Http\Event\UpdateSubEventsStatusValidator;
 use CultuurNet\UDB3\Http\Management\User\CultureFeedUserIdentification;
@@ -81,6 +82,7 @@ class EventControllerProvider implements ControllerProviderInterface
         $controllers->get('/{cdbid}/calsum', 'event_controller:getCalendarSummary');
 
         $controllers->patch('/{eventId}/subEvents', UpdateSubEventsStatusRequestHandler::class . ':handle');
+        $controllers->put('/{cdbid}/status', UpdateStatusRequestHandler::class . ':handle');
         return $controllers;
     }
 }
--- a/app/Event/EventControllerProvider.php
+++ b/app/Event/EventControllerProvider.php
@@ -82,7 +82,7 @@ class EventControllerProvider implements ControllerProviderInterface
         $controllers->get('/{cdbid}/calsum', 'event_controller:getCalendarSummary');
 
         $controllers->patch('/{eventId}/subEvents', UpdateSubEventsStatusRequestHandler::class . ':handle');
-        $controllers->put('/{cdbid}/status', UpdateStatusRequestHandler::class . ':handle');
+        $controllers->put('/{eventId}/status', UpdateStatusRequestHandler::class . ':handle');
         return $controllers;
     }
 }
--- a/app/Place/PlaceControllerProvider.php
+++ b/app/Place/PlaceControllerProvider.php
@@ -79,7 +79,7 @@ class PlaceControllerProvider implements ControllerProviderInterface
         $controllers->put('/{itemId}/images/{mediaObjectId}', 'place_editing_controller:updateImage');
 
         $controllers->get('/{cdbid}/calsum', 'place_controller:getCalendarSummary');
-        $controllers->put('/{cdbid}/status', UpdateStatusRequestHandler::class . ':handle');
+        $controllers->put('/{placeid}/status', UpdateStatusRequestHandler::class . ':handle');
 
         return $controllers;
     }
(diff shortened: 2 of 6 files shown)


src/Http/Event/UpdateStatusRequestHandler.php
```diff
@@ -0,0 +1,63 @@
...
+use CultuurNet\UDB3\Event\ValueObjects\Status;
+use CultuurNet\UDB3\Event\ValueObjects\StatusReason;
+use CultuurNet\UDB3\Event\ValueObjects\StatusType;
+use CultuurNet\UDB3\Http\HttpFoundation\NoContent;
+use CultuurNet\UDB3\Http\Offer\UpdateStatusValidator;
+use CultuurNet\UDB3\Language;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+class UpdateStatusRequestHandler
```
LucWollants, line 18: Not sure if it would make sense to create an abstract class that contains the main logic and has one unimplemented method `CreateUpdateStatus` with a dedicated implementation for Events and one for Places.

bertramakers: Additionally we could even change the `UpdateStatus` command so we only have one concrete command class, inside the `Offer` namespace, and then use the same trick with the combined repository as we did for the projectors? So 1 request handler, 1 command, 1 command handler, and 1 repository that tries to load the event and if it doesn't find it loads the place instead? And then it calls the `updateStatus` method on that aggregate.

jdrieghe: I'm not sure what part of the trade-off is better, but my reasoning here was "when in doubt, make it explicit".

bertramakers: What's the difference with the projectors where you suggested to have one projector using a smart repository that loads either an event or a place? I'm suggesting basically the same but for the write side. 😄

jdrieghe: I care a lot less about model explicitness in projections than in write models. For me, projections are volatile by design and can easily be changed at will and replayed. I also merge the other PR by accident cause I thought the link to the remark you had was about the superfluous test scenario.

I'm okay with refactoring this but let's do it in a separate PR so we can easily evaluate the change at the end?

bertramakers: It's just a suggestion that popped up in my head based on the solution for projectors. The current approach is also fine by me. I'm interested in exploring options to improve the general structure of our code like we did with the projectors, but it doesn't have to happen now if you believe this is the best for now for the write side.

Doing it in a separate PR is also fine, but only if you think it's worth the effort to try out and evaluate.

The things I care most about are the API endpoints/payloads/responses and event types/payloads in the event store, since those two are the hardest to refactor usually. (Because of backwards compatibility in the one case and persisted data in the other.) Aside from that everything is refactor-able so most of the times my comments with suggestions are just that, suggestions 😉 If something is a hard blocker for me I'll try to better make that clear. This is definitely not a hard blocker for me.

app/Place/PlaceControllerProvider.php
```diff
@@ -78,6 +79,7 @@ function (Application $app) {
         $controllers->put('/{itemId}/images/{mediaObjectId}', 'place_editing_controller:updateImage');
 
         $controllers->get('/{cdbid}/calsum', 'place_controller:getCalendarSummary');
+        $controllers->put('/{cdbid}/status', UpdateStatusRequestHandler::class . ':handle');
```
bertramakers, line 82: Nit: `cdbid` is an outdated term, `{placeId}` would be clearer I think.

src/Http/Event/UpdateStatusRequestHandler.php
bertramakers: To loop back to @LucWollants 's suggestion though: I'd go for either 100% explicitness like we have now, or a single request handler/command/command handler. Not an in-between solution with an abstract request handler which has an abstract `CreateUpdateStatus` method. Again just my $0.02 though.

app/Event/EventControllerProvider.php
```diff
@@ -81,6 +82,7 @@ function (Application $app) {
         $controllers->get('/{cdbid}/calsum', 'event_controller:getCalendarSummary');
 
         $controllers->patch('/{eventId}/subEvents', UpdateSubEventsStatusRequestHandler::class . ':handle');
+        $controllers->put('/{cdbid}/status', UpdateStatusRequestHandler::class . ':handle');
```
bertramakers, line 85: (Same as https://github.com/cultuurnet/udb3-silex/pull/537#discussion_r551312650)

bertramakers: I think this is also broken because you use `$eventId` in the request handler but `{cdbid}` here. In Silex the arguments need to have the same name. https://github.com/cultuurnet/udb3-silex/pull/537/files#diff-e1cc22686a32f871adef2a7a06a16fe90f4918ea13069346b390465d951e9e0aR30

bertramakers (changes requested):
I have a hunch that there's a bug with the names of the arguments in the routes vs the request handlers. For places it's fixed now with the rename to `placeId`, but for events it's still a problem if I'm not mistaken. Aside from that it looks good to me 👍

app/Event/EventControllerProvider.php
jdrieghe: Yes, good spot!

bertramakers (approved):

src/Http/Event/UpdateStatusRequestHandler.php
jdrieghe: All good, let's do it in a separate PR

PR #538, opened 2021-01-06
Refactor away duplicate command handling for UpdateStatus

jdrieghe:
### Added
- `OfferRepository` that can load and save both `Event` and `Place` aggregates

### Changed
- `UpdateStatus` is now no longer specific for place or event and is handled through a single request handler, command and command handler.


LucWollants:
Looks good to me.

Could the new class `OfferRepository` benefit from a unit test? Now the `UpdateStatusHandlerTest` only tests the `Event` flow and not the `Place` flow.

jdrieghe:
@LucWollants I wouldn't bother with a unit test for the `OfferRepostiory`. However, the way Broadway's testing tools work, they don't allow me to properly integration test it through the scenarios I think.

LucWollants:
> @LucWollants I wouldn't bother with a unit test for the `OfferRepostiory`. However, the way Broadway's testing tools work, they don't allow me to properly integration test it through the scenarios I think.

But that could be solved by injecting mocks for the place and event repo?

LucWollants (approved):
Looks good to 🚢

src/Offer/OfferRepository.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->placeRepository = $placeRepository;
+    }
+
+    public function save(AggregateRoot $aggregate): void
+    {
+        try {
+            $this->eventRepository->save($aggregate);
+        } catch (AssertionFailedException $e) {
+            $this->placeRepository->save($aggregate);
+        }
```
bertramakers, line 42: Was thinking we would check the Aggregate class name ourselves, but nice that we can simply do this since the underlying repositories already do a check on the class name themselves! 👍

bertramakers:
Late to the punch, but clear and simple solution! 👏 
This should help us a lot in the future to reduce the amount of duplicated boilerplate code and abstract command/command handler classes.
Thanks for the extra effort @jdrieghe !

PR #540, opened 2021-01-08
III-3672 Deserialize status from JSON in calendar on endpoints for major-info and calendar updates

LucWollants:
### Added
- [x] Serialize the top level status
- [x] Serialize the timestamp status
- [x] Validate the top level status
- [x] Validate the timestamp status

### Changed

- [x] Reuse `Timestamp` instead of `TimeSpan` inside deserializer
- [x] Stop validation of opening hours when time spans are invalid inside `CalendarForEventDataValidator`

---

Note: I first planned to modify `StartDateEndDateValidator` and `TimeSpanValidator` so they would also implement `DataValidatorInterface`. But the way they are used inside `CalendarForEventDataValidator` and `CalendarForPlaceDataValidator` are like internal methods that update the array of error messages.

---

Ticket: https://jira.uitdatabank.be/browse/III-3672


tests/Http/Deserializer/Calendar/CalendarJSONParserTest.php
```diff
@@ -93,28 +98,55 @@ public function it_returns_null_when_end_date_is_missing()
     /**
      * @test
      */
-    public function it_can_get_the_time_spans()
+    public function it_can_get_the_status()
```
jdrieghe, line 101:
```suggestion
    public function it_can_get_the_status(): void
```

jdrieghe: There are several other instances of this missing. I'm not sure if it's worth it to add them all now.

jdrieghe (approved):

tests/Http/Deserializer/Calendar/CalendarJSONParserTest.php
LucWollants: Easy enough to modify for the touched files in this pull request.

src/Http/Deserializer/Calendar/CalendarForEventDataValidator.php
```diff
@@ -29,33 +30,42 @@ public function validate(array $data)
...
+        }
+
         $messages = array_merge(
             $messages,
             (new TimeSpanValidator())->validate($data)
         );
 
-        $timeSpans = $calendarJSONParser->getTimeSpans($data);
+        // When the time spans contain errors it makes no sense to validate the rest of the data.
+        if (empty($messages)) {
```
bertramakers, line 47: Nit: An early return could avoid this big `if` body.

src/Http/Deserializer/Calendar/CalendarJSONParser.php
```diff
@@ -36,35 +28,46 @@ public function getEndDate($data)
...
                     $startDate = new \DateTime($timeSpan['start']);
                     $endDate = new \DateTime($timeSpan['end']);
-                    $timeSpans[] = new TimeSpan($startDate, $endDate);
+                    $timestamp = new Timestamp($startDate, $endDate);
+
+                    if (isset($timeSpan['status'])) {
+                        $timestamp = $timestamp->withStatus(Status::deserialize($timeSpan['status']));
+                    }
+
+                    $timestamps[] = $timestamp;
```
bertramakers, line 58: Same nit: Some early returns (or in the case of the `foreach` a `continue`) could reduce the amount of indentations here.

src/Http/Deserializer/Calendar/CalendarJSONParserInterface.php
```diff
@@ -3,34 +3,24 @@
 namespace CultuurNet\UDB3\Http\Deserializer\Calendar;
 
 use CultuurNet\UDB3\Calendar\OpeningHour;
+use CultuurNet\UDB3\Event\ValueObjects\Status;
+use CultuurNet\UDB3\Timestamp;
 
 interface CalendarJSONParserInterface
```
bertramakers, line 9: Do we even need this interface? There's only one possible implementation I think?

bertramakers (approved):
Changes look good, just some small suggestions that have more to do with the structure of the existing code that was already in place before you changed it. I'll leave it up to you to decide what's valuable to improve or not. 👍

src/Http/Deserializer/Calendar/CalendarForEventDataValidator.php
LucWollants: I think that with an early return we have to copy the throw of the validation exception twice?

This part needs to be done twice I think. Once when the status or time spans introduce errors or once when the start or end date introduce errors.
```
        if (!empty($messages)) {
            $e = new DataValidationException();
            $e->setValidationMessages($messages);
            throw $e;
        }
```

src/Http/Deserializer/Calendar/CalendarJSONParserInterface.php
LucWollants: Good remark. That was easy to remove.

src/Http/Deserializer/Calendar/CalendarForEventDataValidator.php
bertramakers: Yes, but you could move the creation of that and throwing to a private method and do something like this in both places:
```
$this->throwIfNotEmpty($messages);
```

src/Http/Deserializer/Calendar/CalendarJSONParser.php
LucWollants: That makes the code indeed much cleaner.

src/Http/Deserializer/Calendar/CalendarForEventDataValidator.php
LucWollants: Indeed! The early return makes the code better readable in combination with the helper method.
I did put the helper method `throwIfNotEmpty` inside a trait.

LucWollants:
> Changes look good, just some small suggestions that have more to do with the structure of the existing code that was already in place before you changed it. I'll leave it up to you to decide what's valuable to improve or not. 👍

It makes sense to adjust the existing code and make better use of early return statements.

jdrieghe (approved):

PR #541, opened 2021-01-12
III-3655 - Console command to bulk update event or place status

jdrieghe:
### Added
- Interactive console command to bulk update status of events for a given SAPI3 query: `bin/udb3.php event:status:update`
- Interactive console command to bulk update status of places for a given SAPI3 query: `bin/udb3.php place:status:update`

---
Ticket: https://jira.uitdatabank.be/browse/III-3655

### DIFF
--- a/app/Console/AbstractUpdateOfferStatusCommand.php
+++ b/app/Console/AbstractUpdateOfferStatusCommand.php
@@ -14,7 +14,9 @@ use CultuurNet\UDB3\Search\ResultsGenerator;
 use CultuurNet\UDB3\Search\SearchServiceInterface;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Question\Question;
 
 abstract class AbstractUpdateOfferStatusCommand extends AbstractCommand
 {
@@ -38,14 +40,11 @@ abstract class AbstractUpdateOfferStatusCommand extends AbstractCommand
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $query = 'ADD_SAPI_QUERY_HERE';
-        $status = new Status(
-            StatusType::unavailable(),
-            [
-                new StatusReason(new Language('nl'), 'Afgelast door Covid-19'),
-                new StatusReason(new Language('en'), 'Cancelled because of Covid-19'),
-            ]
-        );
+        $query = $this->askForQuery($input, $output);
+        $statusType = $this->askForStatusType($input, $output);
+        $reasons = $this->askForReasons($input, $output);
+
+        $status = new Status($statusType, $reasons);
 
         $count = $this->searchResultsGenerator->count($query);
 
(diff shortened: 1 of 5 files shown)


app/Console/AbstractUpdateOfferStatusCommand.php
```diff
@@ -0,0 +1,143 @@
...
+                    true
+                )
+            );
+
+        if (!$confirmation) {
+            return 0;
+        }
+
+        $offers = $this->searchResultsGenerator->search($query);
+        foreach ($offers as $id => $offer) {
```
LucWollants, line 73: Should we show a progress bar here?

LucWollants (approved):
I reviewed the code and tested the command locally.
The only small question is the usefulness of a progress bar?

app/Console/AbstractUpdateOfferStatusCommand.php
jdrieghe: A simple solution would be to render a dot for each finished offer?

LucWollants: Or the [Symfony Progress Bar](https://symfony.com/doc/current//components/console/helpers/progressbar.html)?

bertramakers: Good suggestion @LucWollants !

app/Console/UpdateEventStatusCommand.php
```diff
@@ -0,0 +1,19 @@
...
+
+class UpdateEventStatusCommand extends AbstractUpdateOfferStatusCommand
+{
+    protected function configure(): void
+    {
+        $this->offerType = OfferType::EVENT();
+
+        $this
+            ->setName('event:status:update')
+            ->setDescription('Batch update status of events through SAPI 3 query.');
```
bertramakers, line 17: Nit: You could also do this in the `AbstractUpdateOfferStatusCommand` if you want and make it concrete instead of abstract. For example something like this:
```
public function __construct(
    string $offerType,
    CommandBusInterface $commandBus,
    SearchServiceInterface $searchService
) {
    parent::__construct($commandBus);
	
    $this->setName($offerType . ':status:update');
    $this->setDescription(Batch update status of ' . $offerType . 's through SAPI 3 query.);
	
    $this->commandBus = $commandBus;
    $this->searchResultsGenerator = new ResultsGenerator(
        $searchService,
        ['created' => 'asc'],
        100
    );
}
```

The command base class of Symfony Console doesn't really do anything noteworthy before calling `configure()`. And calling `setName()` and `setDescription()` in `configure()` is basically the same as calling it in the constructor:
```
public function __construct($name = null)
    {
        $this->definition = new InputDefinition();

        if (null !== $name) {
            $this->setName($name);
        }

        $this->configure();

        if (!$this->name) {
            throw new LogicException(sprintf('The command defined in "%s" cannot have an empty name.', \get_class($this)));
        }
    }
```

I think they only do it like this (with `configure()`) so you wouldn't overwrite the constructor and potentially break the check that name should be required, while also preventing a thousand arguments in the constructor on the base class to configure it through that.

The reason I'm suggesting this is so you could get rid of the `abstract` if you want. But it's also okay like this for me since the child classes do very little aside from the setting the name and description.

bertramakers (approved):

app/Console/UpdateEventStatusCommand.php
jdrieghe: I like that. Will clean it up today and add the statusbar too.

bertramakers (approved):
Very nice @jdrieghe 👏

PR #544, opened 2021-01-21
Fix UpdateOfferStatus CLI command

jdrieghe:
### Fixed
- Late refactoring to the update offer status CLI command left it in a state where some required properties were unset. This is now fixed.


LucWollants (approved):
I had the same issue with the `ReindexOffersWithPopularityScore` command, see
https://github.com/cultuurnet/udb3-silex/blob/master/app/Console/ReindexOffersWithPopularityScore.php#L54

PR #545, opened 2021-01-21
Improve ci pipeline

jdrieghe:
### Changed
- Updated PHPCS to latest version

### Removed
- Replaced phing with more granular CI checks

### Fixed
- Coding standard violations on some files
- Incorrect namespace for search


composer.lock
```diff
@@ -8174,52 +8174,6 @@
             ],
             "time": "2018-02-22T13:58:36+00:00"
         },
-        {
-            "name": "cultuurnet/search",
```
bertramakers, line 8178: @jdrieghe this is the actual cause of the "incorrect" search namespace. This is an older package that was mostly used for SAPI2, but it seems like udb3-silex still uses some of its value objects. While we should refactor that to use other value objects instead so we can indeed remove this package, I don't think is as easy as replacing the namespaces with`SearchV3`. The reason the package got removed in your `composer up` is because I removed it from `udb3-php` a week ago because it was unused there. https://github.com/cultuurnet/udb3-php/commit/8c4ea70bc6a6ace57019d30155bd6935031646f7#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L28 
I just didn't realize that it was used in `udb3-silex` _and_ not directly depended upon.

bertramakers (changes requested):

composer.lock
bertramakers: I think we have two options:

1. Add `"cultuurnet/search": "~1.1"` as a direct dependency here and add a ticket to remove it later when we have time to look into what classes are still used and how we can replace those.
2. Keep it like this but then we need to make a list of all the functionality that is impacted by this change and have Corneel test that extensively to make sure nothing is broken.

jdrieghe: I'm technically more a fan of the latter. The functionality we use is covered with tests that are green and it looked like the used value objects were almost identical. What big risk do you see still? We only use the string value of the Query VO to perform a search on Sapi3 in this appliation.

bertramakers: Ah I was mistaken, I thought the usages where in `app` which is not tested. But they're in `src` which is indeed covered by unit tests. And I looked up the usage and it basically gets converted to a string using `getValue()` which coincidentally is also a method on the `V3` class. So I agree that we can leave it out 👍

bertramakers (approved):
Nice!

PR #546, opened 2021-01-25
Fix update status of udb2 imported events

jdrieghe:
### Fixed
- Updating the status of events and places imported from UDB2 was not possible because their calendars weren't set properly in the aggregate

---
Ticket: https://jira.uitdatabank.be/browse/III-3655


bertramakers (approved):
Very clear, thanks for the extra comments!

PR #548, opened 2021-01-29
III-3691 Add ability to change the owner of events/places

bertramakers:
### Added

- Added two CLI commands to change the owner of events/places. Either for one event/place based on its id, or for all events/places that belong to a specific user id.

---
Ticket: https://jira.uitdatabank.be/browse/III-3691

Context is described in the ticket + comments, but a summary:
- When we integrated with Auth0, we refactored UDB3 to determine the user id as follows: If the user has a v1 id in their token, use that to make sure their roles, permissions, entered events/places etc keep working. Also create new events/places with this v1 id. So basically always use the v1 id if there is one. If there's no v1 id, use the v2 (= Auth0) id.
- At one point, there was a bug in Auth0 which caused users that were migrated from v1 and used social media to login to lose their v1 id.
- These users then created events/places with a v2 id. (And complained that they couldn't see or edit their old events/places).
- The bug was fixed in Auth0 and those users now have a v1 id again. However, they now cannot see or edit their events/places created with their v2 id.
- The initial idea of Corneel & Manon was to change the query on the dashboard in the frontend, to also look for events/places created with the v2 id. However that idea would not make it possible to actually edit the events/places.
- The preferred solution now is to have a way to change the "creator"/owner of the events/places to the correct id. So the `creator` in the JSON-LD has to be updated (so the event/place appears on the dashboard), and the user id inside the permission tables need to change so they can actually edit/delete the event/place.

### DIFF
--- a/app/Console/ChangeOfferOwnerInBulk.php
+++ b/app/Console/ChangeOfferOwnerInBulk.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace CultuurNet\UDB3\Silex\Console;
+
+use Broadway\CommandHandling\CommandBusInterface;
+use CultuurNet\UDB3\Offer\Commands\ChangeOwner;
+use CultuurNet\UDB3\Offer\ReadModel\Permission\PermissionQueryInterface;
+use Error;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Logger\ConsoleLogger;
+use Symfony\Component\Console\Output\OutputInterface;
+use ValueObjects\StringLiteral\StringLiteral;
+
+class ChangeOfferOwnerInBulk extends AbstractCommand
+{
+    /**
+     * @var PermissionQueryInterface
+     */
+    private $permissionQuery;
+
+    public function __construct(
+        CommandBusInterface $commandBus,
+        PermissionQueryInterface $permissionQuery
+    ) {
+        parent::__construct($commandBus);
+        $this->permissionQuery = $permissionQuery;
+    }
+
+    public function configure(): void
+    {
+        $this->setName('offer:change-owner-bulk');
+        $this->setDescription('Change the owner of multiple offers to a new user id');
+        $this->addArgument('originalOwnerId', InputArgument::REQUIRED, 'Id of the original owner');
+        $this->addArgument(
+            'newOwnerId',
+            InputArgument::REQUIRED,
+            'Id of the new user. '
+            . 'Can either be a v1 id (e.g. "97f0f81d-a2b6-44c5-ab27-e076d6329f91") '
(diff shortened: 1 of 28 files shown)


src/Offer/Events/AbstractOwnerChanged.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @var string
+     */
+    private $offerId;
+
+    /**
+     * @var string
+     */
+    private $newCreator;
```
LucWollants, line 19: I think the creator becomes the owner? Maybe better to keep the naming consistent and pick either creator or owner?

src/Offer/ReadModel/Permission/Doctrine/DBALRepository.php
```diff
@@ -85,4 +85,13 @@ public function markOfferEditableByUser(StringLiteral $eventId, StringLiteral $u
             // permission record is already in place.
         }
     }
+
+    public function markOfferEditableByNewUser(StringLiteral $eventId, StringLiteral $userId): void
```
LucWollants, line 89: Something very minor, but I wonder if "NewOwner" would be a better name than "NewUser"?

app/Console/ChangeOfferOwnerInBulk.php
```diff
@@ -0,0 +1,89 @@
...
+                    )
+                );
+                $logger->info(
+                    'Successfully changed owner of offer "' . $offerId .'" to user with id "' . $newOwnerId . '"'
+                );
+                $success++;
+            } catch (Error $error) {
+                $logger->error(
+                    sprintf(
+                        'An error occurred while chaging owner of offer "%s": %s with message %s',
```
LucWollants, line 71: Minor typo here, "chaging" should be "changing".

LucWollants (approved):
A very nice pull request with very clear comments.
I already approved because my remarks are minor ones about naming.

app/Console/ChangeOfferOwnerInBulk.php
LucWollants:
```suggestion
                        'An error occurred while changing owner of offer "%s": %s with message %s',
```

src/Offer/Events/AbstractOwnerChanged.php
bertramakers: Yep, it's a leftover from when I called it `CreatorUpdated` first (didn't commit that because I felt that was a weird name) 👍

src/Offer/ReadModel/Permission/Doctrine/DBALRepository.php
LucWollants: You are right, in context as the interface of `DBALRepository` the `newUser` makes more sense. Keep the method name as is.

bertramakers:
Thanks @LucWollants !

PR #549, opened 2021-01-29
III-3732 Add UiTPAS labels config

bertramakers:
### Changed

- UiTPAS labels now get read from config instead of fetched via an HTTP request

---
Ticket: https://jira.uitdatabank.be/browse/III-3732

Related udb3-vagrant PR: https://github.com/cultuurnet/udb3-vagrant/pull/40
Related appconfig PR: https://github.com/cultuurnet/appconfig/pull/40


LucWollants:
I wonder if it would be cleaner and easier to maintain when the labels are stored inside a dedicated YAML file as we do for `user_persmissions`?

jdrieghe (approved):

jdrieghe:
Don't really have a preference about the dedicated yaml file. Configs can always be moved around when we feel like it becomes a problem.

bertramakers:
I agree with @jdrieghe , if it becomes too big it's pretty easy to move it later.

PR #551, opened 2021-02-01
III-3709 Internalize the udb3-php package

LucWollants:
### Added
- Added all the code from [udb3-php package](https://github.com/cultuurnet/udb3-php) into the `src` folder

---
Ticket: https://jira.uitdatabank.be/browse/III-3709


composer.lock
```diff
@@ -974,23 +983,23 @@
         },
         {
             "name": "cultuurnet/culturefeed-php",
-            "version": "dev-master",
+            "version": "1.9.2",
```
bertramakers, line 986: I'm wondering if this won't cause issues, going back from `dev-master` to `1.9.2`?

Looking at the commit references changed, it goes back from this commit on November 25th 2019 https://github.com/cultuurnet/culturefeed-php/commit/8b70c4a4d96450a8f292195dad78720496c14394 to this commit on November 5th 2019: https://github.com/cultuurnet/culturefeed-php/commit/583c4a0c82a7f0f5a8c60df5f2d3c2b10417ecc1

Since the dependency is not changed in the composer.json in this repo, I'm guessing it was set to `dev-master` in a nested dependency that's now changed.

bertramakers (changes requested):
Looks good overall, just a bit concerned about the sudden reverse to an older commit in the `culturefeed-php` dependency

composer.lock
bertramakers: There are newer version available than `1.9.2`, which in theory should not break backward compatibility, so maybe it's as simple as updating to the latest version:

https://github.com/cultuurnet/culturefeed-php/tags

v1.10 is the one with the same commit we were on before: https://github.com/cultuurnet/culturefeed-php/releases/tag/1.10

But there's already a newer v1.12: https://github.com/cultuurnet/culturefeed-php/releases/tag/1.12

LucWollants: I will set it to `dev-master` inside composer.json file and run a `composer update cultuurnet/culturefeed-php`.

bertramakers (approved):

jdrieghe (approved):

PR #552, opened 2021-02-01
III-3733 Add uitpas service endpoints

bertramakers:
### Added

- Added `/uitpas/events/` endpoints from udb3-uitpas-service
- Added `/uitpas/organizers/` endpoints from udb3-uitpas-service
- Added better error handling for CultureFeed exceptions, based on udb3-uitpas-service
- Added 401 status code for response when a token is missing in a request that needs one, using a custom "EntryPoint" in the firewall configuration (also based on udb3-uitpas-service and an improvement over the default redirect to `/login` which doesn't exist)

### Changed

- Refactored code of existing `/uitpas/labels/` endpoint a bit for consistency

---
Ticket: https://jira.uitdatabank.be/browse/III-3733

### DIFF
--- a/app/ErrorHandlerProvider.php
+++ b/app/ErrorHandlerProvider.php
@@ -3,6 +3,8 @@
 namespace CultuurNet\UDB3\Silex;
 
 use Crell\ApiProblem\ApiProblem;
+use CultureFeed_Exception;
+use CultureFeed_HttpException;
 use CultuurNet\Deserializer\DataValidationException;
 use CultuurNet\UDB3\EntityNotFoundException;
 use CultuurNet\UDB3\HttpFoundation\Response\ApiProblemJsonResponse;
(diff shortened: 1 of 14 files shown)


app/ErrorHandlerProvider.php
```diff
@@ -66,6 +82,29 @@ private function createNewApiProblem(Exception $e): ApiProblem
...
+     * implemented a comparison on the error message when this was introduced in udb3-uitpas-service in the past.
+     */
+    private function createNewApiProblemFromCultureFeedException(Exception $exception): ApiProblem
+    {
+        $apiProblem = $this->createNewApiProblem($exception);
+
+        $title = $apiProblem->getTitle();
+
+        // Remove "URL CALLED" and everything after it.
+        $formattedTitle = preg_replace('/URL CALLED.*/', '', $title);
```
jdrieghe, line 101: Is this covered by a test somewhere?

jdrieghe (approved):
Looks good. Only possible improvement I see right now is to cover the regex with a test to make sure it does what we expect.

app/ErrorHandlerProvider.php
bertramakers: No, it was also not tested before in `udb3-uitpas-service`, because we have no tests for anything in `app` since it's hard to test these controller/service providers.

I could extract it to a separate class and add tests for that, but I feel like that would shift focus from consolidating & archiving as much repositories as possible to adding more test coverage for existing functionality.

I tested it manually though just in case. It's easy to see in action on a URL like `https://io.uitdatabank.dev/uitpas/events/bla`

jdrieghe:
🚢  it!

PR #553, opened 2021-02-02
III-3710 Internalize the udb3-models dependency

LucWollants:
### Changed

- Moved the ud3-models dependency into udb3-silex

---
Ticket: https://jira.uitdatabank.be/browse/III-3710


src/Model/Place/ImmutablePlace.php
```diff
@@ -0,0 +1,155 @@
...
+     * @param TranslatedTitle $title
+     * @param TranslatedAddress $address
+     * @return ImmutablePlace
+     */
+    public static function createDummyLocation(
+        Language $mainLanguage,
+        TranslatedTitle $title,
+        TranslatedAddress $address
+    ) {
+        /** @phpstan-ignore-next-line */
```
bertramakers, line 125: Nit: I think this one is easy to fix by making the class and/or constructor private

src/Model/ValueObject/Contact/BookingAvailability.php
```diff
@@ -0,0 +1,39 @@
...
+
+class BookingAvailability extends DateTimeImmutableRange
+{
+    /**
+     * @param \DateTimeImmutable $from
+     * @return static
+     */
+    public static function from(\DateTimeImmutable $from)
+    {
+        /** @phpstan-ignore-next-line */
```
bertramakers, line 15: Nit: Same as `ImmutablePlace`, making the class `final` fixes this I think

src/Model/ValueObject/Contact/BookingAvailability.php
```diff
@@ -0,0 +1,39 @@
...
+        return new static($from, null);
+    }
+
+    /**
+     * @param \DateTimeImmutable $to
+     * @return static
+     */
+    public static function to(\DateTimeImmutable $to)
+    {
+        /** @phpstan-ignore-next-line */
```
bertramakers, line 25: Same nit as above

src/Model/ValueObject/Contact/BookingAvailability.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * @param \DateTimeImmutable $from
+     * @param \DateTimeImmutable $to
+     * @return static
+     */
+    public static function fromTo(\DateTimeImmutable $from, \DateTimeImmutable $to)
+    {
+        /** @phpstan-ignore-next-line */
```
bertramakers, line 36: Again same nit as above 😄

bertramakers (approved):
Looks good! Some of the ignored PHPStan errors can easily be fixed I think. Only the `Collection` is harder.

src/Model/Place/ImmutablePlace.php
LucWollants: You are right! I was so focused on the error inside `Collection` I forgot the "easy" fix to make the class `final`.
I will check what can be fixed with the `final` keyword.

PR #570, opened 2021-02-11
III-3783

jdrieghe:
### Fixed
- Importing an event with a status without reason now works as expected

---
Ticket: https://jira.uitdatabank.be/browse/III-3783


LucWollants (approved):
Nice with the extra test.

PR #571, opened 2021-02-11
III-3777

jdrieghe:
### Changed
- Dependencies can now be installed using composer v2
- CI pipeline now uses composer v2

---
Ticket: https://jira.uitdatabank.be/browse/III-3777


LucWollants (approved):
I already approved but waited with the merge to make sure we can use `composer:v2` inside the Vagrant box

LucWollants:
> I already approved but waited with the merge to make sure we can use composer:v2 inside the Vagrant box

I tested inside the Vagrant box with composer v1:
- `composer install` on the PR and that still worked
- `composer install` after deleting the vendor folder and that also worked

PR #572, opened 2021-02-11
III-3653 Update unique labels

LucWollants:
### Added
- Add script to get all `CultuurNet.UDB3.Label.Events.Created` events and add those to the `labels_unique` table
- The script works with a bulk size of 1000 records to avoid memory issues
- Make it possible to generate a simple `txt` file with a list of all processed labels

---
Ticket: https://jira.uitdatabank.be/browse/III-3653

Note: A follow-up PR will focus on updating `organizers_unique_websites`


app/Console/UpdateUniqueLabels.php
```diff
@@ -0,0 +1,143 @@
...
+                    $messages[] = 'Added label ' . $labelName->toNative() . ' with uuid ' . $labelUuid->toString();
+                } catch (UniqueConstraintViolationException $exception) {
+                    $messages[] = 'Unique exception for label ' . $labelName->toNative() . ' with uuid ' . $labelUuid->toString();
+                }
+
+                $progressBar->advance();
+            }
+
+            $offset += count($labelAddedEvents);
+        } while ($offset < $labelAddedEventsCount);
```
jdrieghe, line 76: It's been a long time since I've seen a do/while construct with a nested loop 😅

jdrieghe (approved):

app/Console/UpdateUniqueLabels.php
LucWollants: Yes, a good old `do-while` construct! Lucky it's not a `for-each` inside another `for-each` 😅

I thought about:
- working with a generator but that would require a select statement for every single label created event
- creating a repository class around the event store

But this command should only be run once. So it's not worth the effort.

PR #574, opened 2021-02-12
III-3809 Set calendar property on aggregate after MajorInfoUpdated

bertramakers:
### Fixed

- Fixed status not being applied correctly or reverting changes made after a MajorInfoUpdated

---
Ticket: https://jira.uitdatabank.be/browse/III-3809


jdrieghe (approved):
Yep, we missed this one. 👍

PR #579, opened 2021-02-17
III-3815

jdrieghe:
### Fixed
- Embedded locations with incorrect status format are now polyfilled to the new nested status format

---
Ticket: https://jira.uitdatabank.be/browse/III-3815

### DIFF
--- a/src/Offer/ReadModel/JSONLD/NewPropertyPolyfillOfferRepository.php
+++ b/src/Offer/ReadModel/JSONLD/NewPropertyPolyfillOfferRepository.php
@@ -82,7 +82,7 @@ final class NewPropertyPolyfillOfferRepository extends DocumentRepositoryDecorat
 
     private function polyfillEmbeddedPlaceStatus(array $json): array
     {
-        if (!isset($json['location']['status']) || isset($json['location']['status']['type']) {
+        if (!isset($json['location']['status']) || isset($json['location']['status']['type'])) {
             return $json;
         }
 
--- a/src/Offer/ReadModel/JSONLD/NewPropertyPolyfillOfferRepository.php
+++ b/src/Offer/ReadModel/JSONLD/NewPropertyPolyfillOfferRepository.php
@@ -82,13 +82,17 @@ final class NewPropertyPolyfillOfferRepository extends DocumentRepositoryDecorat
 
     private function polyfillEmbeddedPlaceStatus(array $json): array
     {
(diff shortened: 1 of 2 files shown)


src/Offer/ReadModel/JSONLD/NewPropertyPolyfillOfferRepository.php
```diff
@@ -78,4 +79,17 @@ function (array $subEvent) {
 
         return $json;
     }
+
+    private function polyfillEmbeddedPlaceStatus(array $json): array
+    {
+        if (!isset($json['location']['status']) || isset($json['location']['status']['type'])) {
```
bertramakers, line 85: Shouldn't we set a default `Available` if there's no status? Or does the calendar summary not need it? I assumed it did because it was crashing on the incorrectly formatted status.

jdrieghe: The calendar summary doesn't care about the status of the location, but deserialization indeed crashes if the format is wrong

jdrieghe: We can always add it for completeness' sake though

bertramakers: Makes sense!

LucWollants: The default status is always set on the place and when the event gets projected the place is embedded with a status inside the event projection.
But the issue was temporary, only during the timeframe that status got projected without type.

jdrieghe: Does that mean I should revert the last commit?

LucWollants: I don't think it is needed...

bertramakers: I think we need it, for example for events who's place has not been explicitly updated.
See an example here: https://io.uitdatabank.be/event/03d993c3-cee4-4394-bca5-180fc5b05027
The event gets a status polyfilled, but the place doesn't because it's cached on the event JSON-LD and the place doesn't get a status polyfilled.

bertramakers (approved):

PR #580, opened 2021-02-17
III-3653 Update unique organizers

LucWollants:
### Added
- Add script to update the unique websites of organizers based on the events `OrganizerCreatedWithUniqueWebsite` and `WebsiteUpdated`

Note: An extra select statement was added to verify what is already inside the unique table. This makes it possible to do an update or insert. It also makes it possible to just skip an organizer that is already correctly stored inside the unique table.

---
Ticket: https://jira.uitdatabank.be/browse/III-3653

### DIFF
--- a/app/Console/UpdateUniqueOrganizers.php
+++ b/app/Console/UpdateUniqueOrganizers.php
@@ -129,7 +129,7 @@ class UpdateUniqueOrganizers extends Command
         // 2. The organizer uuid is not present
         // 3. The organizer uuid is present with another value
 
-        $existingOrganizerUrl = $this->connection->createQueryBuilder()
+        $existingOrganizerUrls = $this->connection->createQueryBuilder()
             ->select('unique_col')
             ->from('organizer_unique_websites')
             ->where('uuid_col = :uuid')
@@ -137,6 +137,7 @@ class UpdateUniqueOrganizers extends Command
             ->execute()
             ->fetchAll(PDO::FETCH_COLUMN);
 
+        $existingOrganizerUrl = \count($existingOrganizerUrls) === 1 ? $existingOrganizerUrls[0] : null;
         if ($existingOrganizerUrl === (string) $organizerUrl) {
             return;
         }
--- a/app/Console/UpdateUniqueOrganizers.php
+++ b/app/Console/UpdateUniqueOrganizers.php
@@ -63,11 +63,11 @@ class UpdateUniqueOrganizers extends Command
                 $organizerUuid = $this->getOrganizerUuid($organizerEvent);
                 $organizerUrl = $this->getOrganizerWebsite($organizerEvent);
 
-                try {
-                    $this->updateOrganizer($organizerUuid, $organizerUrl);
-                    $messages[] = 'Added organizer ' . $organizerUrl . ' with uuid ' . $organizerUuid->toString();
-                } catch (UniqueConstraintViolationException $exception) {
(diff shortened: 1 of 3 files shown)


app/Console/UpdateUniqueOrganizers.php
```diff
@@ -0,0 +1,179 @@
...
+                ]
+            );
+    }
+
+    private function getOrganizerUuid(array $organizerCreatedEvent): Uuid
+    {
+        return Uuid::fromString($organizerCreatedEvent['uuid']);
+    }
+
+    private function getOrganizerWebsite(array $organizerCreatedEvent): Url
```
bertramakers, line 174: Nit: The parameter should probably be `$organizerEvent`? Since `WebsiteUpdated` is not "Organizer created". (Also in the `getOrganizerUuid()` method)

app/Console/UpdateUniqueOrganizers.php
```diff
@@ -0,0 +1,179 @@
...
+    {
+        return $this->connection->createQueryBuilder()
+            ->select('uuid, payload')
+            ->from('event_store')
+            ->where('type = "' . self::ORGANIZER_CREATED . '"')
+            ->orWhere('type = "' . self::ORGANIZER_WEBSITE_UPDATED . '"')
+            ->setFirstResult($offset)
+            ->setMaxResults(self::MAX_RESULTS)
+            ->execute()
+            ->fetchAll(PDO::FETCH_ASSOC);
```
bertramakers, line 106: Should we order the results? Are we sure they will always be in the same order when we fetch the 2nd, 3rd, etc set of 1000 results?

LucWollants: Good remark! Better to add an `orderBy` to make sure we get all the events.
I will also add an `orderBy` on the query for the labels.

LucWollants:
I also noticed I made 2 mistakes:
1. Assuming the result from `fetchAll` inside `updateOrganizer` was a string. This is an array and the code was updated.
2. Assuming that `UniqueConstraintViolationException` is still thrown inside `updateOrganizer`. This was fixed by returning a boolean from `updateOrganizer`

app/Console/UpdateUniqueOrganizers.php
```diff
@@ -129,14 +129,15 @@ private function updateOrganizer(Uuid $organizerUuid, Url $organizerUrl): void
...
-        $existingOrganizerUrl = $this->connection->createQueryBuilder()
+        $existingOrganizerUrls = $this->connection->createQueryBuilder()
             ->select('unique_col')
             ->from('organizer_unique_websites')
             ->where('uuid_col = :uuid')
             ->setParameter('uuid', $organizerUuid->toString())
             ->execute()
             ->fetchAll(PDO::FETCH_COLUMN);
 
+        $existingOrganizerUrl = \count($existingOrganizerUrls) === 1 ? $existingOrganizerUrls[0] : null;
```
bertramakers, line 139: Nit: You don't need the `\` in `\count()` here. And most of the existing code doesn't use FQN for functions in the root namespace.

bertramakers (approved):
Looks great!  👏

app/Console/UpdateUniqueOrganizers.php
LucWollants: Indeed! Must be something added by the new version of PHPStorm.

PR #582, opened 2021-02-18
III-3771 Upgrade to Silex v1.3.6

LucWollants:
### Changed
- Upgrade to `broadway/broadway` v1
- Upgrade to `cultuurnet/udb3-api-guard` v2
- Upgrade to `willdurand/geocoder` v3
- Indirect upgrade to `cultuurnet/valueobjects` v4 through `cultuurnet/udb3-api-guard`
- Upgrade to `silex/silex` v1.3.6 to allow upgrade of `symfony/event-dispatcher` to v3
- Indirect upgrade to Symfony v3 because of the Silex upgrade

### Removed
- Removed `DBALEventStore`
- Removed `deralex/yaml-config-service-provider` dependency

### Added
- Added implementation for `loadFromPlayhead` which makes it possible to load a stream from a certain `playhead` offset
- Added custom implementation for YamlConfigServiceProvider
- Add friendsofphp/php-cs-fixer dependency

---
Ticket: https://jira.uitdatabank.be/browse/III-3771


tests/Role/ReadModel/Users/RoleUsersProjectorTest.php
```diff
@@ -283,11 +283,6 @@ private function mockGetUserById(
             ->willReturn($userIdentityDetails);
     }
 
-    /**
-     * @param UUID $uuid
-     * @param SerializableInterface $payload
-     * @return DomainMessage
-     */
```
bertramakers, line 290: Nit: The return type is now not documented or defined in the code itself

tests/Role/ReadModel/Users/UserRolesProjectorTest.php
```diff
@@ -362,11 +362,6 @@ public function it_does_not_update_projection_when_role_users_not_found_on_role_
         $this->userRolesProjector->handle($domainMessage);
     }
 
-    /**
-     * @param UUID $uuid
-     * @param SerializableInterface $payload
-     * @return DomainMessage
-     */
```
bertramakers, line 369: Nit: The return type is now not documented or defined in the code itself

tests/Label/ReadModels/Relations/ProjectorTest.php
```diff
@@ -476,7 +476,7 @@ private function getRelationId()
         return 'E4CA9DB5-DEE3-42F0-B04A-547DFC3CB2EE';
     }
 
-    private function createDomainMessage(string $id, Serializable $payload): DomainMessage
+    private function createDomainMessage($id, SerializableInterface $payload)
```
bertramakers: Why did you remove the `string` param type and `DomainMessage` return type again here?

composer.json
```diff
@@ -14,6 +14,8 @@
     "ext-json": "*",
     "ext-pdo": "*",
     "ext-tidy": "*",
+    "ext-dom": "*",
+    "ext-xmlreader": "*",
```
bertramakers, line 18: 👍

app/Console/UpdateUniqueLabels.php
```diff
@@ -130,7 +131,7 @@ private function updateLabel(Uuid $labelUuid, LabelName $labelName): void
             );
     }
 
-    private function getLabelUuid(array $labelAddedEvent): Uuid
+    private function getLabelUuid(array $labelAddedEvent): UuidInterface
```
bertramakers, line 134: Nit: We can probably just use the `Uuid` class as return type here as we did before? We also use it everywhere else in this class instead of `UuidInterface`

src/Broadway/AMQP/EventBusForwardingConsumerFactory.php
```diff
@@ -71,15 +62,10 @@ public function __construct(
...
-    /**
-     * @param StringLiteral $exchange
-     * @param StringLiteral $queue
-     * @return EventBusForwardingConsumer
-     */
     public function create(
         StringLiteral $exchange,
         StringLiteral $queue
-    ) {
+    ):EventBusForwardingConsumer {
```
bertramakers, line 68: Super small nit: Missing a space here between `:` and `EventBusForwardingConsumer` 😛

bertramakers (approved):
Wow very nicely done! 👏

tests/Label/ReadModels/Relations/ProjectorTest.php
bertramakers: The `DomainMessage` return type was re-added in a later commit I see, but the `string` is still gone from the final diff

LucWollants: That was indeed a mistake because of doing partial staging to have better commits. It fixed it.

app/Console/UpdateUniqueLabels.php
LucWollants: The issue is that the `Uuid::fromString` returns a `UuidInterface`

src/Broadway/AMQP/EventBusForwardingConsumerFactory.php
LucWollants: Fixed

jdrieghe (approved):
Approved as The Pairing Buddy™

jdrieghe (approved):

jdrieghe:
@LucWollants for that last comment, there's probably an improvement we can do using https://docs.guzzlephp.org/en/stable/testing.html

LucWollants:
Indeed, but I did not change this because https://github.com/php-http/mock-client is maybe another option?

PR #587, opened 2021-02-23
III-3677 Prevent truncated production search

LucWollants:
### Changed

- Using `GROUP_CONCAT` can lead to clipping when the result is bigger than the column size. This was removed and an iteration is used to create the productions which their events.

---
Ticket: https://jira.uitdatabank.be/browse/III-3677


jdrieghe (approved):
Nice!

PR #589, opened 2021-02-23
III-3608 Refactor label commands & handlers

bertramakers:
### Changed

- Automatic creation of label aggregates for new labels now happens in the relevant command handlers instead of editing services
- `AddLabel`, `RemoveLabel`, and `ImportLabels` commands are now handled by individual command handlers instead of `OfferCommandHandler`
- `AddLabel`, `RemoveLabel`, and `ImportLabels` commands for offers are now concrete final classes instead of abstract

---

Todo:

- [x] Move automatic creation of label aggregates to relevant command handlers instead of offer editing services
- [x] Extract command handler logic for label commands
- [x] Remove abstractions in label commands
- [x] Remove `OfferEditingServiceInterface::addLabel()`
- [x] Remove `OfferEditingServiceInterface::removeLabel()`
- [x] Remove `ExternalOfferEditingServiceInterface::addLabel()` -> **I will do this in another PR** #591
- [x] Do the same cleanup for organizers -> **I will do this in another PR**

Functional tests:
I've manually tested adding/removing visible and hidden labels to events and place, but I'll ask Corneel to also test the following manually to be sure:
- Importing labels
- UiTPAS labels that get added automatically
- Curator labels that get added automatically

Bulk labelling should not be impacted by this as this happens over HTTP (for now, I'm going to refactor that in a follow-up PR to get rid of another service interface). Edit: See #591


bootstrap.php
```diff
@@ -648,6 +649,9 @@ function ($app) {
         $commandBus->subscribe($app[ProductionCommandHandler::class]);
         $commandBus->subscribe($app[UpdateStatusHandler::class]);
         $commandBus->subscribe($app[ChangeOwnerHandler::class]);
+        $commandBus->subscribe($app[AddLabelHandler::class]);
```
LucWollants, line 652: I remember that the order of subscribing could have some unwanted side effects. Is that order also important for labels? Or was that more an issue with the event listeners and not the command handlers?

src/Http/Offer/EditOfferRestController.php
```diff
@@ -89,11 +88,7 @@ public function __construct(
 
     public function addLabel(string $cdbid, string $label): Response
     {
-        $this->editService->addLabel(
-            $cdbid,
```
LucWollants, line 93: Not directly changed and not important in this PR, but do we stick with `cdbid` as a term?

bertramakers: No I'd rather just use `id` or `eventId` etc.

src/Offer/CommandHandlers/ImportLabelsHandler.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->labelService = $labelService;
+    }
+
+    public function handle($command): void
+    {
+        if (!($command instanceof ImportLabels)) {
+            return;
+        }
+
+        /** @var \CultuurNet\UDB3\Model\ValueObject\Taxonomy\Label\Label $importLabel */
```
LucWollants, line 39: Maybe import the class? But I guess you used the FQN because it's inside a comment. (I have to find some remarks 😉)

bootstrap.php
bertramakers: You remembered correctly, because you documented that yourself here 😉 https://github.com/cultuurnet/udb3-silex/blob/master/bootstrap.php#L593-L595

It's not clear to me why that is important though, since they listen to different commands. The label command handler handles commands from the `CultuurNet\UDB3\Label` namespace, and the new ones (and the older offer ones) handle commands from the event/place/offer namespace. So the order should not matter.

bertramakers: In any case, I tested it manually and I can add unexisting labels without issue

src/Offer/CommandHandlers/AddLabelHandler.php
```diff
@@ -0,0 +1,70 @@
...
+use CultuurNet\UDB3\Label\ValueObjects\LabelName;
+use CultuurNet\UDB3\Label\ValueObjects\Visibility;
+use CultuurNet\UDB3\Offer\Commands\AddLabel;
+use CultuurNet\UDB3\Offer\OfferRepository;
+use ValueObjects\StringLiteral\StringLiteral;
+
+final class AddLabelHandler implements CommandHandler
+{
+    /**
+     * @var OfferRepository
```
LucWollants, line 20: This class already proofs its value.

src/Offer/CommandHandlers/ImportLabelsHandler.php
bertramakers: Yeah I can import it, but then I'll alias it as `Udb3ModelLabel`. There's no name conflict in this file, but the multiple `Label` value objects are pretty confusing. (We have three of them!)

Normally the one from udb3-models would not have been used in these domain classes, but it seems like the `ImportLabels` command accidentally uses the udb3-models Label object instead of the `CultuurNet\UDB3\Label` object

src/Offer/CommandHandlers/AddLabelHandler.php
bertramakers: Yes this makes it possible to simplify the command handling a lot!

bertramakers: I'm going to add a ticket to clean up the other offer commands & handlers like this in the near future

bertramakers: I already did the refactoring for these ones because I was already touching the label handlers to make sure the new label aggregates always get created in the handler instead of a service that can be circumvented by accident

tests/Event/EventCommandHandlerTest.php
```diff
@@ -35,25 +32,16 @@
...
-use CultuurNet\UDB3\Model\ValueObject\Taxonomy\Label\Labels;
 use CultuurNet\UDB3\OfferCommandHandlerTestTrait;
 use CultuurNet\UDB3\PriceInfo\BasePrice;
 use CultuurNet\UDB3\PriceInfo\Price;
 use CultuurNet\UDB3\PriceInfo\PriceInfo;
 use CultuurNet\UDB3\Theme;
 use CultuurNet\UDB3\Title;
-use ValueObjects\Identity\UUID;
 use ValueObjects\Money\Currency;
-use ValueObjects\StringLiteral\StringLiteral;
```
LucWollants, line 56: I noticed some extra unused imports
```
use CultuurNet\UDB3\Event\Events\LabelAdded;
use CultuurNet\UDB3\Event\Events\LabelRemoved;
use CultuurNet\UDB3\Event\Events\LabelsImported;
```

bertramakers: Woops!

bertramakers: Fixed in 08eb1e1

src/Offer/CommandHandlers/ImportLabelsHandler.php
bertramakers: Fixed in adbb968

bootstrap.php
bertramakers: I'm going to check if I can still add a new label to an event/place if I move the command handler you put at the top at the end 🤔

tests/Offer/CommandHandlers/AddLabelHandlerTest.php
```diff
@@ -0,0 +1,131 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use ValueObjects\Identity\UUID;
+use ValueObjects\StringLiteral\StringLiteral;
+
+final class AddLabelHandlerTest extends CommandHandlerScenarioTestCase
+{
+    /**
+     * @var ReadRepositoryInterface|MockObject
+     */
+    private $labelRepository;
```
LucWollants, line 36: I think this can be replaced with a local variable?

bertramakers: Right! Fixed in 396f9d5

tests/Offer/CommandHandlers/AddLabelHandlerTest.php
```diff
@@ -0,0 +1,131 @@
...
+            ),
+            $this->labelService,
+            $this->labelRepository
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_should_use_existing_visibility_for_existing_labels(): void
```
LucWollants, line 69: Do we need a bit more context in the name about "label adding" or that is implied from the handler?

bertramakers: I figured it would be clear enough since the handler only has one responsibility

tests/Offer/CommandHandlers/RemoveLabelHandlerTest.php
```diff
@@ -0,0 +1,79 @@
...
+    public function it_should_remove_labels_previously_added_even_if_the_visibility_is_different(): void
+    {
+        $id = '4c6d4bb8-702b-49f1-b0ca-e51eb09a1c19';
+
+        $this->scenario
+            ->withAggregateId($id)
+            ->given(
+                [
+                    $this->eventCreated($id),
+                    new LabelAdded($id, new Label('foo', true)),
```
LucWollants, line 47: We also need to test the removal of a label with the same visibility?

tests/Place/PlaceCommandHandlerTest.php
```diff
@@ -14,20 +14,15 @@
 use CultuurNet\UDB3\CalendarType;
 use CultuurNet\UDB3\Description;
 use CultuurNet\UDB3\Event\EventType;
-use CultuurNet\UDB3\Label;
```
LucWollants, line 17: Some extra imports to remove
```
use CultuurNet\UDB3\Label\LabelServiceInterface;
use CultuurNet\UDB3\Label\ReadModels\JSON\Repository\Entity;
use CultuurNet\UDB3\Label\ReadModels\JSON\Repository\ReadRepositoryInterface;
use CultuurNet\UDB3\Label\ValueObjects\Privacy;
use CultuurNet\UDB3\Label\ValueObjects\Visibility;

use ValueObjects\Identity\UUID;

use ValueObjects\StringLiteral\StringLiteral;
```

LucWollants:
I noticed a wrong class or file name:
- file: tests/Place/PlaceCommandHandlerTest.php
- class: PlaceHandlerTest

src/Offer/CommandHandlers/ImportLabelsHandler.php
LucWollants: > Normally the one from udb3-models would not have been used in these domain classes, but it seems like the `ImportLabels` command accidentally uses the udb3-models Label object instead of the `CultuurNet\UDB3\Label` object

Maybe worth changing in a follow-up PR or a ticket?

tests/Offer/CommandHandlers/RemoveLabelHandlerTest.php
bertramakers: I figured that would also work then, but here's an additional test 😉 94aeb27

tests/Place/PlaceCommandHandlerTest.php
bertramakers: The first 5 are already removed in a later commit, the last 2 are removed here: d1009c8

bertramakers:
> I noticed a wrong class name:
> 
>     * file: tests/Place/PlaceCommandHandlerTest.php
> 
>     * class: PlaceHandlerTest
> 
> 
> This has been wrong from the start, but maybe worth fixing?



> I noticed a wrong class name:
> 
>     * file: tests/Place/PlaceCommandHandlerTest.php
> 
>     * class: PlaceHandlerTest
> 
> 
> This has been wrong from the start, but maybe worth fixing?

Both the filename and class name were wrong actually, I fixed them here: 1e87c14

LucWollants (approved):
Nice work!  💯

src/Offer/CommandHandlers/ImportLabelsHandler.php
bertramakers: I'd like to do a larger refactoring (in chunks) to use the udb3-models value objects _everywhere_, so we can also get rid of cultuurnet/valueobjects. But I'm not sure how to create clearly scoped tickets for that that we can pick up in chunks instead of one big ticket that takes weeks.

bertramakers: I think the refactoring of the abstractions in the commands & handlers is unrelated though (but yes fully agreed that the valueobjects should also be made more consistent one way or another)

bertramakers: I'll try to carve out some time later this week to try and create some `TechnicalImprovement` tickets with clearly defined scopes for these refactorings

bertramakers:
> Nice work! 💯

Nice reviewing! 😄

bootstrap.php
bertramakers: See 81125c0 (and the commit message for more info)

jdrieghe (approved):

PR #591, opened 2021-02-24
III-3608 Remove external offer editing service

bertramakers:
### Changed

- Changed `BulkLabelCommandHandler` to dispatch `AddLabel` commands instead of adding labels over HTTP. This setup with "external" events is no longer relevant. It was from the days OMD had its own UDB3 database and a bulk label action might also mean adding labels to events/places on OMD. Nowadays their events/places are stored directly inside UDB3.

### Removed

- Removed `ExternalOfferEditingServiceInterface` and implementation
- Removed unused classes that added an API key and JWT to a HTTP request based on the API key and JWT that are on the request that's being handled (this was needed to authenticate the label HTTP requests sent from `ExternalOfferEditingServiceInterface`, but is now no longer needed)

---

Note: This requires the changes from #589 to ensure new label aggregates get created if the label that gets added in bulk does not exist yet.

Tested this manually with bulk label actions after restarting the supervisor processes.

- Bulk labelling a selection still works for both events and places.
- Bulk labelling the results of a query still works for events
- Bulk labelling the results of a query does not work for places. But this also seems to be a bug on test. I asked Corneel to verify and we'll create a separate bug ticket for that than. I think the cause is that we execute the given query against `/events/` on SAPI3, not `/offers/`.


jdrieghe (approved):
YES YES YES!

PR #592, opened 2021-02-25
III-3340 Fix conclude query

LucWollants:
### Changed

- The query to get concluded events still used the SAPI2 syntax and properties instead of the SAPI3 syntax. This was fixed so the conclude command finds again events to conclude.

---

Ticket: https://jira.uitdatabank.be/browse/III-3340

### DIFF
--- a/app/Console/ConcludeCommand.php
+++ b/app/Console/ConcludeCommand.php
@@ -29,7 +29,7 @@ class ConcludeCommand extends AbstractConcludeCommand
     }
 
 
-    public function configure()
+    public function configure(): void
     {
         $this
             ->setName('event:conclude')
@@ -55,7 +55,7 @@ class ConcludeCommand extends AbstractConcludeCommand
             );
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
     {
         [$lowerDateBoundary, $upperDateBoundary] = $this->processArguments($input);
         $query = $this->createLuceneQuery($lowerDateBoundary, $upperDateBoundary);
@@ -76,20 +76,14 @@ class ConcludeCommand extends AbstractConcludeCommand
         return 0;
     }
 
-    private function createLuceneQuery(Carbon $lowerDateBoundary = null, Carbon $upperDateBoundary = null)
+    private function createLuceneQuery(Carbon $lowerDateBoundary, Carbon $upperDateBoundary): string
     {
         $sapiDateRange = $this->createLuceneDateRangeString($lowerDateBoundary, $upperDateBoundary);
 
         return "_type:event AND availableRange:{$sapiDateRange}";
     }
 
-    /**
-     * @param Carbon|null $lowerDateBoundary
-     * @param Carbon|null $upperDateBoundary
-     *
-     * @return string
-     */
-    private function createLuceneDateRangeString(Carbon $lowerDateBoundary = null, Carbon $upperDateBoundary = null)
+    private function createLuceneDateRangeString(Carbon $lowerDateBoundary, Carbon $upperDateBoundary): string
     {
         $from = $lowerDateBoundary ? $this->getLuceneFormattedDateTime($lowerDateBoundary) : '*';
         $to = $this->getLuceneFormattedDateTime($upperDateBoundary);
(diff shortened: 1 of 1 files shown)


app/Console/ConcludeCommand.php
```diff
@@ -15,7 +15,7 @@
 class ConcludeCommand extends AbstractConcludeCommand
 {
     const TIMEZONE = 'Europe/Brussels';
-    const SOLR_DATE_TIME_FORMAT = 'Y-m-d\TH:i:s\Z';
+    const LUCENE_DATE_TIME_FORMAT = 'Y-m-d\TH:i:s\Z';
```
bertramakers, line 18: I think this date format is not 100% correct for SAPI3.

The dates in SAPI3 are formatted as `yyyy-MM-dd'T'HH:mm:ssZZ`: https://github.com/cultuurnet/udb3-search-service/blob/master/src/ElasticSearch/Operations/json/mapping_event.json#L22

Elasticsearch uses this pattern syntax: https://www.joda.org/joda-time/apidocs/org/joda/time/format/DateTimeFormat.html

```
Symbol  Meaning                      Presentation  Examples
------  -------                      ------------  -------
Z       time zone offset/id          zone          -0800; -08:00; America/Los_Angeles

The count of pattern letters determine the format. 

Zone: 'Z' outputs offset without a colon, 'ZZ' outputs the offset with a colon, 'ZZZ' or more outputs the zone id. 
```

So the dates are formatted as for example: `2021-01-01T00:00:00+01:00`

In PHP, that is `Y-m-d\TH:i:sP`, not `Y-m-d\TH:i:s\Z`: https://www.php.net/manual/en/datetime.format.php

```
P | Difference to Greenwich time (GMT) with colon between hours and minutes | Example: +02:00
Z | Timezone offset in seconds. The offset for timezones west of UTC is always negative, and for those east of UTC is always positive. | -43200 through 50400
```

`Y-m-d\TH:i:sP` is the same as `DATE_ATOM`, which is also what we use as format when indexing the documents. For example: https://github.com/cultuurnet/udb3-search-service/blob/master/src/ElasticSearch/JsonDocument/Properties/AvailabilityTransformer.php#L59

bertramakers: The `\` before `\Z` is also weird, because that escapes it and would keep the `Z` instead of putting the timezone there. (Like `\T`)

app/Console/ConcludeCommand.php
```diff
@@ -76,34 +76,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
     }
 
-    private function createSolrQuery(Carbon $lowerDateBoundary = null, Carbon $upperDateBoundary = null)
+    private function createLuceneQuery(Carbon $lowerDateBoundary, Carbon $upperDateBoundary): string
     {
-        $sapiDateRange = $this->createSolrDateRangeString($lowerDateBoundary, $upperDateBoundary);
+        $sapiDateRange = $this->createLuceneDateRangeString($lowerDateBoundary, $upperDateBoundary);
 
-        return "type:event AND availableto:{$sapiDateRange}";
+        return "_type:event AND availableRange:{$sapiDateRange}";
```
bertramakers, line 83: Nit: Might be a bit easier to read this method if we inline the `createLuceneDateRangeString()` logic? Since it's only called once anyway.

```
private function createLuceneQuery(Carbon $lowerDateBoundary, Carbon $upperDateBoundary): string
{
    $from = $lowerDateBoundary ? $this->getLuceneFormattedDateTime($lowerDateBoundary) : '*';
    $to = $this->getLuceneFormattedDateTime($upperDateBoundary);

    return "_type:event AND availableRange:[{$from} TO {$to}]";
}
```

LucWollants: Good point! Even the `getLuceneFormattedDateTime` has little added value.

app/Console/ConcludeCommand.php
LucWollants: Then I think it's best to just use `$lowerDateBoundary->format(DateTime::ATOM)`

LucWollants:
Thanks, @bertramakers for the detailed review!

bertramakers (approved):
Looking great now! 👍

PR #594, opened 2021-02-26
Strict types

jdrieghe:
### Changed
- Enforce declaration of strict types


jdrieghe:
We're green. This was a lot of fun 😅

LucWollants (approved):
Nice work!

I also did a functional test and creating an event, place, and organizer still works.

PR #600, opened 2021-03-02
III-3842 Update udb3-api-guard

LucWollants:
### Changed
- Update udb3-api-guard
---
Ticket: https://jira.uitdatabank.be/browse/III-3842


.github/workflows/ci.yml
```diff
@@ -54,6 +54,7 @@ jobs:
               uses: shivammathur/setup-php@v2
               with:
                   php-version: 7.1
+                  extensions: tidy, zip
```
bertramakers, line 57: Strange that the tests need `bcmath, tidy`, but the coding standards and phpstan need `tidy, zip` 🤔 
And should these be documented in `composer.json` then?

LucWollants: Crashes because of missing PHP extensions are triggered by the code that is run:
- For tests this is the PHPUnit code that is used and the silex code that is covered (including covered dependency code)
- For static analysing this the PHPStan code
- For code style this is the PHP-CS-Fixer code
- For runtime this is the executed silex code and code from dependencies

PHPStorm has a code inspection for "Extension is missing in composer.json". I did run this on our code base and that resulted in the following set:
```
    "ext-dom": "*",
    "ext-iconv": "*",
    "ext-json": "*",
    "ext-libxml": "*",
    "ext-pcntl": "*",
    "ext-pdo": "*",
    "ext-redis": "*",
    "ext-tidy": "*",
    "ext-simplexml": "*",
    "ext-xmlreader": "*",
```

LucWollants: See https://www.jetbrains.com/help/phpstorm/php-extension-is-missing-in-composer-json.html

bertramakers (approved):

.github/workflows/ci.yml
bertramakers: Yes, I was mostly wondering if we shouldn't just add all required extensions to every Github Actions check. Since they might not need all right now, but since we depend on them in some places of the code it's possible that they will need some later. They're just our app's dependencies, and we also install all composer dependencies for every step for example, not just the ones coincidentally needed 😄

bertramakers (approved):

.github/workflows/ci.yml
LucWollants: Ok, I now understand what you mean. I did interpret it as having the smallest possible set of extensions for every step 😊
But something we can easily fix or finetune in another PR

PR #601, opened 2021-03-02
III-3608 CLI command to import labels from SQL table

bertramakers:
### Added

- Added CLI command to import labels from the `labels_import` table. The script loops over every row in the table and dispatches a `AddLabel` command for it. It does not check if the offer already has the label, because the aggregate does that + the table should get truncated before it gets new data. It also catches every exception that might happen when dispatching the command to ensure that it keeps running, but it can then later print debug info for those caught exceptions.

### Fixed

- Fixed an issue that happened if an `Error` is thrown when running `./bin/udb3.php`. It was passed to the `renderException()` method of the console app, but that can only render exceptions, not errors, resulting in an additional error. That caused the original error to be hidden because it was the second error that would become fatal and show up in the output instead.

---


LucWollants (approved):
Nice! Easy to read and very clean. 🚢

PR #602, opened 2021-03-03
III-3838 Keep setting status on offers in case of exceptions

LucWollants:
### Changed
- Catch exceptions when updating the status of events instead of bailing out. These exceptions will be dumped when the command finishes
- Stop the command when no offers were found
- Default answer for update is false.

---
Ticket: https://jira.uitdatabank.be/browse/III-3838

### DIFF
--- a/app/Console/UpdateOfferStatusCommand.php
+++ b/app/Console/UpdateOfferStatusCommand.php
@@ -67,7 +67,7 @@ class UpdateOfferStatusCommand extends AbstractCommand
                 $output,
                 new ConfirmationQuestion(
                     "This action will update the status of {$count} {$this->getPluralOfferType()} to {$statusType->toNative()}, continue? [y/N] ",
-                    true
+                    false
                 )
             );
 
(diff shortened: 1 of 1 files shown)


app/Console/UpdateOfferStatusCommand.php
```diff
@@ -61,32 +62,48 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
             $progressBar->advance();
         }
 
         $progressBar->finish();
 
+        $output->writeln('');
+        foreach ($exceptions as $exception) {
+            $output->writeln($exception);
+        }
+
```
bertramakers, line 106: Just thinking out loud here: We do the reporting of these exceptions in loops differently in the various CLI commands that do bulk operations. Maybe we should look for a more consistent approach sometime? Not in scope of this PR of course, but something to think about maybe.

bertramakers (approved):

app/Console/UpdateOfferStatusCommand.php
LucWollants: Yeah, I first wanted to make it consistent.

But I think it can be different based on the CLI command:
- Is it something that processes a lot of data?
- Is execution manual or automated?
- Do we gather only errors or all information?
- Do we want to save the gathered feedback?

So a good approach could be to create a `Feedback` class that gathers the information and/or errors and has a few output methods to stream to file our console output.

PR #608, opened 2021-03-08
Bumped PhpStan to level 4

jdrieghe:
### Changed
- We are now compatible with PHPStan Level 4

### Fixed
- A bunch of faulty type hints and unnecessary conditionals
- Warning in test suite about mocking a final class


LucWollants (approved):
Nice work

PR #618, opened 2021-03-15
III-3816 Consistent error/exception logging

bertramakers:
### Added

- Added `/bin` to php-cs-fixer config

### Changed

- Almost all HTTP exceptions now get logged to a `/log/errors.log` file (as well as Sentry), except for common user exceptions
- Sentry integration now happens via Sentry's Monolog handler, which also happens to work with an `exception` key in the log's `context`: https://github.com/getsentry/sentry-php/blob/master/src/Monolog/Handler.php#L56-L58
- Simplified logger instantiation of other log files

---
Ticket: https://jira.uitdatabank.be/browse/III-3816

If this approach looks good, I'll also implement it in `udb3-search-service` and `jwt-provider` (with some small changes since those don't use Silex). The diff will be smaller there though because there are a lot fewer existing loggers in those apps.


bertramakers:
I just realized there are still two things to do:

- [ ] Add the `tags` for Sentry in a Monolog processor that we can push on the Sentry handler, so we don't have to strip them in the processor for the StreamHandler (and can thus remove the `ContextFilteringProcessor`) + the tags will then always be set, not only for errors that go through `ErrorLogger`, but also errors that get logged by AMQP etc.
- [ ] I'll try to rework the `register_shutdown_function` in `worker_bootstrap` to use a `catch/try` instead that catches any `Throwable` and also make that go through `ErrorLogger`

app/Error/LoggerFactory.php
```diff
@@ -0,0 +1,31 @@
...
+use Sentry\Monolog\Handler as SentryHandler;
+use Sentry\State\HubInterface;
+use Silex\Application;
+
+final class LoggerFactory
+{
+    public static function create(
+        Application $app,
+        string $fileNameWithoutSuffix,
+        ?string $loggerName = null
```
LucWollants, line 18: Can we work with just one parameter?
Currently I found on instance that has both parameters
```
$logger = LoggerFactory::create($app, 'amqp', 'amqp.event_bus_forwarder');
```
I guess this is to have one file for all AMQP messages but a dedicated channel for the `event_bus_forwarder`?

bertramakers: So far this was the only one that used the extra `$loggerName` I think, but I found 15 more usages of `new Logger()` that I overlooked earlier because they weren't decorated with the `SentryPsrLoggerDecorator`. And some of those, located here, use the same StreamHandler but a different logger name. So they log to the same file but with a different logger name prefixed in the logs. For example:

```
root@udb3-web01-test:~# cat /var/www/udb-silex/log/udb2.log.1
[2021-02-09 16:34:06] udb2-events-cdbxml-enricher.DEBUG: retrieving cdbxml from http://test.open.uitdatabank.be/api/v3/event/8c76fbf0-1923-4d57-9ce5-09dcae2cdc64 [] []
[2021-02-09 16:34:15] udb2-events-cdbxml-enricher.DEBUG: sendRequest took 8978 ms. [] []
[2021-02-09 16:34:15] udb2-events-cdbxml-enricher.DEBUG: retrieved cdbxml [] []
[2021-02-09 16:34:15] udb2-events-cdbxml-enricher.DEBUG: cdbxml is valid [] []
[2021-02-09 16:34:15] udb2-events-cdbxml-enricher.DEBUG: found event in cdbxml [] []
[2021-02-09 16:34:15] udb2-events-to-udb3-event-applier.INFO: No existing offer with the same id found so it is safe to import. {"offer-id":"8c76fbf0-1923-4d57-9ce5-09dcae2cdc64"} []
[2021-02-09 16:34:16] udb2-events-to-udb3-event-applier.INFO: Offer succesfully created. {"offer-id":"8c76fbf0-1923-4d57-9ce5-09dcae2cdc64"} []
```

I'm assuming that this makes grepping etc easier, so I would keep the option to use a different logger name than the file name? But in most cases they can be the same (1 file and 1 logger) so that's why I made it optional.

LucWollants: Yeah, the grouping can be nice, so indeed keep it.

bin/udb3.php
```diff
@@ -132,9 +132,9 @@
 try {
     $consoleApp->run();
 } catch (\Exception $exception) {
-    $app[SentryErrorHandler::class]->handle($exception);
+    $app[ErrorLogger::class]->log($exception);
     $consoleApp->renderException($exception, new ConsoleOutput());
```
LucWollants, line 136: I now notice that this function is deprecated. Not related to this change but related to the Symfony upgrade, maybe best to use `renderThrowable` now?

LucWollants:
Is it worth on writing a very small part in the readme about:
- How logging works (for example the `errors.log` file, the discard of user errors, the Sentry integration)
- How to use logging

LucWollants:
While reading the comments on https://jira.uitdatabank.be/browse/III-3816 I did check the `debug` setting for `udb3-silex`. I noticed this is set to `true` in [bootstrap.php](https://github.com/cultuurnet/udb3-silex/blob/1ed0bb6ca2346485335f34b5013ba1b95720e841/bootstrap.php#L71). Does it makes sense to move this to `config.yml`?

app/Error/LoggerFactory.php
```diff
@@ -0,0 +1,79 @@
...
+use Monolog\Handler\GroupHandler;
+use Monolog\Handler\StreamHandler;
+use Monolog\Logger;
+use Monolog\Processor\PsrLogMessageProcessor;
+use Sentry\Monolog\Handler as SentryHandler;
+use Sentry\State\HubInterface;
+use Silex\Application;
+
+final class LoggerFactory
+{
```
LucWollants, line 16: This makes creating a logger much easier and intuitive! 👍

LucWollants:
I did some functional tests:
- An `InvalidArgumentException`
    - got reported as an API problem
    - got added to the `error.log`
- An `EntityNotFoundException`
    - got reported as an API problem
    - not added to the `error.log`
- An `Error`
    - did not get reported as an API problem
    - got added to the `error.log`

bin/udb3.php
bertramakers: Yes, that did not exist until we updated Symfony Console recently.

bertramakers: Thanks for mentioning it!

bertramakers:
> Is it worth on writing a very small part in the readme about:
> 
>     * How logging works (for example the `errors.log` file, the discard of user errors, the Sentry integration with empty/non-empty DNS, the difference between "intentional" logging and "exception/error" logging)
> 
>     * How to use logging

Good suggestion, I'll add it. "How to use logging" -> How do you see that exactly? Do you mean what to do if you want to add an extra logger?

bertramakers:
> While reading the comments on https://jira.uitdatabank.be/browse/III-3816 I did check the `debug` setting for `udb3-silex`. I noticed this is set to `true` in [bootstrap.php](https://github.com/cultuurnet/udb3-silex/blob/1ed0bb6ca2346485335f34b5013ba1b95720e841/bootstrap.php#L71). Does it makes sense to move this to `config.yml`?

I overlooked that, I thought it wasn't used. But with these changes `debug` doesn't really do anything because nothing gets caught/handled by Silex itself now, we all do it ourselves. In other apps it overrides how we handle errors/exceptions and shows a pretty page by Whoops, but we don't have that here (yet). In the meantime I could add a "poor man's debug" that adds some extra info to the API Problem JSON if enabled, like the original exception class and stacktrace etc?

LucWollants:
> Good suggestion, I'll add it. "How to use logging" -> How do you see that exactly? Do you mean what to do if you want to add an extra logger?

Indeed just a simple line of code referring to `LoggerFactory` which explains itself.

LucWollants:
> In the meantime I could add a "poor man's debug" that adds some extra info to the API Problem JSON if enabled, like the original exception class and stacktrace etc?

If we don't do anything else with this debug mode I would just remove it?

bertramakers:
> If we don't do anything else with this debug mode I would just remove it?

I was thinking that if we keep it and make it do something useful, we could enable it by default locally and if Geoffrey/Bram/Simon run into 500's again on their local installation it's extra easy to figure out what's going on?

LucWollants:
>I was thinking that if we keep it and make it do something useful, we could enable it by default locally and if Geoffrey/Bram/Simon run into 500's again on their local installation it's extra easy to figure out what's going on?

Indeed this could help to streamline reporting errors and helping them out.

bin/udb3.php
bertramakers: Ah but we are still on Symfony Console `v3.4.47` and that `renderThrowable` does not exist in that version yet? And `renderException` has been marked as deprecated in `4.x` if I'm not mistaken

jdrieghe (approved):
Nice work here

jdrieghe (approved):

bertramakers:
I added the documentation and debug mode. I didn't use `renderThrowable` in `bin/udb3.php` because we're not on `symfony/console 4.x` yet so it's not yet available.

LucWollants (approved):
Nice work!

PR #621, opened 2021-03-18
III-3526

jdrieghe:
### Changed
- When importing places or events with an availableFrom date in the past, we will use the current date instead

---
Ticket: https://jira.uitdatabank.be/browse/III-3526

---
Note that I changed the interface's signature to better match how it was used in reality. It now always expects a default and never returns `null` (as it already did in production).

### DIFF
--- a/tests/Model/Import/Offer/Udb3ModelToLegacyOfferAdapterTest.php
+++ b/tests/Model/Import/Offer/Udb3ModelToLegacyOfferAdapterTest.php
@@ -403,7 +403,7 @@ class Udb3ModelToLegacyOfferAdapterTest extends TestCase
     /**
      * @test
      */
-    public function it_should_return_default_available_form_if_there_is_none()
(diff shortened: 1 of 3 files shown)


tests/Model/Import/Offer/Udb3ModelToLegacyOfferAdapterTest.php
```diff
@@ -403,10 +403,11 @@ public function it_should_return_a_contact_point_if_there_is_one()
     /**
      * @test
      */
-    public function it_should_return_no_available_from_by_default()
+    public function it_should_return_default_available_form_if_there_is_none()
```
bertramakers: Nit: Typo in the test name

bertramakers (commented):
Clean and elegant fix! Thanks for also cleaning up the interface and making the default required now. 👍

bertramakers (approved):

PR #622, opened 2021-03-18
III-3526 - Comment by bert

jdrieghe:
### Fixed
- https://jira.uitdatabank.be/browse/III-3526?focusedCommentId=104620&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-104620

---
Ticket: https://jira.uitdatabank.be/browse/III-3526

### DIFF
--- a/src/Event/Commands/CreateEvent.php
+++ b/src/Event/Commands/CreateEvent.php
@@ -122,11 +122,12 @@ class CreateEvent extends AbstractCreateCommand
         return $this->location;
     }
 
-    /**
-     * @return DateTimeImmutable|null
-     */
-    public function getPublicationDate()
+    public function getPublicationDate(DateTimeImmutable $default): ?DateTimeImmutable
     {
+        if ($this->publicationDate && $this->publicationDate < $default) {
+            return $default;
+        }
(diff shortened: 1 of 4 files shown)


src/Event/Commands/CreateEvent.php
```diff
@@ -122,11 +122,12 @@ public function getLocation()
         return $this->location;
     }
 
-    /**
-     * @return DateTimeImmutable|null
-     */
-    public function getPublicationDate()
+    public function getPublicationDate(DateTimeImmutable $default): ?DateTimeImmutable
```
bertramakers, line 125: Is there still a use case to make the return type nullable? If the `$this->publicationDate` is `null`, can't we also return the `$default`? 🤔

jdrieghe: I don't think so, because I think it's possible to create events without publication date and then publish them manually later.

bertramakers: Ah yes I overlooked the concept of drafts :)

bertramakers (approved):

PR #625, opened 2021-03-18
III-3887 Convert hidePast parameter value to boolean

bertramakers:
### Fixed

- Fixed TypeError about hidePast argument

---
Ticket: https://jira.uitdatabank.be/browse/III-3887

### DIFF
--- a/src/Http/Event/ReadEventRestController.php
+++ b/src/Http/Event/ReadEventRestController.php
@@ -111,7 +111,7 @@ class ReadEventRestController
 
         $style = $request->query->get('style', 'text');
         $langCode = $request->query->get('langCode', 'nl_BE');
-        $hidePastDates = $request->query->get('hidePast', false);
+        $hidePastDates = (bool) $request->query->get('hidePast', false);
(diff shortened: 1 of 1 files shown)


jdrieghe (approved):

src/Http/Event/ReadEventRestController.php
```diff
@@ -111,7 +111,7 @@ public function getCalendarSummary(string $cdbid, Request $request): string
 
         $style = $request->query->get('style', 'text');
         $langCode = $request->query->get('langCode', 'nl_BE');
-        $hidePastDates = $request->query->get('hidePast', false);
+        $hidePastDates = (bool) $request->query->get('hidePast', false);
```
LucWollants, line 114: This solution is perfect for me. I just want to point out that the `Request` is a `ParameterBag` and contains "typed" methods like `getBoolean`

bertramakers: Thanks I'll keep that in mind! It will get refactored anyway when we switch to PSR-7 request/response objects

PR #629, opened 2021-03-19
III-3892 Limit productions results in a join that groups rows by production_id

bertramakers:
### Fixed

- Fixed productions returned when searching with a `limit`. When the `GROUP_CONCAT` was removed, the `limit` was applied to rows per event instead of rows per production

---
Ticket: https://jira.uitdatabank.be/browse/III-3892

### DIFF
--- a/src/Event/Productions/ProductionRepository.php
+++ b/src/Event/Productions/ProductionRepository.php
@@ -99,9 +99,19 @@ class ProductionRepository extends AbstractDBALRepository
      */
     public function search(string $keyword, int $start, int $limit): array
     {
-        $query = $this->createSearchQuery($keyword)
+        $subQuery = $this->createSearchQuery($keyword)
             ->setFirstResult($start)
             ->setMaxResults($limit);
+
+        $query = $this->getConnection()->createQueryBuilder()
+            ->select('p1.production_id, p1.name, p1.event_id')
+            ->from($this->getTableName()->toNative(), 'p1')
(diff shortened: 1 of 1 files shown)


LucWollants (approved):
Nice and clever solution!

My first impression was: "Why not use `WHERE IN`?". So I executed 2 queries on production:

1. The `INNER JOIN`
```
SELECT
	p1.production_id, p1.name, p1.event_id
FROM
	productions as p1
INNER JOIN
	(
		SELECT
			DISTINCT production_id
		FROM
			productions
		WHERE
			MATCH(name) AGAINST('Alex*' IN BOOLEAN MODE)
	) as p2
ON
	p1.production_id = p2.production_id
```

1. The `WHERE IN`
```
SELECT
	p1.production_id, p1.name, p1.event_id
FROM
	productions as p1
WHERE
	p1.production_id
IN
	(
		SELECT
			DISTINCT production_id
		FROM
			productions
		WHERE
			MATCH(name) AGAINST('Alex*' IN BOOLEAN MODE)
	)
```

Turns out the `INNER JOIN` is faster in this case 😊

jdrieghe (approved):

bertramakers:
I also tried the `WHERE IN` first @LucWollants , but it turns out you can't use `LIMIT` in a `WHERE IN` subquery! https://stackoverflow.com/questions/12810346/alternative-to-using-limit-keyword-in-a-subquery-in-mysql/51877655

PR #520, opened 2020-10-26
III-3583 Extend DocumentRepository with includeMetadata URL parameter

LucWollants:
### Added
- Extend DocumentRepository with includeMetadata URL parameter

---
Ticket: https://jira.uitdatabank.be/browse/III-3583

PR #633, opened 2021-03-23
III-3819

jdrieghe:
### Added
- `EventCreated` events now also project the `createdByApiConsumer` property

---

---
### Todo
- [ ] PlaceCreated
- [ ] EventCopied
- [ ] Add temporary projector to fix existing projections without having to replay everything
- [x] Add configuration mapping to appconfig and vagrant box repos
- [ ] Add polyfill that sets this property to 'unknown' for all (old) events without a known API consumer

### DIFF
--- a/app/Event/EventJSONLDServiceProvider.php
+++ b/app/Event/EventJSONLDServiceProvider.php
@@ -76,7 +76,8 @@ class EventJSONLDServiceProvider implements ServiceProviderInterface
                         new EventJsonDocumentLanguageAnalyzer()
                     ),
                     new EventTypeResolver(),
-                    $app['config']['base_price_translations']
+                    $app['config']['base_price_translations'],
+                    $app['config']['udb_api_key']
                 );
 
                 return $projector;
--- a/config.yml.dist
+++ b/config.yml.dist
@@ -117,7 +117,7 @@ jwt:
       iss: http://culudb-jwt-provider.dev
   auth0:
     keys:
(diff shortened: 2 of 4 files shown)


bertramakers (approved):
Code so far looks good to me. Can either be continued upon in this PR or this can be merged and the rest can be done in other PRs.

Something we didn't discuss yet and also didn't see in the todo's is what to do with `EventImportedFromUDB2` and `PlaceImportedFromUDB2`. Those events and places should also get a `createdByApiConsumer` property, but maybe it's enough if that is always `unknown` through the polyfill repository? On the other hand if we remove that polyfill later after a full replay, the property would disappear from those events/places if it's not set in the projectors for those events.

And also (unrelated): The `metadata` property is now always included, which feels a bit weird because we have an `includeMetadata` URL parameter that is `false` by default. https://github.com/cultuurnet/udb3-silex/blob/332f24c85d5cb61faf1b00f930735d3fe4d5c036/src/Http/Event/ReadEventRestController.php#L55
One solution I see is explicitly unsetting the `metadata` before returning the event/place if the URL parameter is false? Unless you see another solution?

Some more context about that URL parameter:
- https://jira.uitdatabank.be/browse/III-3583
- https://github.com/cultuurnet/udb3-silex/pull/520

jdrieghe:
Gonna close this one in favour of a new approach with a dedicated read model and a polyfill

PR #634, opened 2021-03-23
III-3881 Consistent logger names

bertramakers:
### Changed

- Renamed log files for more consistency and predictability

---
Ticket: https://jira.uitdatabank.be/browse/III-3881

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -513,8 +513,8 @@ $app['event_repository'] = $app->share(
     }
 );
 
-$app['logger.command_bus'] = $app::share(
-    function ($app) {
+$app['logger_factory.resque_worker'] = $app::protect(
+    function ($queueName) use ($app) {
         $redisConfig = [
             'host' => '127.0.0.1',
             'port' => 6379,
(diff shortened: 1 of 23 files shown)


bootstrap.php
```diff
@@ -513,8 +513,8 @@ function ($app) {
     }
 );
 
-$app['logger.command_bus'] = $app::share(
-    function ($app) {
+$app['logger_factory.resque_worker'] = $app::protect(
```
LucWollants, line 516: Is it not a `logger` instead of a `logger_factory` that is stored here?

LucWollants: For example here it is called a logger https://github.com/cultuurnet/udb3-silex/pull/634/files#diff-6fe69b9de46369cbfc5f0fc587f6afbe09d8a3736fa3066f274dc61fdc13cdf2R181

bertramakers: No, using `protect` means that it will always return a closure. So `$app['logger_factory.resque_worker']` is not a logger but a function returning a logger. See also the usage here for example https://github.com/cultuurnet/udb3-silex/pull/634/commits/7141a7cf7735388e13cc10a2810396c48b041ef5#diff-3fad7630b5b5be90ba86eefe7b8b4e70c821e0bb38867851059e6441bf00841bR126

bertramakers: > For example here it is called a logger https://github.com/cultuurnet/udb3-silex/pull/634/files#diff-6fe69b9de46369cbfc5f0fc587f6afbe09d8a3736fa3066f274dc61fdc13cdf2R181

That's using `$app->share()` which returns a singleton so it's a logger, not a function

LucWollants: Ok, subtile differences. Thanks for pointing out!

LucWollants (approved):

bootstrap.php
bertramakers: Thanks for reviewing and making me double check to be sure! :)

PR #635, opened 2021-03-26
III-3819 - Projector and repository for OfferMetadata

jdrieghe:
### Added
- Table, Projector and Repository for `OfferMetadata` read model

---

---
Next up: adding a polyfill to add this metadata to the JsonLD read model when necessary, similar to the popularity score

### DIFF
--- a/src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
+++ b/src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
@@ -39,69 +39,38 @@ class OfferMetadataProjector implements EventListener
 
     public function applyEventCreated(EventCreated $eventCreated, DomainMessage $domainMessage)
     {
-        try {
-            $offerMetadata = $this->repository->get($eventCreated->getEventId());
-        } catch (EntityNotFoundException $e) {
-            $offerMetadata = OfferMetadata::default($eventCreated->getEventId());
-        }
-
-        $createdByApiConsumer = $this->getCreatedByApiConsumerFromMetadata($domainMessage->getMetadata());
-        $offerMetadata = $offerMetadata->withCreatedByApiConsumer($createdByApiConsumer);
-
-        $this->repository->save($offerMetadata);
+        $this->projectMetadataForOffer($eventCreated->getEventId(), $domainMessage->getMetadata());
     }
 
     public function applyPlaceCreated(PlaceCreated $placeCreated, DomainMessage $domainMessage)
     {
-        try {
-            $offerMetadata = $this->repository->get($placeCreated->getPlaceId());
-        } catch (EntityNotFoundException $e) {
-            $offerMetadata = OfferMetadata::default($placeCreated->getPlaceId());
-        }
-
-        $createdByApiConsumer = $this->getCreatedByApiConsumerFromMetadata($domainMessage->getMetadata());
-        $offerMetadata = $offerMetadata->withCreatedByApiConsumer($createdByApiConsumer);
-
-        $this->repository->save($offerMetadata);
+        $this->projectMetadataForOffer($placeCreated->getPlaceId(), $domainMessage->getMetadata());
     }
 
     public function applyEventCopied(EventCopied $eventCopied, DomainMessage $domainMessage)
     {
-        try {
-            $offerMetadata = $this->repository->get($eventCopied->getItemId());
-        } catch (EntityNotFoundException $e) {
-            $offerMetadata = OfferMetadata::default($eventCopied->getItemId());
-        }
-
-        $createdByApiConsumer = $this->getCreatedByApiConsumerFromMetadata($domainMessage->getMetadata());
-        $offerMetadata = $offerMetadata->withCreatedByApiConsumer($createdByApiConsumer);
-
-        $this->repository->save($offerMetadata);
+        $this->projectMetadataForOffer($eventCopied->getItemId(), $domainMessage->getMetadata());
     }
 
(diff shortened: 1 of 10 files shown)


bertramakers (approved):
Very nice so far, this was a great idea @jdrieghe! Much more simple and flexible 👍

src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            $offerMetadata = $this->repository->get($eventCreated->getEventId());
+        } catch (EntityNotFoundException $e) {
+            $offerMetadata = OfferMetadata::default($eventCreated->getEventId());
+        }
+
+        $createdByApiConsumer = $this->getCreatedByApiConsumerFromMetadata($domainMessage->getMetadata());
+        $offerMetadata = $offerMetadata->withCreatedByApiConsumer($createdByApiConsumer);
+
+        $this->repository->save($offerMetadata);
```
LucWollants, line 51: Maybe here is some duplicated code to extract into a private method?

src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
```diff
@@ -0,0 +1,125 @@
...
+use CultuurNet\UDB3\Place\Events\PlaceImportedFromUDB2;
+
+class OfferMetadataProjector implements EventListener
+{
+    use DelegateEventHandlingToSpecificMethodTrait;
+
+    /**
+     * @var OfferMetadataRepository
+     */
+    private $repository;
```
LucWollants, line 25: Maybe `offerMetadataRepository` is a better name?

src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
```diff
@@ -0,0 +1,125 @@
...
+            $offerMetadata = OfferMetadata::default($placeImportedFromUDB2->getActorId());
+        }
+
+        $createdByApiConsumer = $this->getCreatedByApiConsumerFromMetadata($domainMessage->getMetadata());
+        $offerMetadata = $offerMetadata->withCreatedByApiConsumer($createdByApiConsumer);
+
+        $this->repository->save($offerMetadata);
+    }
+
+    private function getCreatedByApiConsumerFromMetadata(Metadata $metadata): string
```
LucWollants, line 110: Maybe an `CreatedByApiConsumer` value object can help here with `other` and `unknown`?

src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
jdrieghe: I added this improvement to the follow-up PR: https://github.com/cultuurnet/udb3-silex/pull/636

LucWollants:
There is some code to handle a change in how something got created. I don't think that this can happen?
Maybe that is only to handle a replay? I it's to handle a replay then there is a script that cleans the MySQL read models.

src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
jdrieghe: I added this improvement to the follow-up PR: https://github.com/cultuurnet/udb3-silex/pull/636

src/Offer/ReadModel/Metadata/OfferMetadataProjector.php
jdrieghe: Yeah, I wasn't sure if this was enough to add another VO to the codebase as it's only used in this spot.

LucWollants:
There is this class https://github.com/cultuurnet/udb3-silex/blob/6f629a3bbfa4f6caec085b8642d2a58ff05de522/app/Console/PurgeModelCommand.php

jdrieghe:
@LucWollants should our table be added there then?

LucWollants:
Yes, I would take the approach of purging it. That can also make the logic a bit easier because a row should never be updated.

src/Offer/ReadModel/Metadata/OfferMetadataRepository.php
```diff
@@ -0,0 +1,52 @@
...
+class OfferMetadataRepository extends AbstractDBALRepository
+{
+    public const TABLE = 'offer_metadata';
+
+    public function __construct(Connection $connection)
+    {
+        parent::__construct($connection, new StringLiteral(self::TABLE));
+    }
+
+    public function get(string $offerId)
```
LucWollants, line 21: This only seems to be used inside the tests?

jdrieghe: Yes, we use it in the follow-up PR to enrich the jsonld

bertramakers:
I'd actually prefer the idempotency to be honest. We also sometimes have to do partial replays after problems on production, and we cannot purge the read models then. If a projector is not idempotent it leads to issues like this and it makes those replays after issues a lot harder: https://jira.uitdatabank.be/browse/III-3884

Can we add the idempotency back in another PR?

bertramakers:
Also fun fact about that purge command, Kristof actually never used that in the past. I've pointed it out to him for future replays, but he didn't know about its existence. He cleared the read models himself, which lead to other issues (clearing the unique tables). Just as an example/warning that we cannot assume that something is done in a specific way just because a command exists for it. Idempotency seems better to avoid issues vs assuming that the read models will always be cleared before a (partial) replay is done.

PR #636, opened 2021-03-26
III-3819 - Add decoration layer that embeds the offer metadata

jdrieghe:
### Added
- When requesting places or events with metadata, the `createdByApiConsumer` property will now be embedded. If that information is not available (due to a not yet available read model or the offer being a very old event), we will default to `unknown`.

---

### DIFF
--- a/src/Offer/ReadModel/Metadata/OfferMetadataEnrichedOfferRepository.php
+++ b/src/Offer/ReadModel/Metadata/OfferMetadataEnrichedOfferRepository.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\ReadModel\Metadata;
+
+use CultuurNet\UDB3\EntityNotFoundException;
+use CultuurNet\UDB3\ReadModel\DocumentRepository;
+use CultuurNet\UDB3\ReadModel\DocumentRepositoryDecorator;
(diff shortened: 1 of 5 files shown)


bertramakers (approved):
Looking good! 👍

src/Offer/ReadModel/Metadata/OfferMetadataEnrichedOfferRepository.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        try {
+            $offerMetadata = $this->offerMetadataRepository->get($jsonDocument->getId());
+        } catch (EntityNotFoundException $e) {
+            $offerMetadata = OfferMetadata::default($jsonDocument->getId());
+        }
+
+        return $jsonDocument->applyAssoc(
+            function (array $body) use ($offerMetadata) {
+                $body['metadata']['createdByApiConsumer'] = $offerMetadata->getCreatedByApiConsumer();
```
LucWollants, line 57: Does this assume there already is something inside `$body['metadata']`?

LucWollants (approved):

src/Offer/ReadModel/Metadata/OfferMetadataEnrichedOfferRepository.php
jdrieghe: It does not: https://3v4l.org/ljBQY

PR #641, opened 2021-04-07
III-3945 Remove Conclude feature

LucWollants:
### Removed
- Removed Conclude CLI code
- Removed Conclude command
- Removed Conclude aggregate method
- Removed Conclude projectors

Note: the `Concluded` event is kept to prevent a crash when loading an event stream

---
Ticket: https://jira.uitdatabank.be/browse/III-3945


jdrieghe (commented):
Is it safe to remove the event class? What do we do with existing events in the event store? Will it be possible to load a stream without the necessary event class?

bin/udb3.php
```diff
@@ -57,7 +57,7 @@
 $consoleApp = $app['console'];
 $consoleApp->setCatchExceptions(false);
 
-// An udb3 system user is needed for conclude and geocode commands.
+// An udb3 system user is needed for geocode commands.
```
bertramakers, line 60: Nit: The commands for updating the status of a single offer or multiple offers also needs this system user.

bertramakers:
> Is it safe to remove the event class? What do we do with existing events in the event store? Will it be possible to load a stream without the necessary event class?

Good point. I'm not sure if there's an easy way to ignore those messages? That would be the cleanest way I guess, but if we just leave them out of the stream the playhead inside the aggregate will become incorrect compared to the playhead inside the event store, since the playhead in the Event aggregate will be 1 fewer per missing event that's left out of the stream when reconstructing from history.

bertramakers:
Maybe adding the `Concluded` event class back again with a docblock explaining why it's kept around would be the least confusing.

LucWollants:
> Is it safe to remove the event class? What do we do with existing events in the event store? Will it be possible to load a stream without the necessary event class?

Good remark! I think that will crash so it needs to be added again. Another option could be to add it to https://github.com/cultuurnet/udb3-silex/blob/master/src/BackwardsCompatiblePayloadSerializerFactory.php but not sure if we can return "nothing"

PR #642, opened 2021-04-08
Remove obsolete endpoints

jdrieghe:
### Removed
- Endpoints that were temporarily added for testing purposes


bertramakers:
Fingers crossed that someone or something is not using it in the meantime 🤞 
But it should not be an endpoint for 3rd parties, and if our own app is using it we'll notice soon enough (and the impact should be limited to user management)

bertramakers (approved):

PR #645, opened 2021-04-12
Make UserIdentityController PSR-7 compliant

jdrieghe:
### Added
- PSR-7 compatible `JsonLdResponse` and `ApiProblemJsonResponse` classes

### Changed
- `UserIdentityController` is now PSR-7 compliant

### Removed
- Unused `CultureFeedUserIdentityResolver`

---
Ticket: https://jira.uitdatabank.be/browse/III-3858


src/Http/Response/ApiProblemJsonResponse.php
```diff
@@ -0,0 +1,37 @@
...
+class ApiProblemJsonResponse extends JsonResponse
+{
+    public function __construct(ApiProblem $problem, $headers = [])
+    {
+        $headers += ['Content-Type' => 'application/problem+json'];
+
+        $status = 400;
+        if (null !== $problem->getStatus()) {
+            $headers += [
+                'X-Status-Code' => $problem->getStatus(),
```
bertramakers, line 24: I think `X-Status-Code` is not needed anymore. It seems to be something related to Symfony: https://symfony.com/blog/new-in-symfony-3-3-deprecated-x-status-code

bootstrap.php
```diff
@@ -947,7 +948,7 @@ function ($app) {
     function ($app) {
         return new \CultuurNet\UDB3\Role\ReadModel\Users\RoleUsersProjector(
             $app['role_users_read_repository'],
-            $app['user_identity_resolver']
+            $app[Auth0UserIdentityResolver::class]
```
bertramakers, line 951: I think we can now also remove the `user_identity_resolver` definition itself , right?

bertramakers (approved):
Looks great, thanks for the extra boy scouting! 🤩

PR #648, opened 2021-04-13
Clean up deprecated exception

jdrieghe:
### Changed
- Replaced some usages of the deprecated `DocumentRepository::get` with `DocumentRepository::fetch`

### Removed
- Deprecated DocumentGoneException

---
There are 46 more usages of `DocumentRepository::get` but I'll remove those tomorrow in a separate PR. This one is big enough already.


bertramakers:
Nice cleanup so far! It will be good to get rid of the deprecated `get()` soon. 🙌  I noticed that 5 of the 46 remaining usages are actually implementations of the `get()` method in decorators that call the `get()` on the decoratee. So those can just be thrown out once the other 41 usages are removed and the decorated `get()` methods are not necessary anymore.

bertramakers (approved):

PR #649, opened 2021-04-14
Remove unused LocalOfferReadingService

jdrieghe:
### Removed
- Unused `LocalOfferReadingService` and its interface. This was imported while internalizing services from our external packages but it's never used.


bertramakers (approved):
🧹

PR #650, opened 2021-04-14
III-3653 Handle duplicate organizer websites

bertramakers:
### Fixed

- Fixed integrity constraint violations when repopulating the `organizer_unique_websites` table and another organizer than the one currently being handled already has the same website (caused by the table being empty and the unique check thus not working, causing organizers to have duplicate websites)


LucWollants (approved):

LucWollants (approved):
👍

PR #652, opened 2021-04-14
Diagon Alley

jdrieghe:
### Removed
- There is no more difference between a document being `GONE` and `NOT_FOUND`. We cleaned up all related code for that.

---
Bonus points for the person who figures out the PR title 🧙


LucWollants (approved):
Nice work! Wegisweg

PR #653, opened 2021-04-14
Simplify service provisioning and remove unused feature toggle

jdrieghe:
### Changed
- Cleaned up some service provisioning

### Removed
- Unused `import-from-sapi` feature toggle and all the code related to it


app/UDB2IncomingEventServicesProvider.php
```diff
@@ -194,21 +185,12 @@ function (Application $app) use ($importFromSapi) {
 
                 $enricher->setLogger($app['logger.xml-imports.enricher']);
 
-                if ($importFromSapi) {
-                    $eventUrlFormat = $app['udb2_cdbxml_enricher.event_url_format'];
```
bertramakers, line 198: Nit: We can also remove these parameters here now: https://github.com/cultuurnet/udb3-silex/blob/cf8d528775aac8574170bcff24c530f6cb1a56d9/bootstrap.php#L1077-L1078

bertramakers (approved):
Always nice to see some UDB2-related code go away 😍

features.yml.dist
```diff
@@ -7,7 +7,3 @@ toggles:
     name: jobs-endpoint
     conditions: {}
     status: inactive
-  import-from-sapi:
```
LucWollants, line 10: It seems like this file https://github.com/cultuurnet/appconfig/blob/main/files/udb3/production/udb3-silex/features.yml could also use a cleanup.

jdrieghe: Yes, after the next chained PR is deployed we can remove the features.yml for every environment

app/UDB2IncomingEventServicesProvider.php
jdrieghe: Good catch. I pushed it in the next PR to avoid having to merge everything: https://github.com/cultuurnet/udb3-silex/pull/654/commits/d79caafd4dee2b78967a9fb4fe0624f074f240fc

features.yml.dist
bertramakers: Oh I missed this and commented the same on your other PR about feature toggles 😄

app/UDB2IncomingEventServicesProvider.php
bertramakers: Perfect!

PR #654, opened 2021-04-14
Remove feature toggles

jdrieghe:
### Removed
- Irrelevant feature toggles
- Dependency on `qandidate/toggle`


bertramakers:
Nice job! I'm assuming that the feature toggles still defined in the production config are in reality not used anymore anywhere? https://github.com/cultuurnet/appconfig/blob/main/files/udb3/production/udb3-silex/features.yml
So we can also remove that file (and the ones for acc/test) after this is deployed?

bertramakers (approved):

jdrieghe:
@bertramakers Correct, can be cleaned up after this is live.

PR #656, opened 2021-04-16
Introduce RetryingCommandBus for optimistic locking

jdrieghe:
### Fixed
- Introduced a retrying command bus as a form of optimistic locking for race conditions in an aggregate that was loaded through concurrent requests.

---
Ticket: https://jira.uitdatabank.be/browse/III-3919


LucWollants (approved):
Nice solution!

PR #658, opened 2021-04-19
III-3500 - Simplify purge command

jdrieghe:
### Changed
- Simplified implementation of purge command for read models. 
- Write models can no longer be purged.

---
Ticket: https://jira.uitdatabank.be/browse/III-3500


jdrieghe:
I checked, all necessary read model tables are in the purge command's configuration right now. Nothing needed to be added.

bertramakers (approved):
Nice cleanup! 👍

LucWollants:
Indeed nice cleanup. I wonder if everything can be moved to the `PurgeModelCommand`?
We only need a list of read model tables and iterate over them to do a truncate?

jdrieghe:
@LucWollants if we don't need the special case for SQLite in the DBALPurgeService then it can be simplified even further like that, indeed.

LucWollants:
I think we only run it when we want to replay and that is something we do manual on real MySQL data and not the SQLite test database.

jdrieghe:
@LucWollants https://github.com/cultuurnet/udb3-silex/pull/659

PR #660, opened 2021-04-20
III-3884

jdrieghe:
### Fixed
- Projections of LabelCreated will no longer break when running a replay

---
Ticket: https://jira.uitdatabank.be/browse/III-3884

### DIFF
--- a/src/Label/ReadModels/JSON/Projector.php
+++ b/src/Label/ReadModels/JSON/Projector.php
@@ -61,17 +61,20 @@ class Projector extends AbstractProjector
 
     public function applyCopyCreated(CopyCreated $copyCreated)
     {
-        $entity = $this->readRepository->getByUuid($copyCreated->getUuid());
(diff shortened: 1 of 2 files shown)


jdrieghe:
@bertramakers I'm not sure if this was what you had in mind for the projections?

bertramakers:
I had no specific implementation in mind but I think this fixes it for `Created` and it's clear what's going on so nice job! 👍 
I didn't know there was also a `CopyCreated`. Maybe we should also add the same check there?
Those two `apply` methods are the only ones that call `WriteRepositoryInterface::write()` it seems, so if we add the same check for `CopyCreated` we've covered every possible case where an insert happens for an existing label name/uuid.

LucWollants:
Also seems fine to me. I gonna do a quick functional test by renaming a label, because a rename will trigger a copy.

LucWollants (approved):
Works like a charm

PR #664, opened 2021-05-31
III-4010 Term label decorator

bertramakers:
### Added

- Added decorator around event/place JSON-LD repositories to update/add term labels on-the-fly from config

---
Ticket: https://jira.uitdatabank.be/browse/III-4010

Related config PRs:
- https://github.com/cultuurnet/appconfig/pull/70
- https://github.com/cultuurnet/udb3-vagrant/pull/63

Note that this config still needs to be double checked to make sure the label names are correct

### DIFF
--- a/src/Term/TermRepository.php
+++ b/src/Term/TermRepository.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Term;
+
+use CultuurNet\UDB3\Model\ValueObject\Taxonomy\Category\Category;
+use CultuurNet\UDB3\Model\ValueObject\Taxonomy\Category\CategoryID;
+use CultuurNet\UDB3\Model\ValueObject\Taxonomy\Category\CategoryLabel;
+
+final class TermRepository
+{
+    /**
+     * @var array
+     */
+    private $mapping;
+
+    public function __construct(array $mapping)
+    {
+        $this->mapping = $mapping;
+    }
+
+    public function getById(string $id): Category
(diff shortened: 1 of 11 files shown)


src/Term/TermRepository.php
```diff
@@ -0,0 +1,34 @@
...
+        $this->mapping = $mapping;
+    }
+
+    public function getById(string $id): Category
+    {
+        if (!isset($this->mapping[$id])) {
+            throw new TermNotFoundException($id);
+        }
+
+        $label = $this->mapping[$id]['name']['nl'] ?? null;
```
LucWollants, line 29: Should this not be created based on the main language of the offer?

bertramakers: Might be good at some point in the future, but right now they're also NL for events/places that have another `mainLanguage`, for example: https://io.uitdatabank.be/event/2357cb2e-b607-4d8f-9021-0dd28eafabc4

I figured it was best to keep the scope of this PR limited and only update the potentially outdated or incorrect Dutch label to the correct Dutch version.

bertramakers: I also created https://jira.uitdatabank.be/browse/III-4024 earlier today. If we pick that up we can add the multilinguality on that endpoint, and document this label as deprecated. Integrations can find all term labels on that endpoint then.

LucWollants: Ok, very clear. Thanks for explaining!

LucWollants (approved):
Very clear code which is covered by some meaningful tests.

PR #666, opened 2021-06-16
Remove dependency on username or email in current user's token

bertramakers:
### Added

- `current_user_id` in service container
- `current_user_is_god_user` in service container

### Changed

- History projector to use user id instead of username for the author of a change
- Usage of email in "Door mij ingevoerd" saved search by fetching the required user info from Auth0
- `/user` endpoint by fetching the required user info from Auth0
- Refactored usages of two `UserIdentification` interfaces that combined user id with a check if it's a god user to only use either one, instead of putting these two things in one object that gets used in a lot of places while only one property is needed
- Moved system user id out of `UserIdentityDetails`, because that class is only used as a return object of the `Auth0UserIdentityResolver` and the system user id does not belong on it

### Removed

- `Udb3Token::userName()` method
- `Udb3Token::email()` method
- Two `UserIdentification` interfaces that contained user id + check if it's a god user
- Unused `UserIdentityDetails` factory
- `current_user` and with that all last usages of `CultureFeed_User` 🎉 

---
Ticket: https://jira.uitdatabank.be/browse/III-4052

Note: This is a big PR because I thought the best way to make sure that we don't have any leftover code that depends on username or email from the `current_user` object, was to remove the `current_user` object completely and always work with a user id. In hindsight I should probably have tried to do this in smaller steps, but after I started removing the userName and email methods from Udb3Token there was no way to keep this in a working state for 100% sure in-between PRs if I didn't clean up every usage of the `current_user` object.

Some todo's for follow-up PRs:
- Move `/user` endpoint to a controller class and add tests for it
- Add tests for `Sapi3FixedSavedSearchRepository`


app/Offer/OfferControllerProvider.php
```diff
@@ -105,23 +101,18 @@ function (Application $app) {
                     return new OfferPermissionsController(
                         $permissionsToCheck,
                         $app['offer_permission_voter'],
-                        $currentUserId
+                        $app['current_user_id'] ? new StringLiteral($app['current_user_id']) : null
```
LucWollants, line 104: Should we try to avoid using a `StringLiteral` here instead of a plain `string`?

bertramakers: This controller passes the user id down to the permission voter injected above in `isAllowed` calls. So it would mean we'd have to refactor the permission voter as well to not use `StringLiteral`. Or we need to inject it as a nullable string and then wrap it in StringLiteral before passing it to the permission voter, but then we need to do another null check when wrapping it. So I did it here so I can just do it in this null check already.

LucWollants:
Very clean!
I also did some manual tests and everything is still working.

LucWollants (approved):

PR #668, opened 2021-06-23
III-4061 Remove api key requirement for access tokens

bertramakers:
### Added

- Added check that id tokens can only be used if they are provided by the JWT provider (for backward compatibility)
- Client id from access tokens in Sentry errors
- Client id from access tokens in event store metadata

### Removed

- Removed API key requirement for access tokens (that have an `azp` claim)

---
Ticket: https://jira.uitdatabank.be/browse/III-4061

Screen Shot 2021-06-23 at 16 26 10

### DIFF
--- a/src/Jwt/Udb3Token.php
+++ b/src/Jwt/Udb3Token.php
@@ -49,4 +49,19 @@ final class Udb3Token
         }
         return null;
     }
+
+    public function audienceContains(string $audience): bool
+    {
+        if (!$this->token->hasClaim('aud')) {
+            return false;
+        }
+
+        // The aud claim can be a string or an array. Convert string to array with one value for consistency.
+        $aud = $this->token->getClaim('aud');
+        if (is_string($aud)) {
+            $aud = [$aud];
+        }
+
+        return in_array($audience, $aud, true);
+    }
 }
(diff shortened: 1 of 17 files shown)


src/Jwt/Udb3Token.php
```diff
@@ -64,4 +64,16 @@ public function audienceContains(string $audience): bool
 
         return in_array($audience, $aud, true);
     }
+
+    public function canUseEntryAPI(): bool
+    {
+        $apis = $this->token->getClaim('https://publiq.be/publiq-apis', '');
```
LucWollants, line 78: Is the difference between the claim name and the metadata name intentional?
https://github.com/cultuurnet/udb3-search-service/blob/main/src/Http/Authentication/AuthenticateRequest.php#L151

LucWollants (approved):
Very clear commits.

src/Jwt/Udb3Token.php
bertramakers: Custom claims in Auth0 always have to be namespaced as an URL. We could also namespace the metadata of course, but there it's not required and maybe overkill? I have no strong opinion on it but if you think we should change it we'd have to change the metadata check in SAPI3 of course.

bertramakers: See also https://auth0.com/docs/tokens/create-namespaced-custom-claims

LucWollants: No it's perfect to keep it as is. Just wanted to make sure there is no correlation between those "key" values.

PR #670, opened 2021-06-25
III-4076 Fix old jwt provider tokens

bertramakers:
### Changed

- Renamed `uitid` and `auth0` to `v1` and `v2` (corresponding the old and new versions of both uitid v1 and v2 and jwt provider v1 and v2)
- Restructured config for jwt validation and auth0 api credentials (unrelated to jwt validation)
- Injected `Parser` instead of `JwtDecoderServiceInterface` into `JwtListener`
- Removed (now) unused `JwtDecoderServiceInterface::parse()` method
- Renamed `JwtDecoderServiceInterface` to `JwtValidator` because it now only validates
- Combined claim validation methods inside `JwtValidator` from three into one
- Injected two `JwtValidator` instances directly into `JwtAuthenticationProvider`, to easier differentiate between v1 and v2 tokens
- Moved extra validation of v2 tokens to decorator of `JwtValidator` so the `JwtAuthenticationProvider` does not need to know how to validate v2 tokens specifically

### Removed

- `FallbackJwtDecoderService`
- `JwtEncoderService` and `JwtEncoderServiceInterface`

### Fixed

- Fixed validation of old (v1) jwt provider tokens by not checking for `azp` or `aud` if the signature is signed by the v1 key (via the refactorings above)

---
Ticket: https://jira.uitdatabank.be/browse/III-4076


bertramakers:
This also requires some config changes in udb3-vagrant and appconfig, which I will make PRs for on Monday

LucWollants:
> This also requires some config changes in udb3-vagrant and appconfig, which I will make PRs for on Monday

Maybe we need to make those changes non-breaking? Because the current `config.yml.dist` introduces a breaking change.

LucWollants (approved):
Very clean and easy to follow.

I did a manual test with an access token and an id token. Both cases allowed me to access my offers and modify them or create new offers.

bertramakers:
> > This also requires some config changes in udb3-vagrant and appconfig, which I will make PRs for on Monday
> 
> Maybe we need to make those changes non-breaking? Because the current `config.yml.dist` introduces a breaking change.

The `.dist` file is not used anywhere, it's just an example. I'm not even sure if it's worth keeping around.

The appconfig changes are here, backward-compatible: https://github.com/cultuurnet/appconfig/pull/76

And the udb3-vagrant config here, not backward-compatible since it doesn't update automatically and when someone runs `make up` they will also get the changes from this PR so no backward-compatibility needed: https://github.com/cultuurnet/udb3-vagrant/pull/69

PR #671, opened 2021-06-28
III-4066 Extend log with Auth0 client id

LucWollants:
### Added
- Extend log with Auth0 client id
---
Ticket: https://jira.uitdatabank.be/browse/III-4066

### DIFF
--- a/src/History/Log.php
+++ b/src/History/Log.php
@@ -39,6 +39,11 @@ class Log implements JsonSerializable
     /**
      * @var string
      */
+    private $auth0ClientId;
+
+    /**
+     * @var string
+     */
     private $api;
 
     /**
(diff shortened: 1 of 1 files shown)


bertramakers:
Might be good to add a test for this. Preferably maybe one in `HistoryProjectorTest` of the Event and/or Place namespace to see that it also gets projected correctly, since there is no `LogTest` right now.

LucWollants:
> Might be good to add a test for this. Preferably maybe one in HistoryProjectorTest of the Event and/or Place namespace to see that it also gets projected correctly, since there is no LogTest right now.

I did not find a dedicated test for the history projector, or is it embedded in another test?
But it's possible to create one of course.

LucWollants:
@bertramakers I added a data provider for a different metadata structure. I only added this for the `Event` test because for `Place` it's less configurable and it does reuse the same method from `Log`

bertramakers (approved):

PR #672, opened 2021-06-28
III-4067 Support auth_api_client_id metadata for consumer

LucWollants:
### Added
- Support auth_api_client_id metadata for consumer
---
Ticket: https://jira.uitdatabank.be/browse/III-4067

Note: the mapping config still needs to be updated in the appconfig repo and the udb3-vagrant repo and needs to represent the new id for the user interface


bertramakers (approved):
Looks good!

PR #673, opened 2021-06-29
III-4065 support client access token

LucWollants:
### Added
- Add support for access tokens
---
Ticket: https://jira.uitdatabank.be/browse/III-4065

### DIFF
--- a/src/Jwt/JwtV2Validator.php
+++ b/src/Jwt/JwtV2Validator.php
@@ -35,7 +35,7 @@ final class JwtV2Validator implements JwtValidator
         $this->baseValidator->validateClaims($token);
 
         $udb3Token = new Udb3Token($token);
-        if ($udb3Token->isAccessToken()) {
+        if ($udb3Token->isAccessToken() || $udb3Token->isClientToken()) {
             $this->validateAccessToken($udb3Token);
         } else {
             $this->validateIdToken($udb3Token);
--- a/src/Jwt/Udb3Token.php
+++ b/src/Jwt/Udb3Token.php
@@ -34,6 +34,9 @@ final class Udb3Token
             return $this->token->getClaim('https://publiq.be/uitidv1id');
         }
 
+        if ($this->endsWith($this->token->getClaim('sub'), self::CLIENTS)) {
+            return 'client|' . rtrim($this->token->getClaim('sub'), self::CLIENTS);
+        }
         return $this->token->getClaim('sub');
     }
 
--- a/src/Jwt/Udb3Token.php
+++ b/src/Jwt/Udb3Token.php
@@ -12,6 +12,8 @@ use Lcobucci\JWT\Token;
  */
 final class Udb3Token
 {
+    private const CLIENTS = '@clients';
+
     /**
      * @var Token
      */
@@ -46,6 +48,18 @@ final class Udb3Token
         // if it has no specific aud. However we require our integrators to always include the "https://api.publiq.be"
         // aud, so access tokens should always have an azp in our case.
         return !is_null($this->getClientId());
+
+    public function isClientToken(): bool
+    {
+        if ($this->getClientId() === null) {
+            return false;
+        }
+
+        if (!$this->token->hasClaim('sub')) {
+            return false;
+        }
+
+        return $this->endsWith($this->token->getClaim('sub'), self::CLIENTS);
     }
 
     public function getClientId(): ?string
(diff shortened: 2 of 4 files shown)


src/Jwt/Udb3Token.php
```diff
@@ -45,7 +50,28 @@ public function isAccessToken(): bool
         // This does not 100% guarantee that the token is an access token, because an access token does not have an azp
         // if it has no specific aud. However we require our integrators to always include the "https://api.publiq.be"
         // aud, so access tokens should always have an azp in our case.
-        return !is_null($this->getClientId());
+        if ($this->getClientId() === null) {
+            return false;
+        }
+
+        if (!$this->token->hasClaim('sub')) {
+            return true;
```
LucWollants, line 58: @bertramakers I wonder if we should not beter return false here? Because a `sub` is a requirement for both an access and client token?

bertramakers: I'm not sure what this change is meant for. A client access token is technically an access token, just like a user access token. Are you trying to differentiate between user and client access tokens?

I would not change this specific method then, because we need to know it's a an access token (regardless of if it's a client or user access token) to check the EntryAPI permissions here https://github.com/cultuurnet/udb3-silex/blob/b34a30e4a7faa1dfef32ff93f90924dc561905cd/src/Jwt/JwtV2Validator.php#L38

bertramakers: The different kind of Auth0 tokens are as follows:

- Access tokens
  - Client access tokens
  - User access tokens
- Id tokens

src/Jwt/Udb3Token.php
```diff
@@ -34,6 +34,9 @@ public function id(): string
             return $this->token->getClaim('https://publiq.be/uitidv1id');
         }
 
+        if ($this->endsWith($this->token->getClaim('sub'), self::CLIENTS)) {
+            return 'client|' . rtrim($this->token->getClaim('sub'), self::CLIENTS);
+        }
```
bertramakers, line 39: Like I said in Slack I would keep the suffix and not prefix it, to avoid confusion in the future. The goal of the prefix was just to have a way to quickly see if a "user" id was a client or not, and we can do that with the `@clients` suffix from Auth0 which we just didn't see at first. If we change it to a custom prefix it will probably get confusing in the long run because it's then different from what is used in Auth0 itself.

src/Jwt/JwtV2Validator.php
```diff
@@ -35,7 +35,7 @@ public function validateClaims(Token $token): void
         $this->baseValidator->validateClaims($token);
 
         $udb3Token = new Udb3Token($token);
-        if ($udb3Token->isAccessToken()) {
+        if ($udb3Token->isAccessToken() || $udb3Token->isClientToken()) {
```
bertramakers, line 38: This should just stay "isAccessToken()`, there's no need to differentiate between user access tokens and client access tokens here I think.

And if there is, we should rename `isAccessToken()` to `isUserAccessToken()`.

tests/Jwt/JwtV2ValidatorTest.php
```diff
@@ -98,21 +98,32 @@ public function it_verifies_the_basic_claims_via_the_decoratee(): void
      */
     public function it_verifies_the_permission_to_use_entry_api_if_azp_claim_is_present(): void
     {
-        $tokenWithPermission = $this->createTokenWithClaims(
+        $accessTokenWithPermission = $this->createTokenWithClaims(
```
bertramakers, line 101: Should be `$userAccessTokenWithPermissions` to avoid confusion since it's not clear if this is an instance of a user access token or a client access token like this

tests/Jwt/JwtV2ValidatorTest.php
```diff
@@ -98,21 +98,32 @@ public function it_verifies_the_basic_claims_via_the_decoratee(): void
...
             [
                 'azp' => 'foobar',
                 'https://publiq.be/publiq-apis' => 'ups entry',
             ]
         );
 
+        $clientTokenWithPermission = $this->createTokenWithClaims(
```
bertramakers, line 108: Should probably better be `$clientAccessTokenWithPermission` for consistency

bertramakers:
I've tested the `main` branch locally with a client access token and it actually just works already. So I don't think we need an extra PR to support these. I've tested creating an event, requesting saved searches, and the user info endpoint. That last one returned a `404` as expected (since there's no user), but also a blank body instead of an error message. I'll create a nice-to-have ticket for that.

PR #674, opened 2021-06-29
III-4085 Remove Udb3Token

bertramakers:
### Changed

- Renamed `JwtUserToken` to `JsonWebToken` because it is now also used for client tokens in the Symfony Firewall, and `JwtToken` would be redundant
- Moved methods from `Udb3Token` to `JsonWebToken` and `JwtV2Validator`
- Made `$app['jwt']` return `JsonWebToken` instead of `Udb3Token`
- `JsonWebToken` does not use `Udb3Token` internally anymore
- Made `Token` from `lcobucci/jwt` an implementation detail, it's only used inside `JsonWebToken` now
- Improved error message for id tokens

### Removed

- Unused `Udb3Token` after above refactorings

---
Ticket: https://jira.uitdatabank.be/browse/III-4085


src/Jwt/Symfony/Authentication/JsonWebToken.php
```diff
@@ -0,0 +1,141 @@
...
+        $key = new Key($publicKey, $keyPassphrase);
+        return $this->token->verify($signer, $key);
+    }
+
+    public function getCredentials(): string
+    {
+        return $this->jwt;
+    }
+
+    public function toSentryTags(): array
```
LucWollants, line 132: I wonder if this is required here? It also seems to be used only once inside `SentryHandlerScopeDecorator`

bertramakers: Yeah I get what you mean. I wasn't sure, but I was trying to not "leak" the `Token` class used internally into the rest of the application so the app only knows about `JsonWebToken` and doesn't use the `Token` class directly somewhere. But then I either need to add a generic `getClaim()` method on `JsonWebToken`, which I also wanted to avoid so we have a very specific interface and more control over what claims are being read. So the only other solution I thought of was adding methods like `getUidClaim()`, `getUiTIDv1IdClaim()`, and `getSubClaim()`, but then that might get confusing with `getUserId()` which should be used normally _except_ in the Sentry tags?

LucWollants: Good point! Then it's indeed much better to keep `toSentryTags` method inside `JsonWebToken`

bertramakers: Ah I think I've found another potential solution, I'll push it so you can see what you think

LucWollants (approved):
Nice cleanup! The "centralised" and "specific" code makes it easier to understand.

src/Jwt/Symfony/Authentication/JsonWebToken.php
```diff
@@ -14,6 +14,11 @@
 
 class JsonWebToken extends AbstractToken
 {
+    public const TYPE_V1_JWT_PROVIDER_TOKEN = 'v1_jwt_provider_token';
```
LucWollants, line 17: We need a `TYPE_` prefix? It also sounds good without the prefix 😊

LucWollants (approved):
Small tweak of adding the type method but so much added value. Nicely done!

src/Jwt/Symfony/Authentication/JsonWebToken.php
bertramakers: No not really, I'll remove it because it might make the places where it's used outside of this class a bit easier to read.

bertramakers: Done in d23e862

PR #675, opened 2021-06-30
III-4062 Move current user endpoint to UserController

LucWollants:
### Changed
- Move current user endpoint from `index.php` to `UserControllerProvider`

---
Ticket: https://jira.uitdatabank.be/browse/III-4062

### DIFF
--- a/tests/Http/User/UserIdentityControllerTest.php
+++ b/tests/Http/User/UserIdentityControllerTest.php
@@ -57,7 +57,7 @@ class UserIdentityControllerTest extends TestCase
             (new ServerRequest())->withAttribute('emailAddress', 'jane.doe@anonymous.com')
         );
 
-        $this->assertJsonResponse(new JsonLdResponse($userIdentity), $response);
+        $this->assertJsonResponse(new JsonLdResponse($this->userIdentityToArray($userIdentity)), $response);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/Http/User/UserIdentityControllerTest.php
```diff
@@ -0,0 +1,145 @@
...
+        );
+
+        $this->userIdentityResolver->expects($this->once())
+            ->method('getUserById')
+            ->with(new StringLiteral('current_user_id'))
+            ->willReturn($userIdentity);
+
+        $response = $this->userIdentityController->getCurrentUser();
+
+        $userIdentityAsArray = $userIdentity->jsonSerialize();
```
bertramakers, line 108: Nit: There's also no test for `UserIdentityDetails`, so this can return whatever. Maybe hardcode the expected JSON (as an array) instead, so we don't get a self-fulfilling prophecy here.

(For example I could make `jsonSerialize()` return an empty array and nowhere would that cause a test to fail)

We could also add tests specifically for `UserIdentityDetails` but I think it's just fine to hardcode the expected JSON here, it's also easier to read at a glance.

bertramakers (approved):

tests/Http/User/UserIdentityControllerTest.php
LucWollants: Good catch! I added a method to convert to an array and also used that inside the `it_can_get_user_identity_by_email` test.

bertramakers (approved):

PR #676, opened 2021-06-30
III-4078 Fix partial match for unique website

LucWollants:
### Changed
- Changed unique website check from partial match to full match
---
Ticket: https://jira.uitdatabank.be/browse/III-4078

### DIFF
--- a/app/Console/UpdateUniqueOrganizers.php
+++ b/app/Console/UpdateUniqueOrganizers.php
@@ -64,7 +64,7 @@ class UpdateUniqueOrganizers extends Command
                 $organizerUuid = $organizerEvent['uuid'];
                 $organizerUrl = $this->getOrganizerWebsite($organizerEvent);
 
-                $updated = $this->updateOrganizer($organizerUuid, $organizerUrl);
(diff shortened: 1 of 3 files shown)


bertramakers (commented):
If we clean up the values in the table with the script and they all become normalized, do we still need the preflight request then? Won't we see that it's unique or not when we try to write the new URL into the table and the unique constraint check fails?

If I remember correctly the preflight check was because the URLs in the table were not all in the same normalized format. Some had `http://` in front, others `https://`, others nothing, ... etc.

LucWollants:
> If we clean up the values in the table with the script and they all become normalized, do we still need the preflight request then? Won't we see that it's unique or not when we try to write the new URL into the table and the unique constraint check fails?

True! That preflight code can be removed. But I think we can only do that after the script has run.

bertramakers:
> > If we clean up the values in the table with the script and they all become normalized, do we still need the preflight request then? Won't we see that it's unique or not when we try to write the new URL into the table and the unique constraint check fails?
> 
> True! That preflight code can be removed. But I think we can only do that after the script has run.

Ah yes that's correct, so we should leave it in for now 👍

bertramakers (approved):

PR #677, opened 2021-07-05
III-4084 Improve error handling on user endpoints

bertramakers:
### Added

- Added check that the current token is not a client access token
- Added factory for new api problems with types
- Added factory methods for api problems for some existing problems that already had the correct `type`

### Changed

- Improved error response when the current user is not found
- Improved error response when a user is not found by email address

---
Ticket: https://jira.uitdatabank.be/browse/III-4084

Some notes: 

- We should at some point improve the errors in general to include actual types instead of `about:blank`, and make sure the `title` is always the same for the given `type`. Hopefully this `ApiProblems` factory can be a step in right direction
- We will set up redirects for the error types to actual documentation soon (OPS team will do this) An example of documentation can be found here: https://publiq.stoplight.io/docs/authentication/docs/errors.md
- I will create a PR to also document the new `token-not-supported` type in the auth documentation
- The UDB error types will need to be documented once we spent time porting the documentation of UDB to Stoplight


src/Http/ApiProblem/ApiProblems.php
```diff
@@ -0,0 +1,77 @@
...
+            ->setStatus(400);
+    }
+
+    public static function userNotFound(string $detail): ApiProblem
+    {
+        return (new ApiProblem())
+            ->setType('https://api.publiq.be/probs/uitdatabank/user-not-found')
+            ->setTitle('User not found')
+            ->setDetail($detail)
+            ->setStatus(400);
```
LucWollants, line 75: Should this have a 404 status code?

bertramakers: See the later commits. It's a bit of a special case because the endpoint exists so it would be weird it if it would return a `404` imho. But I handled it differently by returning a `token not supported` if the user cannot be found. In any case it should not occur in reality and if it does it's because the user is deleted or the token is incorrect.

src/Http/User/UserIdentityController.php
```diff
@@ -23,41 +24,53 @@ class UserIdentityController
...
     {
+        $notFoundResponse = new ApiProblemJsonResponse(
+            ApiProblems::userNotFound('No user found for the given email address.')
+        );
+
         try {
             $emailAddress = new EmailAddress($request->getAttribute('emailAddress'));
         } catch (InvalidNativeArgumentException $e) {
-            return $this->createUserNotFoundResponse();
+            return $notFoundResponse;
```
LucWollants, line 48: Is this not a bad request instead of a not found request?

bertramakers: I'm not sure what you mean? It uses status code `404` like it did before. And the path is `users/emails/{emailAddress}`, for example `users/emails/luc@madewithlove.be` if the email address is not found that endpoint does not exist so a `404` no?

LucWollants: Indeed the actual status code is not changed.
But with a fresh look on the errors I wonder what it means if the email address is not present inside the request or has a wrong format. Is that not a bad request?

bertramakers: Oh yes I see now what you mean and you are 100% correct, I'll change it!

bertramakers: I thought this was about the user not being found when searching by email

bertramakers: I also forgot to change the "user not found" back to a `404` as it should be for the search by email address (if the email is valid), so I also did that

LucWollants (approved):

PR #680, opened 2021-07-06
III-3494 Cleanup the type hinting of commands

LucWollants:
### Changed
- Cleanup the type hinting of commands

Note: not essential for III-3494 but always better to create new code with return types

---
Ticket: https://jira.uitdatabank.be/browse/III-3494


bertramakers (approved):
Thanks for the cleanup and improvements! 👏

PR #681, opened 2021-07-07
III-4078 Fix trailing slash stripping

bertramakers:
### Changed

- URL is not normalized using regex anymore but using `parse_url()` parts (via the existing `Url` value object)

### Fixed

- Fixed trailing slash not being stripped if a query or fragment was present

---
Ticket: https://jira.uitdatabank.be/browse/III-4078


LucWollants:
I think this method also needs to be updated https://github.com/cultuurnet/udb3-silex/blob/master/app/Console/UpdateUniqueOrganizers.php#L200.
So I will create a class to make a unique URL and that class can be reused in both places.

jdrieghe (approved):

PR #683, opened 2021-07-08
III-3494 Add new endpoint to rename production

LucWollants:
### Added
- Add new endpoint to rename production

Note: an urldecode is not needed that is already done by Silex

---
Ticket: https://jira.uitdatabank.be/browse/III-3494


bertramakers (commented):
Looks good, but I'm going to change it so the name is not a URL parameter but inside a JSON body like we do for other rename endpoints, for example: https://documentatie.uitdatabank.be/content/json-ld/latest/events/event-name.html

bertramakers:
@jdrieghe can you review my extra commits? I've already reviewed @LucWollants 's commits, but of course another pair of eyes can always be useful to spot something that I overlooked if you have the time.

jdrieghe (approved):
Looking good (only checked the last couple of commits after your comment though)

PR #685, opened 2021-07-09
Add bookingAvailability through polyfill logic

LucWollants:
### Added
- Added BookingAvailability value object
### Changed
- Polyfill the top level booking availability if missing
- Polyfill the sub event booking availability if missing
- Polyfill the embedded location booking availability if missing

Note: this will not fix/change an already set value of `bookingAvailability`, validation should happen inside the Event or Place aggregate and always result in a consistent `BookingAvailabilityUpdated` event with the correct type.

---
Ticket: https://jira.uitdatabank.be/browse/III-4072

### DIFF
--- a/src/Event/ValueObjects/BookingAvailability.php
+++ b/src/Event/ValueObjects/BookingAvailability.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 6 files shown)


bertramakers (approved):
Looks great! 👌

PR #688, opened 2021-07-15
III-4133 Fix /user and /saved-searches for v1 tokens

bertramakers:
### Fixed

- Fixed 400 response on `/user` for v1 tokens
- Fixed email not being included in "door mij ingevoerd" saved search on `/saved-searches` for v1 tokens

---
Ticket: https://jira.uitdatabank.be/browse/III-4133

### DIFF
--- a/src/Jwt/Symfony/Authentication/JsonWebToken.php
+++ b/src/Jwt/Symfony/Authentication/JsonWebToken.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace CultuurNet\UDB3\Jwt\Symfony\Authentication;
 
 use CultuurNet\UDB3\User\UserIdentityDetails;
+use CultuurNet\UDB3\User\UserIdentityResolver;
+use Exception;
 use InvalidArgumentException;
 use Lcobucci\JWT\Parser;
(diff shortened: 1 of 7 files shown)


src/Jwt/Symfony/Authentication/JsonWebToken.php
```diff
@@ -87,6 +90,34 @@ public function getUserId(): string
         return $this->token->getClaim('sub');
     }
 
+    public function containsUserIdentityDetails(): bool
+    {
+        // Tokens from the V1 JWT provider (= custom) and V2 JWT provider (= Auth0 ID tokens) contain the username and
+        // email. Auth0 access tokens do not.
+        return $this->hasClaims(['nick', 'email']) || $this->hasClaims(['nickname', 'email']);
```
jdrieghe: You could replace this with `return !is_null($this->getUserIdentityDetails());`.
That way we are sure the logic in both methods never drifts apart.

jdrieghe (approved):

src/Jwt/Symfony/Authentication/JsonWebToken.php
bertramakers: Good point, but this method was not needed after all

PR #689, opened 2021-07-15
Fix token serialization bug

bertramakers:
### Fixed

- Fixed tokens not being serialized correctly when commands are queued

### Removed

- Removed null check on property that should never be null

---

See https://github.com/cultuurnet/udb3-silex/pull/686


bertramakers:
Fixed in #693

PR #691, opened 2021-07-26
Hotfixes

bertramakers:
### Fixed

- #688 
- #690 

---

Just a PR to merge the recent hotfixes into the main branch.

I've temporarily disabled `Automatically delete head branches` so this `hotfixes` branch doesn't get deleted, since we still need to deploy it to deploy the second fix. And I would keep it around until Corneel is back so we can deploy any other hotfixes if needed.


jdrieghe (approved):

bertramakers:
Enabled `Automatically delete head branches` again now.

PR #692, opened 2021-07-27
III-4141 Fix various broken permissions

bertramakers:
### Fixed

- Fixed most permissions not working for users that were not god users, except for the `AANBOD_` ones and `ORGANISATIES_BEWERKEN`. The core issue was that there was no security class or permission voter that allowed commands with the affected permissions except for the god user voter. But this was very hidden due to the complexity of the permission system and so many abstractions. I've added a default voter for permissions that do not have a specific one, that just checks if the user has the required permission and nothing else.

### Changed

- Moved security classes in the `Offer` namespace to the `Security` namespace, since all were used for more than just offers (e.g. organizers, labels, roles, ...)
- Renamed some security classes for more clarity
- When Symfony Security detects that a user should not be able to access a specific route, the error response is no longer a `500` but a `403` now.
- When a command is not allowed, the error response is no longer a `400` but a `403` now.

### Removed

- Removed multiple layers of abstractions to better make sense of the security & permission system
- Removed already unused methods/classes related to the security & permission system

---
Ticket: https://jira.uitdatabank.be/browse/III-4141

The acceptance tests, which cover a part of this functionality, were still green after the refactorings:

> 67 scenarios (67 passed)
> 726 steps (726 passed)

However they currently only test editing/deleting offers and organizers (as an owner, as a user with permission, and a user without permission). But at least that should still definitely work.

I've also tested the following permissions manually with a user that has them via a role instead of being a god user:
- Organisaties bewerken
- Organisaties beheren
- Gebruikers beheren
- Labels beheren
- Voorzieningen bewerken
- Producties aanmaken

(Both the scenario where the user has the permission and when they don't)

There's also a "Films aanmaken" permission but that has no backend check, it's only used in the new "Filminvoer" frontend project. (WIP)


LucWollants (approved):
Very nice cleanup, makes the complex code more readable and fixes the bug!

PR #693, opened 2021-08-02
Fix serialising a JWT inside the metadata context for queued command

LucWollants:
### Fixed
- Fix serialising a JWT inside the metadata context for queued command

### Removed
- Remove validation for possible null value of token which should no longer happen

Note: I noticed to late there was already an open PR https://github.com/cultuurnet/udb3-silex/pull/689 but that PR had no reference to the ticket so I missed it.

---
Ticket: https://jira.uitdatabank.be/browse/III-4134

### DIFF
--- a/app/CommandHandling/ContextFactory.php
+++ b/app/CommandHandling/ContextFactory.php
@@ -29,7 +29,7 @@ final class ContextFactory
         }
 
         if ($jwt) {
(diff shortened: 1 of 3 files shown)


bertramakers (approved):
I hadn't gotten around to testing my draft PR yet, which is why it was still in draft. As I was unsure if the string conversion fixed it. Great that we both came to the same solution 👍 😃

PR #695, opened 2021-08-03
III-4030 Update Calendar and Timestamp to support booking availability

LucWollants:
**Note:** This PR focusses on the preliminary work on `Calendar` and `Timestamp` to support booking availability. In a following PR the command, command handler and aggregate will be updated.

### Added
- Added guard to only allow updating booking availability on `single` and `multiple` calendar types
- Added `with` method to create `Calendar` with booking availability on top level
- Added `with` method to create `Calendar` with booking availability on sub events
- Added `with` method to create `Timestamp` with booking availability

### Changed
- Serialise and deserialise the booking availability from `Calendar`
- Project the booking availability from `Calendar`

---
Ticket: https://jira.uitdatabank.be/browse/III-4030


src/Calendar.php
```diff
@@ -444,13 +444,13 @@ private function determineCorrectTopStatusForProjection(): Status
...
      * This method can determine the top level booking availability from the sub events aka timestamps
-     * - If there are no timestamps this is a periodic or permanent calendar which is always available
+     * - For a periodic or permanent calendar this is always available
      * - If one of the timestamps is available then the top level is available
      * - If all of the timestamps are unavailable the top level is also unavailable
      */
     private function deriveBookingAvailabilityFromSubEvents(): BookingAvailability
     {
-        if (empty($this->timestamps)) {
+        if (!$this->allowsUpdatingBookingAvailability()) {
```
bertramakers, line 474: I slightly preferred the `if (empty($this->timestamps))` here, because it decouples the write and read side a bit better. The `allowsUpdatingBookingAvailability` is used when updating, and  `deriveBookingAvailabilityFromSubEvents` to determine the current state. It's a problem that we have in other places as well, that we accidentally validate data that is already persisted anyway which can lead to more bugs because the historically persisted data may not always comply to the latest domain rules.

src/Calendar.php
```diff
@@ -462,4 +464,18 @@ private function deriveBookingAvailabilityFromSubEvents(): BookingAvailability
...
     }
+
+    /**
+     * A projection can require a potential fix:
+     * - For a periodic or permanent calendar this is always available
+     * - If there are timestamps the top level status is calculated
+     */
+    private function determineCorrectTopBookingAvailabilityForProjection(): BookingAvailability
+    {
+        if (!$this->allowsUpdatingBookingAvailability()) {
```
bertramakers, line 494: Again, here I would just check if the event has subEvents or not. You can only calculate the top availability if there are subEvents anyway, regardless of which calendar types can be updated. If there are no subEvents, there are no booking availabilities to use to calculate it anyway.

src/Calendar.php
```diff
@@ -14,9 +14,11 @@
 use CultuurNet\UDB3\Model\ValueObject\Calendar\CalendarWithSubEvents;
 use CultuurNet\UDB3\Model\ValueObject\Calendar\OpeningHours\OpeningHour as Udb3ModelOpeningHour;
 use CultuurNet\UDB3\Model\ValueObject\Calendar\SubEvent;
+use CultuurNet\UDB3\Offer\ValueObjects\BookingAvailability;
 use DateTime;
 use DateTimeInterface;
 use DateTimeZone;
+use DomainException;
```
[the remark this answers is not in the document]
LucWollants: Yes, fully agree. I added the custom exception `UpdateBookingAvailabilityNotAllowed` now.

src/Calendar.php
LucWollants: Good point! I updated this.

src/Calendar.php
```diff
@@ -441,4 +441,25 @@ private function determineCorrectTopStatusForProjection(): Status
         // for the new status type.)
         return new Status($expectedStatusType, []);
     }
+
+    /**
+     * This method can determine the top level booking availability from the sub events aka timestamps
+     * - If there are no timestamps this is a periodic or permanent calendar which is always available
+     * - If one of the timestamps is available then the top level is available
+     * - If all of the timestamps are unavailable the top level is also unavailable
```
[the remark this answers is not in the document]
LucWollants: Thanks 😀

bertramakers (approved):
Thanks for the quick fixes! Looks perfect now! 👏

PR #697, opened 2021-08-06
III-4153 Improve body validation error responses

bertramakers:
### Added

- Added new concept of `RequestBodyParser` that should replace the old way of data deserialization/validation, **using PSR7 requests**
- `ApiProblemException`, currently only thrown by `RequestBodyParser` but also usable in other places to immediately throw an `ApiProblem` instead of having to throw a custom exception and then converting it to an ApiProblem in the error handler
- Added `slim/psr7` package to instantiate PSR7 request/response objects
- Added PSR7 `JsonResponse` based on `slim/psr7`

### Changed

- Changed error response for data validation errors in the `PUT /events/{eventId}/bookingAvailability` endpoint, using the new `RequestBodyParser` concept that does most of the heavy lifting 
- The few custom PSR7 response classes that we already had to extend from `JsonResponse` based on `slim/psr7` instead of `Zend\Diactoros\Response\JsonResponse`

### Removed

- Support for PHP 7.1, 7.2, and 7.3 (the new `slim/psr7` package requires PHP 7.4 and all servers should be on 7.4 now)

### Fixed

- Tests that were in an incorrect directory

---
Ticket: https://jira.uitdatabank.be/browse/III-4153

Note: Although https://jira.uitdatabank.be/browse/III-3835 has not been completely picked up so far, the service provided necessary to work with PSR7 requests and responses was already added apparently. See https://github.com/cultuurnet/udb3-silex/blob/master/bootstrap.php#L1121

So I've made the new `RequestBodyParser` work with PSR7 requests so we don't need to refactor it again in the future. And the UpdateBookingAvailability request handler as well so we don't need to inject the Diactoros factory there to convert HTTP Foundation requests into PSR7 requests for the `RequestBodyParser`.

I suggest that from now on we always use PSR7 requests/responses in new controllers, and clean up existing controllers as we go. So when adding new controllers, make sure to not use requests/response classes that have `HttpFoundation` in their namespace.

Todo:

- [x] Add `@deprecated` tags to old deserializers/validators/exceptions to point to this new concept of validation so it's clear what the preferred approach is moving forward


src/Http/ApiProblem/ApiProblems.php
```diff
@@ -66,6 +66,36 @@ public static function tokenNotSupported(string $detail): ApiProblem
...
+
+    public static function bodyInvalidData(string $detail, string $jsonPointer): ApiProblem
+    {
+        $problem = (new ApiProblem())
+            ->setType('https://api.publiq.be/probs/body/invalid-data')
+            ->setTitle('Invalid body data')
+            ->setDetail($detail)
+            ->setStatus(400);
+
+        $problem['jsonPointer'] = $jsonPointer;
```
LucWollants, line 94: Not a big deal but I wonder if we should make our own `ApiProblem` class which has a method to set a `jsonPointer`?

src/Http/Request/Body/ContentMediationRequestBodyParser.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Request\Body;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+final class ContentMediationRequestBodyParser implements RequestBodyParser
```
LucWollants, line 9: Just out of curiosity, how should `mediation` be interpreted in this case?

tests/Http/Offer/UpdateBookingAvailabilityRequestBodyParserTest.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->assertEquals($expected, $actual);
+    }
+
+    /**
+     * @test
+     */
+    public function it_fails_on_empty_body(): void
+    {
+        $given = $this->requestBuilder->withBodyFromString('')->build('PUT');
+        $this->expectException(ApiProblemException::class);
```
LucWollants, line 50: I wonder if we should also try to validate the `type` of the `ApiProblem` here that should be `https://api.publiq.be/probs/body/missing`

src/Http/Request/Body/ContentMediationRequestBodyParser.php
bertramakers: Haha it should be "Negotiation". Don't know how that word got in my head instead. https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation

bertramakers: I'll change it!

tests/Http/Offer/UpdateBookingAvailabilityRequestBodyParserTest.php
bertramakers: I also considered it but then we have to do a `try/catch` and use `$this->fail()` inside the `try` after the line with the `->parse()`. It's doable but a bit dirty. But indeed a better test. What do you think?

LucWollants: I guess it can be useful to already verify the type because in the future we will have more types and it's important that they match. I checked the PHPUnit docs but it seems there is no method to get hold of the thrown exception itself. So an extra try/catch seems to be required.

bertramakers: I'll see if I can find a more elegant solution and if not I'll do it with a try/catch. Maybe in a reusable trait then so we can use it in other tests to check the problem type.

src/Http/ApiProblem/ApiProblems.php
bertramakers: I'll check in how many places we already use this `ApiProblem` class and if it's easy to replace them with an instance of our own class. Because creating our own does not seem a lot of work and I see the added value (also in having 1 less dependency, especially for such a simple class), but I'd like to not have 2 classes named `ApiProblem` in our codebase and replacing all occurences might be some work. I'll investigate that first and if it's too much I'll create a follow-up ticket.

bertramakers: There are 110 usages, of which 49 in our own code and the rest in the `vendor` directory. So 49 that we would need to replace. Only 21 "new instance creations" though, 4 of which in tests. The rest are just imports, types, and comments. So it seems doable. But maybe something for another PR though, so I'll do that next but not in this PR.

LucWollants: Perfect approach. Thanks!

tests/Http/Offer/UpdateBookingAvailabilityRequestBodyParserTest.php
LucWollants: I notice some people have been asking for an `assertException` which has the exception as an input parameter. But that's not part of PHPUnit https://github.com/sebastianbergmann/phpunit/issues/3071

src/Http/Request/Body/ContentMediationRequestBodyParser.php
bertramakers: See 5625e05 (I also fixed a merge conflict in 4954700 because I forgot to pull in your changes and you changed the docblock to a property type, so I had to rename it in that as well).

LucWollants (approved):

tests/Http/Offer/UpdateBookingAvailabilityRequestBodyParserTest.php
LucWollants: Nice approach with the callable inside the trait!

PR #699, opened 2021-08-09
Internalize api problem class

bertramakers:
### Added

- Our own `ApiProblem` value object, which is also an `Exception` so it doesn't need to be wrapped in an Exception class
- `toHttpFoundationResponse()` method on PSR7 response classes, so we don't need to duplicate classes like `ApiProblemJsonResponse` for places where we cannot use PSR7 yet

### Removed

- `crell/api-problem` dependency
- HttpFoundation version of `ApiProblemJsonResponse`
- `ApiProblemException`

---

I've added a `blank()` factory method to `ApiProblem` so as not to change existing (bad) API problems that use `about:blank`. I've marked this as deprecated so no one uses it for new cases, and we can easily find the bad problems. One especially problematic usage will be in `WebErrorHandlerProvider` where basically every exception is converted to an `about:blank` API problem. But cleaning that up seems better for a separate ticket to improve the errors in general.

I tried to make the constructor of `ApiProblem` private so that it can only be created via static factory methods, but because the constructor of `Exception` is public that's not possible apparently. So I made it always construct a 500 Internal Server Error, as some kind of "default".

❓ When to use `throw ApiProblem` and when to use `return new ApiProblemJsonResponse`
➡️ Personally I would use `throw ApiProblem` as much as possible since it will get converted to `ApiProblemJsonResponse` automatically by the error handler, it's one import less, and it better resembles the error flow of PHP apps in general (throwing instead of returning). We still need `ApiProblemJsonResponse` for places where we NEED to work with a Response though, like in middlewares and symfony's security component.


bertramakers:
@LucWollants I still need to test this manually a bit, but the code changes themselves should be ready for review

LucWollants (approved):
Very clean and easy to use!

PR #700, opened 2021-08-10
Fix "Trying to clone an uncloneable object of class CultuurNet\UDB3\Http\ApiProblem\ApiProblem"

bertramakers:
### Fixed

- Fixed "Trying to clone an uncloneable object of class CultuurNet\UDB3\Http\ApiProblem\ApiProblem"


LucWollants:
It seems that the `Exception` class can not be cloned, see https://www.php.net/manual/en/exception.clone.php

LucWollants (approved):

PR #701, opened 2021-08-11
III-4149 Patch sub events

LucWollants:
### Added
- Added VO `SubEventUpdate` to describe sub event change
- Added new command `UpdateSubEvents`
- Added new handler for the `UpdateSubEvents` handler

### Removed
- Removed the obsolete command and handlers, see https://github.com/cultuurnet/udb3-silex/pull/702

---
Ticket: https://jira.uitdatabank.be/browse/III-4149


src/Event/ValueObjects/SubEventUpdate.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function withBookingAvailability(BookingAvailability $bookingAvailability): SubEventUpdate
+    {
+        $clone = clone $this;
+        $clone->bookingAvailability = $bookingAvailability;
+        return $clone;
+    }
+
+    public function getSubEventIndex(): int
```
bertramakers, line 36: Nit: Maybe this would better be named `getSubEventId()` because it's called `id` in the JSON of the update request, and there's also a ticket in consideration to add it to the JSON of the complete event details: https://jira.uitdatabank.be/browse/III-4126
I know that it's actually just the index in the list, but in the long run this name might be less confusing if we just talk about "id" everywhere.

src/Event/Commands/UpdateSubEvents.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @var SubEventUpdate[]
+     */
+    private array $subEventUpdates;
+
+    public function __construct(string $eventId, SubEventUpdate ...$updateSubEvents)
+    {
+        if (empty($updateSubEvents)) {
+            throw new InvalidArgumentException('At least one UpdateSubEvent is required');
```
bertramakers, line 24:
```suggestion
            throw new InvalidArgumentException('At least one SubEventUpdate is required');
```

src/Event/Event.php
```diff
@@ -374,6 +376,49 @@ public function updateSubEventsStatus(array $statuses): void
...
+                $subEventUpdate->getBookingAvailability() ?: $timestamp->getBookingAvailability()
+            );
+
+            $timestamps[$index] = $updatedTimestamp;
+        }
+
+        $updatedCalendar = new Calendar(
+            $this->calendar->getType(),
+            $this->calendar->getStartDate(),
+            $this->calendar->getEndDate(),
```
bertramakers, line 410: Nit: I think the startDate and endDate can be `null` if the `Calendar` has subEvents/timestamps, because the `Calendar` will then calculate them automatically anyway? Afaik `startDate` and `endDate` only need to be specified explicitly in the constructor for `periodic` calendars.

src/Http/Event/UpdateSubEventsRequestHandler.php
```diff
@@ -0,0 +1,80 @@
...
+            if (isset($update['bookingAvailability'])) {
+                $subEventUpdate = $subEventUpdate->withBookingAvailability(
+                    BookingAvailability::fromNative($update['bookingAvailability']['type'])
+                );
+            }
+
+            $updateSubEvents[] = $subEventUpdate;
+        }
+
+        $this->commandBus->dispatch(new UpdateSubEvents($eventId, ...$updateSubEvents));
```
bertramakers, line 59: Nit: This will now throw if there are no `SubEventUpdate` objects inside the array. Would you suggest that I make the JSON schema for the validation require at least one subEvent so that error would get caught earlier and converted to the right ApiProblem (`body/invalid-data`), or do we just check here if the array count is `>= 1` and if not just return `NoContent` and do nothing, or remove the check from the command and also accept an empty array (which would make the command handler do nothing if I'm not mistaken so we'd also return `NoContent` without actually doing anything then).

bertramakers (commented):
Nice work, just some minor feedback/questions 👌 As discussed I will add the request validation in another branch and make a PR that points to this one.

src/Event/ValueObjects/SubEventUpdate.php
LucWollants: I changed the name to `subEventId`

src/Event/Event.php
LucWollants: Good point, makes it already easier for the update of start and end date also.

src/Http/Event/UpdateSubEventsRequestHandler.php
LucWollants: I did place the validation inside `UpdateSubEvents` if we would have another entry point then an API call, for example a CLI command, in that case the `UpdateSubEvents` would throw.
I also opted for a throw because in theory this requires at least one `SubEventUpdate` as defined by the type hinting, but how PHP handles the splat operator it also allows an empty array.

So I would keep it here, but also don't mind if it gets removed. It's more of a safety against a "missing" language feature of typed arrays or generics.

src/Event/Event.php
LucWollants: This seems to have a side-effect. The accessor methods `getStartDate` and `getEndDate` do calculate it based on the time stamps. But internal code directly uses the `startDate` and `endDate` properties which are empty.

LucWollants: Only an issue with the tests creating the `CalendarUpdated` event but not on the projectors itself. So fixes the failing tests and the rest is fine as it is now.

src/Http/Event/UpdateSubEventsRequestHandler.php
bertramakers: I was just thinking that in theory it wouldn't harm if someone would dispatch this command without a any `SubEventUpdate` inside, it could just be a no-op then. But yeah I don't mind keeping it if we either add a check here then so we don't dispatch the command if there are no updates inside, OR we make the request validation expect at least one subEvent update and return an ApiProblem if you just do a request with literally `[]`

LucWollants: Ok, sounds good. Then we keep it like it is for now. Easy enough to refactor something like this if it would become inconsistent.

bertramakers (approved):
🎉

PR #703, opened 2021-08-12
III-4149 Validation for subEvent(s) updates

bertramakers:
### Added

- Added validation of subEvent update request body using JSON schema
- Added a generic way to validate request bodies using JSON schema

### Changed

- `ApiProblem::invalidBodyData()` to use schema errors (multiple possible) instead of 1 jsonPointer

---
Ticket: https://jira.uitdatabank.be/browse/III-4149

Todo:

- [x] Investigate if we can just use a complete OpenAPI file instead of having a schema per incoming request, so we don't need to store separate schema files in Stoplight and can keep everything inside one OpenAPI file (easier to work with in Stoplight). Using an OpenAPI file, we could also make the request body parser find the correct schema automatically because it could detect what URL and method are being requested and find the correct body schema via that in the OpenAPI file. For example using https://github.com/thephpleague/openapi-psr7-validator (however that doesn't have a stable 1.0 release yet, we'd need to check if it supports OpenAPI 3.1.0, we'd need to check if the errors it produces have jsonPointers, etc) **-> We're sticking to separate files with JSON schemas for now**
- [x] Find a way to dynamically pull in the OpenAPI or JSON schema file(s) instead of copying them over from the Stoplight repository and risking that they get out of sync. Some possible solutions:
  - Loading the schema via an URL instead of a local file. Easiest by possible problems: URL not reachable, OpenAPI or JSON schema file changed and it's not compatible yet because we still need to make changes in UDB, ... (I.e. the request validation can crash due to external factors outside the control of UDB and we will need to coordinate updates to the schema with updates to UDB) -> Not really an option IMHO
  - Pulling in the schema at "build"-time (in Jenkins, using a custom composer script for example) via an URL and storing it on the filesystem. This way it can't suddenly change at runtime, but it can still suddenly change after a random deploy if the schema has changed at some point -> Also not really an option IMHO unless the schema that we pull in at build-time is pulled in via an URL that points to a specific commit, random example: https://github.com/cultuurnet/stoplight-docs-uitdatabank/blob/0003ca33c6a0f4cec7979199e7f4fc4cedb6e76c/models/event-bookingInfo.json -> this will never change because the commit is in the URL)
  - Adding the https://github.com/cultuurnet/stoplight-docs-uitdatabank repository as a git submodule. This way we can also "freeze" it at a specific commit so the files don't suddenly change, but we can still easily update them by updating the git submodule reference to a newer commit. But git submodules always feel a bit dirty IMHO.
  - **-> We ended up going with a composer package for the Stoplight schema's**
- [x] Make it easier to construct a request body parser for new request handlers (reduce necessary boilerplate code)


src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,55 @@
...
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Http\ApiProblem\SchemaError;
+use Opis\JsonSchema\Errors\ErrorFormatter;
+use Opis\JsonSchema\Validator;
+use Psr\Http\Message\ServerRequestInterface;
+
+final class JsonSchemaValidatingRequestBodyParser implements RequestBodyParser
+{
+    private const SCHEMA_LOCATIONS = __DIR__ . '/../../../../vendor/publiq/stoplight-docs-uitdatabank/models/';
+    private const MAX_ERRORS = 100;
```
bertramakers, line 16: I opted for private constants here for now because there's no real use case to make them configurable per instance now. Only downside is that this couples this code a bit to how the `publiq/stoplight-docs-uitdatabank` package is structured but for loading the schemas from the filesystem I think that is not a really big problem and it makes it a lot easier to instantiate this class this way.

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,55 @@
...
+
+    private RequestBodyParser $baseParser;
+    private Validator $validator;
+    private string $jsonSchema;
+
+    private function __construct(string $jsonSchema)
+    {
+        $this->jsonSchema = $jsonSchema;
+        $this->baseParser = new JsonRequestBodyParser();
```
bertramakers, line 25: This will need to become configurable in a follow-up PR to make it possible to add use a custom `RequestBodyParser` that first transforms the data before validating it. But since that is not needed for the endpoints that use it now (`PUT /events/{eventId}/bookingAvailability` and `PATCH /events/{eventId}/subEvents`) I decided to hardcode it to `JsonRequestBodyParser` for now here.

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,55 @@
...
+
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Http\ApiProblem\SchemaError;
+use Opis\JsonSchema\Errors\ErrorFormatter;
+use Opis\JsonSchema\Validator;
+use Psr\Http\Message\ServerRequestInterface;
+
+final class JsonSchemaValidatingRequestBodyParser implements RequestBodyParser
+{
+    private const SCHEMA_LOCATIONS = __DIR__ . '/../../../../vendor/publiq/stoplight-docs-uitdatabank/models/';
```
LucWollants, line 15: Maybe we can add one PHP file inside `publiq/stoplight-docs-uitdatabank` with some constants for file names and file locations? Then we have a bit more flexibility inside the `publiq/stoplight-docs-uitdatabank` package on where to store something without having to modify these hardcoded paths.

src/Http/Request/Body/ContentNegotiationRequestBodyParser.php
```diff
@@ -8,14 +8,28 @@
...
 {
-    public function parse(ServerRequestInterface $request): array
+    private RequestBodyParser $jsonRequestBodyParser;
+
+    public function __construct()
+    {
+        $this->jsonRequestBodyParser = new JsonRequestBodyParser();
+    }
+
+    public function withJsonRequestBodyParser(RequestBodyParser $jsonRequestBodyParser)
```
LucWollants, line 18: Maybe remove this method and add a required parameter to the constructor? Then the caller decides to construct a `JsonRequestBodyParser` on the fly or for example use the new `JsonSchemaValidatingRequestBodyParser`

tests/Http/Event/UpdateSubEventsRequestHandlerTest.php
```diff
@@ -0,0 +1,287 @@
...
+                ),
+            ],
+        ];
+    }
+
+    /**
+     * @test
+     * @dataProvider invalidDataProvider
+     */
+    public function it_throws_an_api_problem_when_given_with_valid_data($data, array $expectedSchemaErrors): void
```
LucWollants, line 186: I think it should be `invalid_data` instead of `valid_data`

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $this->jsonSchema = $jsonSchema;
+        $this->baseParser = new JsonRequestBodyParser();
+        $this->validator = new Validator(null, self::MAX_ERRORS);
+    }
+
+    public static function forStoplightModelFileName(string $stoplightModelFileName): self
+    {
+        $stoplightModelFileName = ltrim($stoplightModelFileName, '/');
+        $schema = file_get_contents(self::SCHEMA_LOCATIONS . $stoplightModelFileName);
```
LucWollants, line 32: I guess the impact of this disk IO will be minimal? I know we should not do any premature performance optimisations but I'm always sceptical if I see any disk IO 😉

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
LucWollants: I made a remark about the tight coupling here https://github.com/cultuurnet/udb3-silex/pull/703#discussion_r690993746

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
bertramakers: Should be the same as the disk IO of loading the yaml config files. And it should usually only be one schema per request.

tests/Http/Event/UpdateSubEventsRequestHandlerTest.php
bertramakers: Good find 🙈 Fixed in 9e8c390

src/Http/Request/Body/ContentNegotiationRequestBodyParser.php
bertramakers: I'm just going to remove this `ContentNegotiationRequestBodyParser` because the more I think about it, the less there seems to be a need for it:

- We currently only support JSON
- If we want to support other formats in the future, we can for example convert them to JSON first when they enter the application and then we don't need to support multiple formats in each controller / request handler

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
LucWollants: Yes, so no need to worry about it 😀

src/Http/Request/Body/JsonSchemaValidatingRequestBodyParser.php
bertramakers: As discussed I added a file with constants in this PR instead.

src/Http/Request/Body/ContentNegotiationRequestBodyParser.php
bertramakers: Removed!

src/Http/Offer/UpdateBookingAvailabilityRequestHandler.php
```diff
@@ -35,6 +36,7 @@ public function __construct(
 
     public function handle(ServerRequestInterface $request, string $offerId): ResponseInterface
     {
+        /** @var stdClass $data */
```
bertramakers, line 39: Added this because PHPStan complained `Cannot access property $type on array|object.` in this file. Not sure why it complains when the return type is `array|object` as defined in the PSR7 standard for `getParsedBody()`, but does not when using `stdClass` (= an object).

bertramakers: Okay apparently it also works if I just typehint it to `object`. Maybe because you cannot do `->` on an array. That would also explain why it doesn't complain in the other request handler where we use `RequestBodyParser`, because there the data is an array and we use `foreach ()` on it which works for both `array` and `object`. Not sure how to fix this easily without having to typehint/typecast in request handlers where an object is expected 🤔

bertramakers: For now I just typecasted the return value to `(object)` in this request handler.

- If it's already an object (as it should be or the validation should throw) it will remain an object
- If it's an associative array for some reason, it will become an object. If the associative array had the right "properties", everything will keep working. If it didn't, the app will crash when an undefined property is accessed but it should do so anyway because the data was not correct then and the validation didn't catch it. So this should crash and get reported in Sentry and we should fix the validation then.
- If it's a sequential array for some reason, it will also become an object with properties for the sequential values. Of course the expected properties won't be set then, but again the app will crash as it should as soon as it tries to access one.

Overall I think this is just a part of the app where we would expect it to crash if we expect an object but get an array, because then the data was bad and the validation didn't catch it (= bug = internal server error). But of course PHPStan tries to warn us that we might be dealing with an array here, but there's not much we can do about it except add an `if (!is_object($data)) { throw ApiProblem::invalidBodyData(...); }` in every request handler where we expect an object but that's a lot for something that the validation layer should catch. Or we cast it and we "trick" PHPStan into thinking we're good.

LucWollants (approved):
Nice improvements with the combined parser!

PR #706, opened 2021-08-18
III-4070 Improved token cache

LucWollants:
### Added
- Added new `Auth0Token`

### Changed
- Use `Auth0Token` inside `Auth0ManagementTokenRepository` and `Auth0ManagementTokenGenerator`
- Use expired value inside `Auth0ManagementTokenProvider` to determine a token that is about to expire

---
Ticket: https://jira.uitdatabank.be/browse/III-4070

### DIFF
--- a/src/User/CacheRepository.php
+++ b/src/User/CacheRepository.php
@@ -4,31 +4,46 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\User;
 
+use DateTimeImmutable;
 use Doctrine\Common\Cache\Cache;
 
 class CacheRepository implements Auth0ManagementTokenRepository
 {
-    /**
-     * @var Cache
-     */
-    private $cache;
+    private const TOKEN_KEY = 'auth0_token';
 
+    private Cache $cache;
 
     public function __construct(Cache $cache)
     {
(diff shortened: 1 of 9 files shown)


src/User/CacheRepository.php
```diff
@@ -4,31 +4,46 @@
...
-    public function token(): ?string
+    public function token(): ?Auth0Token
     {
-        if (!$this->cache->contains('token')) {
+        if (!$this->cache->contains(self::TOKEN_KEY)) {
             return null;
         }
-        return $this->cache->fetch('token');
+
+        $tokenAsArray = json_decode($this->cache->fetch(self::TOKEN_KEY), true);
```
bertramakers, line 27: Nit: The token that's currently in the cache (if any) when we deploy this will just be a string, so maybe we need to check if `$tokenAsArray` is false and if so also return `null`.

LucWollants: To solve that I changed the key value for Redis. So we create a new key/value pair the first time it gets deployed.

bertramakers: Perfect 👍

bertramakers (approved):

PR #712, opened 2021-08-23
III-4168 fix $ref in schemas

bertramakers:
### Changed

- Load JSON schemas via URI, by configuring a resolver on the `Validator`

### Fixed

- Resolving of `$ref` in JSON schema validation

---
Ticket: https://jira.uitdatabank.be/browse/III-4168

I noticed this while working with `$ref` in the changes to make the calendar endpoint work with JSON schema validation. So the next PR will better cover this with tests, when I add tests for the JSON schema validation of that endpoint.


LucWollants (approved):
Nice work on finding this solution

LucWollants:
Thanks for refactor to make the naming and single responsibilities clearer.

LucWollants (approved):

PR #713, opened 2021-08-24
III-4168 Use associative arrays in request handlers

bertramakers:
### Changed

- Parsed body is now associative arrays instead of objects (after validation), so we can use the existing denormalization/deserialization/etc logic to convert the JSON to value objects

---
Ticket: https://jira.uitdatabank.be/browse/III-4168


LucWollants (approved):
I already approved, but should `AssociativeArrayRequestBodyParser` get a unit test?

bertramakers:
Ah I overlooked your comment. I'll add one in another PR!

bertramakers:
I figured the most important parts were covered by the existing tests for the request handlers, but can't hurt to add one

PR #715, opened 2021-08-24
III-4168 Use denormalizers in request handlers

bertramakers:
### Changed

- Existing request handlers now use Model Denormalizers
- `RequestBodyParserFactory::createBaseParser()` doesn't append `AssociativeArrayRequestBodyParser` by default anymore (because this breaks when one of the request body parsers denormalizes the data first, and for that denormalization the data already needs to be associative)

### Added

- Model VO and Denormalizer for `SubEventUpdate`
- Denormalizer for Model `Status` (extracted from `CalendarDenormalizer`)
- Denormalizer for Model `BookingAvailability` (extracted from `CalendarDenormalizer`)


src/Event/ValueObjects/SubEventUpdate.php
```diff
@@ -47,4 +48,26 @@ public function getBookingAvailability(): ?BookingAvailability
     {
         return $this->bookingAvailability;
     }
+
+    public static function fromUdb3ModelsSubEventUpdate(
+        Udb3ModelsSubEventUpdate $udb3ModelsSubEventUpdate
+    ): self {
+        $update = new self($udb3ModelsSubEventUpdate->getSubEventId());
+
+        $udb3ModelsStatus = $udb3ModelsSubEventUpdate->getStatus();
```
LucWollants, line 57: Because this code is inside the `SubEventUpdate` class itself, this shorthand would also be valid
```
$update->status = $udb3ModelsSubEventUpdate->getStatus() ? Status::fromUdb3ModelStatus($udb3ModelsSubEventUpdate->getStatus()) : null;
```
But I am perfectly happy with the `with` methods 😊

src/Http/Event/UpdateSubEventsRequestHandler.php
```diff
@@ -34,7 +31,8 @@ public function __construct(CommandBus $commandBus)
         $this->commandBus = $commandBus;
 
         $this->updateSubEventsParser = RequestBodyParserFactory::createBaseParser(
-            JsonSchemaValidatingRequestBodyParser::fromFile(JsonSchemaLocator::EVENT_SUB_EVENT_PATCH)
+            new JsonSchemaValidatingRequestBodyParser(JsonSchemaLocator::EVENT_SUB_EVENT_PATCH),
+            new DenormalizingRequestBodyParser(new SubEventUpdatesDenormalizer(), SubEventUpdates::class)
```
LucWollants, line 35: If we extend this Symfony interface with one method `getClassName` we don't need the extra parameter  `SubEventUpdates::class` and can remove `\CultuurNet\UDB3\Event\ValueObjects\` 

So the `SubEventUpdatesDenormalizer` will need to implement one extra method, but that encapsulates the class name better.

src/Model/ValueObject/Calendar/SubEventUpdate.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Model\ValueObject\Calendar;
+
+final class SubEventUpdate
```
LucWollants, line 7: The class `SubEventUpdate` inside the `CultuurNet\UDB3\Event\ValueObjects` namespace is rather new. Maybe we can discard `CultuurNet\UDB3\Event\ValueObjects\SubEventUpdate` and directly start using `CultuurNet\UDB3\Model\ValueObject\Calendar\SubEventUpdate`? I believe we wanted to make more use of the `Udb3Models` and avoid duplicating VOs?

bertramakers: Done!

src/Http/Event/UpdateSubEventsRequestHandler.php
bertramakers: As agreed on Slack I've just aliased the Legacy VOs for now. I'll also mark more VOs as deprecated in another PR.

src/Event/ValueObjects/SubEventUpdate.php
bertramakers: This class does not exist anymore after your other suggestion to remove it, so "done" 😉

LucWollants (approved):
Thanks for the extra improvements.

PR #716, opened 2021-08-25
III-4169 Mark old value objects as deprecated

bertramakers:
### Added

- Added `@deprecated` warnings to all old value objects I could find

### Removed

- Removed abstract `TrimmedString` value object that was only used in 2 classes (one of which deprecated) and offered little added value


src/Offer/ValueObjects/BookingAvailabilityType.php
```diff
@@ -6,6 +6,10 @@
 
 use InvalidArgumentException;
 
+/**
+ * @deprecated
+ *   Use Use CultuurNet\UDB3\Model\ValueObject\Calendar\BookingAvailabilityType instead where possible.
```
LucWollants, line 11: Small typo here 😉

src/ValueObject/SapiVersion.php
```diff
@@ -7,6 +7,9 @@
 use ValueObjects\Enum\Enum;
 
 /**
+ * @deprecated
+ *   Everything should use V3 by now so remove usages and this class.
```
LucWollants, line 11: Sounds like a good candidate for a separate ticket?

LucWollants (approved):

src/Offer/ValueObjects/BookingAvailabilityType.php
bertramakers: Thanks, I'll fix it! I've also found more VOs 🙈

src/ValueObject/SapiVersion.php
bertramakers: Created one: https://jira.uitdatabank.be/browse/III-4185

src/Offer/ValueObjects/BookingAvailabilityType.php
bertramakers: Fixed and also deprecated remaining VOs I found

PR #728, opened 2021-09-07
III-4126 sub event ids

bertramakers:
### Added

- Added `id` property to `subEvent` objects on event JSON-LD

---
Ticket: https://jira.uitdatabank.be/browse/III-4126


src/Calendar.php
```diff
@@ -325,11 +325,11 @@ public function toJsonLd(): array
 
         $jsonLd['bookingAvailability'] = $this->determineCorrectTopBookingAvailabilityForProjection()->serialize();
 
-        $timestamps = $this->getTimestamps();
+        $timestamps = array_values($this->getTimestamps());
```
LucWollants, line 328: I find this one of the misleading internal function in PHP
> array_values() returns all the values from the array and indexes the array numerically.
You would not expect this to index the array 😅

src/Offer/ReadModel/JSONLD/NewPropertyPolyfillOfferRepository.php
```diff
@@ -77,7 +78,8 @@ function (array $subEvent) {
                     $subEvent
                 );
             },
-            $json['subEvent']
+            $json['subEvent'],
+            range(0, count($json['subEvent']) - 1)
```
LucWollants, line 82: Clever 👌

LucWollants (approved):
Good improvement to make read and write models more equal

src/Calendar.php
bertramakers: I think it makes sense. 😄 It's the opposite of `array_keys()` which returns a list of the given array's keys only, so `array_values()` returns a list of the given array's values only (i.e. without keys). Lists are sequential arrays in PHP, so their "keys" are 0, 1, 2, etc.

Maybe it makes more sense with an example like this:
```
$associative = [
  'a' => 'foo', 
  'b' => 'bar'
];

$keys = array_keys($associative);
$values = array_values($associative);

var_dump($keys === ['a', 'b']); // true
var_dump($keys === [0 => 'a', 1 => 'b']); // true

var_dump($values === ['foo', 'bar']); // true
var_dump($values === [0 => 'foo', 1 => 'bar']); // true
```

PR #737, opened 2021-09-23
III-4220 Add CLI command to update booking availability

LucWollants:
### Added
- Added new CLI command to update the booking availability of events based on custom query and with calendar type single or multiple

---
Ticket: https://jira.uitdatabank.be/browse/III-4220

### DIFF
--- a/app/Console/UpdateBookingAvailabilityCommand.php
+++ b/app/Console/UpdateBookingAvailabilityCommand.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Silex\Console;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Offer\Commands\UpdateBookingAvailability;
+use CultuurNet\UDB3\Offer\ValueObjects\BookingAvailability;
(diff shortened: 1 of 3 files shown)


jdrieghe (approved):

app/Console/UpdateBookingAvailabilityCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        parent::__construct($commandBus);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $sapiQuestion = new Question(
+            'Provide SAPI query for events to update (periodic and permanent events are excluded by default)' . PHP_EOL
+        );
+        $sapiQuery = $this->getHelper('question')->ask($input, $output, $sapiQuestion);
+        $sapiQuery .= ' AND (calendarType:single OR calendarType:multiple)';
```
bertramakers, line 45: I think we need to wrap the original query inside brackets here to be safe, to prevent issues like we had in https://jira.uitdatabank.be/browse/III-3061

LucWollants: Nice catch, will be fixed with https://jira.uitdatabank.be/browse/III-4228

PR #739, opened 2021-09-24
III-4185 Remove SapiVersion value object

LucWollants:
### Removed
- Removed the `SapiVersion` value object and added a fallback to 'v3'

Note: it is still present in the URL of the endpoints to update the constraints and inside the projection of the role

---
Ticket: https://jira.uitdatabank.be/browse/III-4185


jdrieghe (approved):
Looks good, thanks for the step-by-step commits!
For future reference: it's easier to review when there are separate commits for functional changes and structural changes.
E.g. some commits changed the `SapiVersion` dependency and cleaned up a bunch of type hints at the same time. Splitting those commits reduces cognitive load during review.

bertramakers (approved):
Nice work @LucWollants! 👍 I think we should also pick up the suggestion you added in the ticket, to change the API endpoint and the frontend to remove the version from the URL. But in another PR of course.

LucWollants:
> For future reference: it's easier to review when there are separate commits for functional changes and structural changes.

@jdrieghe I fully agree, I went a bit too aggressive with the PHP 7.4 tweaks 😄

PR #746, opened 2021-09-30
III-4177 Singular/plural events & places endpoint support

bertramakers:
### Changed

- Every event endpoint now works with both `/event/` and `/events/` to avoid confusion and for consistency
- Every place endpoint now works with both `/place/` and `/places/` to avoid confusion and for consistency
- [internal] RequestHandler classes now implement PSR-15 `RequestHandlerInterface`
- [internal] You can now use RequestHandler class names as controllers, without having to append `. ':handle'`
- [internal] General cleanup of EventControllerProvider, PlaceControllerProvider, OfferControllerProvider

### Fixed

- Routes that do not exist now return a `404` instead of a `400`

### Added

- [internal] Added `psr/http-server-middleware` for PSR-15 `RequestHandlerInterface`

---
Ticket: https://jira.uitdatabank.be/browse/III-4177

Note: I started off trying to use a PSR-15 middleware to rewrite the URL of incoming requests from `/event/` to `/events/` and from `/place/` to `/places/` as was written as a suggestion in the ticket. But due to middlewares working very differently in Silex (hard to modify the request URL, and returning `null` when the middleware does not return a Response instead of calling the next request handler as in PSR-15) it was practically impossible to use a PSR-15 middleware and wrap it in an adapter for Silex. Also just using a Silex `->before()` middleware in itself was pretty hard because it's hard to change the URL of the incoming request in that middleware. So instead I just mounted the controller providers for the event and place routes both as the singular and plural names for now. Once we use a PSR7 router like we do in SAPI3 and the JWT provider we can more easily do this with a middleware.
**Edit**: I was able to register a "catch-all" route for URLs that do not match any route, and rewrite the URL in there and then send an internal sub-request.


bertramakers:
Converting this back to a draft because I think I found a way to not have to register the controller providers twice (once for the singular and again for the plural).

bertramakers:
Okay the new approach with the catch-all route at the end of the route registrations makes is possible to rewrite URLs, which we'll also need for https://jira.uitdatabank.be/browse/III-4178

LucWollants (approved):
Nice work! Only one failing acceptance but that will be fixed in another PR.

PR #747, opened 2021-09-30
III-4232 add video feature

JonasVHG:
### Added

- Add VideoAdded Event

---
Ticket: https://jira.uitdatabank.be/browse/III-4232


tests/Offer/CommandHandlers/AddVideoHandlerTest.php
```diff
@@ -50,7 +50,7 @@ protected function createCommandHandler(
     /**
      * @test
      */
-    public function it_will_add_video_to_event(): void
+    public function it_will_add_a_video_to_an_offer(): void
```
LucWollants, line 53: The name can stay, because it's still an event that is tested here.

app/Offer/OfferControllerProvider.php
```diff
@@ -48,8 +50,11 @@ public function connect(Application $app): ControllerCollection
         $controllers->put('/{offerId}/status/', UpdateStatusRequestHandler::class);
         $controllers->put('/{offerId}/booking-availability/', UpdateBookingAvailabilityRequestHandler::class);
 
-        $controllers->put('/{cdbid}/type/{typeId}/', "{$controllerName}:updateType");
-        $controllers->put('/{cdbid}/theme/{themeId}/', "{$controllerName}:updateTheme");
+        $controllers->post('/{offerId}/videos', AddVideoRequestHandler::class);
```
bertramakers, line 53: This will need to be registered with a trailing slash after you merge in the latest commits from master. (All routes should be registered with a trailing slash since a recent commit.)
It will keep working with and without trailing slash from the outside though, so the acceptance tests that use `/videos` (no trailing slash) for example should keep working.

src/Event/Event.php
```diff
@@ -64,7 +65,9 @@
 use CultuurNet\UDB3\Media\Image;
 use CultuurNet\UDB3\Model\ValueObject\Calendar\SubEventUpdate;
 use CultuurNet\UDB3\Model\ValueObject\MediaObject\CopyrightHolder;
+use CultuurNet\UDB3\Model\ValueObject\MediaObject\Video;
 use CultuurNet\UDB3\Model\ValueObject\Taxonomy\Label\Labels;
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID as ModelUUID;
```
bertramakers, line 70: Nit: Preferable we'd alias the old `UUID` class instead, so when we remove that one later we don't need to clean up this alias on the new class later.

app/Offer/OfferControllerProvider.php
LucWollants: Sorry for that, I did a wrong merge and forgot about the trailing slash.

src/Event/Event.php
LucWollants: Good point! I replaced all usages:
```
use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
use ValueObjects\Identity\UUID as LegacyUUID;
```

PR #748, opened 2021-09-30
III-4178 Use kebab-casing in URLs and automatically transform known camelCase operations/resources

bertramakers:
### Changed

- All URLs now use `kebab-casing`
- Existing routes with `camelCasing` automatically get rewritten to `kebab-casing` internally for backward compatibility

---
Ticket: https://jira.uitdatabank.be/browse/III-4178

This builds upon the work in #746 so that PR as well as its related Stoplight docs PR (https://github.com/cultuurnet/stoplight-docs-uitdatabank/pull/24) should be merged first


LucWollants (approved):
Step by step the consistency is approved!

As discussed this will not break the AngularJS app and is not worth the effort to change it. Maybe when a bigger refactor is needed for a new feature (for example video urls) it can be evaluated if making it also consistent in AngularJS is an option.

PR #750, opened 2021-10-04
III-4148 ignore trailing slashes

bertramakers:
### Changed

- Routes can now be accessed with our without trailing slash, but they MUST be registered with a trailing slash for this to work! If you register a route without a trailing slash it will work without the trailing slash, but it won't work with a trailing slash.

---
Ticket: https://jira.uitdatabank.be/browse/III-4148

Note: For this to be possible the internal routes have to be registered in a consistent manner, so either all with or all without trailing slash. I had to pick trailing slash, because if you do for example `$app->mount('/places', new PlaceControllerProvider())` and then `$app->post('/', function () { ... });` in that mounted controller provider, the `POST` is registered as `POST /places/` (so with trailing slash). And there is now way to do something like `$app->post('', function () { ... })` (no path).

This behavior is consistent with the logic that `/example` CAN be `/example/`, but it's not necessarily always the same. For example `/file.html` is not the same as `/file.html/`. So in some cases the route without trailing slash is the same as the route with trailing slash, unless you register it without trailing slash. Then it's only accessible without trailing slash.

See also https://github.com/silexphp/Silex/issues/149

### DIFF
--- a/app/Event/EventControllerProvider.php
+++ b/app/Event/EventControllerProvider.php
@@ -22,30 +22,30 @@ class EventControllerProvider implements ControllerProviderInterface, ServicePro
         $controllers = $app['controllers_factory'];
 
         $controllers->post('/', 'event_editing_controller:createEvent');
-        $controllers->get('/{cdbid}', 'event_controller:get');
-        $controllers->delete('/{cdbid}', 'event_editing_controller:deleteEvent');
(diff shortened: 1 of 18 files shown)


LucWollants (approved):
I did a manual test with several endpoints and they work both with and without a trailing slash

PR #751, opened 2021-10-05
III-4232 Add validation for the add video endpoint of place and event

LucWollants:
### Added
- Added validation for the add video endpoint of place and event

---
Ticket: https://jira.uitdatabank.be/browse/III-4232


tests/Http/Offer/AddVideoRequestHandlerTest.php
```diff
@@ -0,0 +1,149 @@
...
+        $this->assertCallableThrowsApiProblem(
+            ApiProblem::bodyInvalidData(
+                new SchemaError(
+                    '/url',
+                    'The string should match pattern: ^http(s?):\/\/(www\.)?((youtube\.com\/watch\?v=([^\/#&?]*))|(vimeo\.com\/([^\/#&?]*)))'
+                )
+            ),
+            fn () => $this->addVideoRequestHandler->handle($addVideoRequest)
+        );
+    }
```
bertramakers, line 148: Nit: A test that checks that `copyrightHolder` is a string and not for example an object with language keys could be useful as well

bertramakers (approved):

tests/Http/Offer/AddVideoRequestHandlerTest.php
LucWollants: I added an extra test with `copyrightHolder` value `123`

PR #774, opened 2021-10-20
III-4247 UpdateType request handler

bertramakers:
### Changed

- [internal] Refactored route to update offer (event)type to use a PSR-15 `RequestHandlerInterface` implementation
- [internal] Refactored Offer `UpdateType` command to not be abstract anymore
- [internal] Refactored handler of `UpdateType` to separate `CommandHandler` implementation
- [internal] Changed some usages of deprecated `EventType` VO class to `Category` VO class

### Removed

- [internal] Unused service & factory methods

---
Ticket: https://jira.uitdatabank.be/browse/III-4247


bertramakers:
I did a manual test using Postman. I will also add an acceptance test. Good that you pointed out that there isn't one yet, because I just ran them expecting them to test it.

PR #790, opened 2021-11-02
III-4287 Refactor the update organizer url

LucWollants:
### Added
- Added new `UpdateWebsite` command handler
- Added new `UpdateWebsite` request handler

### Changed
- Refactor UpdateWebsite command to work with `Url` from Model namespace
- Refactor Organizer::updateWebsite method to work with `Url` from Model namespace

### Removed
- Removed obsolete code for handling updateWebsite

---
Ticket: https://jira.uitdatabank.be/browse/III-4287

### DIFF
--- a/tests/Organizer/CommandHandler/UpdateWebsiteHandlerTest.php
+++ b/tests/Organizer/CommandHandler/UpdateWebsiteHandlerTest.php
@@ -9,12 +9,13 @@ use Broadway\CommandHandling\Testing\CommandHandlerScenarioTestCase;
 use Broadway\EventHandling\EventBus;
 use Broadway\EventStore\EventStore;
 use CultuurNet\UDB3\Language;
+use CultuurNet\UDB3\Model\ValueObject\Web\Url;
 use CultuurNet\UDB3\Organizer\Commands\UpdateWebsite;
 use CultuurNet\UDB3\Organizer\Events\OrganizerCreatedWithUniqueWebsite;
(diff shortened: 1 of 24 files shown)


tests/Organizer/CommandHandler/UpdateWebsiteHandlerTest.php
```diff
@@ -0,0 +1,46 @@
...
+            new Language('nl'),
+            Url::fromNative('https://www.madewithlove.be'),
+            new Title('Organizer Title')
+        );
+
+        $this->scenario
+            ->withAggregateId($id)
+            ->given([$organizerCreated])
+            ->when(new UpdateWebsite($id, Url::fromNative('https://www.publiq.be')))
+            ->then([new WebsiteUpdated($id, Url::fromNative('https://www.publiq.be'))]);
```
bertramakers: Nit: I usually also add a test in these command handler test files to assert that there is no event recorded when there was no change. I know this overlaps with the tests for the aggregate itself but it's easy to do here, saves us a test in another file, and tests the whole write side a bit more in an "integrated" way. But not blocking or required.

bertramakers (approved):
Looks great ✨ 👌

tests/Organizer/CommandHandler/UpdateWebsiteHandlerTest.php
LucWollants: That is indeed also an option using "features" tests instead of unit tests.

PR #791, opened 2021-11-02
III-4290 Refactor updating contact point of an organizer

LucWollants:
### Added
- Added new `UpdateContactPoint` command handler
- Added new `UpdateContactPoint` request handler

### Removed
- Removed obsolete code for handling UpdateContactPoint

---
Ticket: https://jira.uitdatabank.be/browse/III-4290


LucWollants:
I did not refactor `Organizer::updateContactPoint` to use the `ContactPoint` class from the `Model` namespace as this would lead to a lot of conversions because that method is also used inside `Organizer::create`.

src/Http/Organizer/LegacyContactPointRequestBodyParser.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Organizer;
+
+use CultuurNet\UDB3\Http\Request\Body\RequestBodyParser;
+use Psr\Http\Message\ServerRequestInterface;
+
+class LegacyContactPointRequestBodyParser implements RequestBodyParser
```
bertramakers, line 10: Nit: In other places this is usually called `BackwardCompatible...RequestBodyParser`. Either name is fine by me but maybe rename the others if you prefer `Legacy...RequestBodyParser`.

src/Organizer/Serializers/UpdateContactPointDenormalizer.php
```diff
@@ -0,0 +1,63 @@
...
+    public function __construct(string $organizerId)
+    {
+        $this->organizerId = $organizerId;
+    }
+
+    public function denormalize($data, $type, $format = null, array $context = []): UpdateContactPoint
+    {
+        $telephoneNumbers = [];
+        if (isset($data['phone'])) {
+            foreach ($data['phone'] as $phone) {
```
bertramakers, line 30: Nit: I think you can remove the `if (isset(...))` by doing this:

```
foreach ($data['phone'] ?? [] as $phone) {
  // ...
}
```

bertramakers: Or use `array_map` so you don't need the `$telephoneNumbers = [];` and make it a one-liner.

For example:

```
$telephoneNumbers = array_map(fn ($phone) => new TelephoneNumber($phone), $data['phone'] ?? []);
```

tests/Http/Organizer/UpdateContactPointRequestHandlerTest.php
```diff
@@ -0,0 +1,268 @@
...
+                    new EmailAddresses(new EmailAddress('info@publiq.be')),
+                    new Urls(new Url('https://www.publiq.be'))
+                ),
+            ],
+            'update multiple contact point information' => [
+                '{
+                    "phone": ["016 10 20 30", "016 11 22 33"],
+                    "email": ["info@publiq.be", "info@cn.be"],
+                    "url": ["https://www.publiq.be", "https://www.cn.be"]
+                }',
```
bertramakers, line 117: More a general comment / FYI: I'm definitely okay with hardcoding JSON in the tests, but if it becomes too complex to write you can also use `Json::encode([... your data as an assoc array ...])` if you want

bertramakers: And/or add a `withBodyFromArray()` or something like that to `Psr7RequestBuilder` that just takes an assoc array and converts it to Json so we don't have to repeat that in every test file.

tests/Http/Organizer/UpdateContactPointRequestHandlerTest.php
```diff
@@ -0,0 +1,268 @@
...
+            ->build('PUT');
+
+        $this->assertCallableThrowsApiProblem(
+            ApiProblem::bodyInvalidData(
+                new SchemaError('/phone/0', 'Minimum string length is 1, found 0')
+            ),
+            fn () => $this->updateContactPointRequestHandler->handle($updateUrlRequest)
+        );
+    }
+}
```
bertramakers, line 268: Very extensive tests, nice 👍

bertramakers (approved):
Looks great, just some minor suggestions or general comments. Nothing blocking or that _needs_ to be picked up, so approving already

src/Http/Organizer/LegacyContactPointRequestBodyParser.php
LucWollants: Nicely spotted. This was indeed a sub comment inside this commit https://github.com/cultuurnet/udb3-silex/pull/791/commits/5bd28e03cc52a4beb5a470d5cc14325b31d708e9

bertramakers: Ah I overlooked that because I reviewed the complete diff now for brevity's sake 😉

LucWollants: But both names are fine for me. You can pick what you prefer 😄

src/Organizer/Serializers/UpdateContactPointDenormalizer.php
LucWollants: The `array_map` is indeed much more elegant.

tests/Http/Organizer/UpdateContactPointRequestHandlerTest.php
LucWollants: I added this new method `withBodyFromArray`.

src/Http/Organizer/LegacyContactPointRequestBodyParser.php
LucWollants: As discussed on Slack I took the `Legacy` variant

PR #797, opened 2021-11-04
III-4238 add video to export

JonasVHG:
### Added

- Export feature for videos
- 
---
Ticket: https://jira.uitdatabank.be/browse/III-4238


src/EventExport/Format/TabularData/TabularDataEventFormatter.php
```diff
@@ -683,6 +687,26 @@ function ($tariffs, $cardSystemPrices) {
                 },
                 'property' => 'bookingAvailability',
             ],
+            'videos.url' => [
+                'name' => 'videos URL',
+                'include' => function ($event) {
+                    if (!property_exists($event, 'videos') || !is_array($event->videos)) {
```
LucWollants, line 693: You could pass `$event` to `formatVideo` instead of `array $videos`.
That would make it possible to do the check `if (!property_exists($event, 'videos') || !is_array($event->videos))` inside `formatVideo`

JonasVHG: Fixed

LucWollants (approved):
Nice work!

PR #799, opened 2021-11-05
Fix warning that $controller is not a string or int

bertramakers:
### Fixed

- Fixed warning that `$controller` is not a `string` or `int` when checking if the controller is a class name of a `RequestHandlerInterface` implementation (for example when the `$controller` is actually a closure)


LucWollants (approved):
I also did a test and this fixes the warning.

PR #800, opened 2021-11-05
III-4291 Refactor deleting organizer to PSR-7

LucWollants:
### Added
- Added new `DeleteOrganizerHandler` for the `DeleteOrganizer` command
- Added new `DeleteOrganizerRequestHandler` for the `DELETE` endpoint

### Changed
- Prevent multiple `OrganizerDeleted` events when deleting an organizer multiple times

### Removed
- Removed obsolete code

---
Ticket: https://jira.uitdatabank.be/browse/III-4291


src/Organizer/CommandHandler/DeleteOrganizerHandler.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function handle($command): void
+    {
+        if (!$command instanceof DeleteOrganizer) {
+            return;
+        }
+
+        $this->eventRelations->deleteOrganizer($command->getItemId());
+        $this->placeRelations->deleteOrganizer($command->getItemId());
```
bertramakers, line 35: Shouldn't this be handled by a read model projector? The relations tables are read models no?

bertramakers: Oh I see this uses the relation read models to find the related events/places and then dispatches `DeleteOrganizer` commands for them so it actually performs a write operation

bertramakers (approved):

PR #801, opened 2021-11-05
III-4261 Update event audience

bertramakers:
### Added
- Added new RequestHandler to update an event's audience(Type)
- Add new command handler for `UpdateAudience`
- Add API problem for "incompatible audience type"

### Changed
- Use new RequestHandler for `PUT /events/{eventId}/audience` route
- Refactor UpdateAudience to use new Model VOs

### Removed
- Remove old controller method to update event audience
- Remove old service method to update event audience (only used in old controller method)
- Remove old command handler for `UpdateAudience`

---
Ticket: https://jira.uitdatabank.be/browse/III-4261

**Note:** 
Apparently when an event's location is set or updated (through `Event::create()`, `Event::updateMajorInfo()`, `Event::updateLocation()`) its audience type can automatically be updated to `education` if the new location is a specific id for CultuurKuur schools (checked via `isDummyPlaceForEducation()` on the `LocationId`). These events' audienceType can then not be updated to anything else than `education` as long as they have a location id specific to CultuurKuur. If an update is attempted the aggregate throws `IncompatibleAudienceType`, which should be caught by the new RequestHandler and converted to an API problem. This API problem might need a new error type specific to UDB since this is more of a domain error than a JSON schema error. The docs should probably also mention this (see https://github.com/cultuurnet/stoplight-docs-uitdatabank/pull/54).

The same logic for automatically setting the audienceType to `education` is also applied in the CdbXml imports but not inside the Event aggregate but inside `CultuurNet\UDB3\UDB2\Event\EventImporter::create()` at the very bottom). The same check when trying to update it afterwards is also done (in the aggregate).


composer.json
```diff
@@ -58,7 +58,7 @@
     "predis/predis": "~1.0",
     "psr/http-server-middleware": "^1.0",
     "psr/log": "^1.0",
-    "publiq/stoplight-docs-uitdatabank": "dev-unreleased",
+    "publiq/stoplight-docs-uitdatabank": "dev-III-4261-update-audience",
```
jdrieghe, line 61: Should this be changed again?

jdrieghe (approved):

PR #802, opened 2021-11-08
III-4284 Refactor get organizer

LucWollants:
### Added
- Added new `GetOrganizerRequestHandler`

### Removed
- Removed obsolete `ReadOrganizerRestController`

---
Ticket: https://jira.uitdatabank.be/browse/III-4284


src/Http/Organizer/GetOrganizerRequestHandler.php
```diff
@@ -0,0 +1,42 @@
...
+        try {
+            $organizer = $this->organizerService->getEntity($organizerId);
+
+            return (new JsonLdResponse())->withBody(
+                (new StreamFactory())->createStream($organizer)
+            );
+        } catch (EntityNotFoundException $exception) {
+        }
+
+        throw ApiProblem::organizerNotFound($organizerId);
```
jdrieghe, line 40: Should this live inside the `catch`?

src/Http/Organizer/GetOrganizerRequestHandler.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $routeParameters = new RouteParameters($request);
+        $organizerId = $routeParameters->getOrganizerId();
+
+        try {
+            $organizer = $this->organizerService->getEntity($organizerId);
+
+            return (new JsonLdResponse())->withBody(
+                (new StreamFactory())->createStream($organizer)
+            );
```
jdrieghe, line 36: I wonder if there's a less convoluted way to construct this. Maybe we should build custom tooling for it?

jdrieghe (approved):
Some questions, nothing blocking. Nice work @LucWollants

src/Http/Organizer/GetOrganizerRequestHandler.php
LucWollants: I changed the improved the `JsonResponse` constructor and it now only does a `json_encode` on types different from `string`.

jdrieghe (approved):

PR #803, opened 2021-11-09
III-4352 Add recommendationFor property to JSONLD projection of offers.

LucWollants:
### Added
- Add new value objects for `Recommendation` and `Recommendations`
- Add new repo `DBALRecommendationsRepository`
- Add new enricher `RecommendationForEnrichedOfferRepository`

### Fixed
- Fixed the recommendations table by removing the primary key and adding an index

---
Ticket: https://jira.uitdatabank.be/browse/III-4352


src/Event/Recommendations/DBALRecommendationRepository.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Event\Recommendations;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\FetchMode;
+
+final class DBALRecommendationRepository implements RecommendationsRepository
```
jdrieghe, line 10: There's a naming discrepancy here between `Recommendation` and `Recommendations`. I suggest we choose the singular version for the interface as well.

app/Migrations/Version20211109143216.php
```diff
@@ -0,0 +1,21 @@
...
+namespace CultuurNet\UDB3\Silex\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+class Version20211109143216 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $schema->getTable('event_recommendations')->addIndex(['event_id']);
```
jdrieghe, line 14: Wasn't it possible to add this line to the migration script that drops the primary key?

src/Event/Recommendations/DBALRecommendationRepository.php
LucWollants: Nicely spotted! I went for a consistent plural naming because recommendations are always a collection.

app/Migrations/Version20211109143216.php
LucWollants: Unfortunate a drop of a primary key needs one SQL command in one transaction. Also see comment here https://github.com/cultuurnet/udb3-silex/pull/803/commits/23b0ba01ecb09a66440f407c76be39a87e6bff34

src/Event/Recommendations/RecommendationsForEnrichedOfferRepository.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $recommendations = $this->recommendationsRepository->getByRecommendedEvent($jsonDocument->getId());
+
+        if ($recommendations->isEmpty()) {
+            return $jsonDocument;
+        }
+
+        return $jsonDocument->applyAssoc(
+            function (array $body) use ($recommendations) {
+                $body['metadata']['recommendationFor'] = $this->recommendationsToArray($recommendations);
```
jdrieghe, line 49: Should this also become `recommendationsFor` then?

LucWollants: Ok, now I'm also getting a bit confused 😄

But I think the rule is:
- Recommendations is always plural
- RecommendationFor is always singular

The code should reflect this now.

jdrieghe (approved):

src/Event/Recommendations/RecommendationsForEnrichedOfferRepository.php
LucWollants: With an example:
- Event A has as **recommendations** event 1 and event 2
- Event B has as **recommendations** event 1 and event 3
- Event 1 is a **recommendationFor** event A and event B

PR #805, opened 2021-11-15
Feature/iii 4346 polyfill for contenturl

JonasVHG:
### Added

- Extend polyfil for contentURL
---
Ticket: https://jira.uitdatabank.be/browse/III-4346

### DIFF
--- a/app/Media/MediaServiceProvider.php
+++ b/app/Media/MediaServiceProvider.php
@@ -102,9 +102,9 @@ class MediaServiceProvider implements ServiceProviderInterface
             }
         );
 
-        $app['image_url_repository'] = $app->share(
+        $app['media_url_repository'] = $app->share(
             function (Application $app) {
-                return new MediaUrlRepository($app['config']['media']['media_url_polyfill']);
+                return new MediaUrlRepository($app['config']['media']['media_url_mapping']);
             }
         );
     }
--- a/app/Media/MediaServiceProvider.php
+++ b/app/Media/MediaServiceProvider.php
@@ -7,6 +7,7 @@ namespace CultuurNet\UDB3\Silex\Media;
 use Broadway\UuidGenerator\Rfc4122\Version4Generator;
 use CultuurNet\UDB3\Iri\CallableIriGenerator;
 use CultuurNet\UDB3\Media\ImageUploaderService;
+use CultuurNet\UDB3\Media\MediaUrlRepository;
 use CultuurNet\UDB3\Media\MediaManager;
 use CultuurNet\UDB3\Media\MediaObjectRepository;
 use CultuurNet\UDB3\Media\Serialization\MediaObjectSerializer;
@@ -98,6 +99,12 @@ class MediaServiceProvider implements ServiceProviderInterface
                 $mediaManager->setLogger(LoggerFactory::create($app, LoggerName::forService('media', 'manager')));
 
                 return $mediaManager;
+            }
+        );
+
+        $app['image_url_repository'] = $app->share(
+            function (Application $app) {
+                return new MediaUrlRepository($app['config']['media']['media_url_polyfill']);
             }
         );
     }
--- a/config.yml.dist
+++ b/config.yml.dist
@@ -98,15 +98,15 @@ media:
     fr: Droits d'auteur gérés par %s
     de: Urheberrecht gehandhabt von %s
     en: Copyright handled by %s
-  media_url_polyfill:
+  media_url_mapping:
     udb2:
       enabled: true
       legacy_url: https://media.uitdatabank.be/
-      new_url: https://images-dev.uitdatabank.be/
+      url: https://images-dev.uitdatabank.be/
     udb3:
       enabled: true
       legacy_url: https://io.uitdatabank.dev/images/
-      new_url: https://images-dev.uitdatabank.be/
+      url: https://images-dev.uitdatabank.be/
       
 cdbxml_proxy:
   enabled: "true"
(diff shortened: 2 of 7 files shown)


src/Media/MediaUrlRepository.php
```diff
@@ -0,0 +1,25 @@
...
+final class MediaUrlRepository
+{
+    private array $mapping;
+
+    public function __construct(array $mapping)
+    {
+        $this->mapping = $mapping;
+    }
+
+    public function updateUrl(string $oldUrl): string
```
LucWollants, line 16: This is not doing a real update, it returns an updated URL. So I think a better name is `getUpdatedUrl`

config.yml.dist
```diff
@@ -98,7 +98,16 @@ media:
     fr: Droits d'auteur gérés par %s
     de: Urheberrecht gehandhabt von %s
     en: Copyright handled by %s
-
+  media_url_polyfill:
```
LucWollants, line 101: I think a better name is `media_url_mapping`

config.yml.dist
```diff
@@ -98,7 +98,16 @@ media:
...
+    udb2:
+      enabled: true
+      legacy_url: https://media.uitdatabank.be/
+      new_url: https://images-dev.uitdatabank.be/
```
LucWollants, line 105: You could just make it `url`, because of `legacy_url` it's clear that this is the new one.

src/Media/MediaUrlRepository.php
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Media;
+
+final class MediaUrlRepository
+{
+    private array $mapping;
+
+    public function __construct(array $mapping)
```
LucWollants, line 11: I would make this name a bit more explicit and call it `mediaUrlMapping`

src/Media/MediaUrlRepository.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        $this->mapping = $mapping;
+    }
+
+    public function updateUrl(string $oldUrl): string
+    {
+        foreach ($this->mapping as $udb_variant => $settings) {
```
LucWollants, line 18: The name of variables always follows the camelCasing convention and does not use underscores.

LucWollants: If you use the name `mediaUrlMappings` as the private property then `mediaUrlMapping` is also a better and more meaningful name then `settings`

app/Media/MediaServiceProvider.php
```diff
@@ -100,6 +101,12 @@ function (Application $app) {
                 return $mediaManager;
             }
         );
+
+        $app['image_url_repository'] = $app->share(
```
LucWollants, line 105: Mixing image and media here, I think it should be `image_url_repository`

tests/Offer/ReadModel/JSONLD/MediaUrlOfferRepositoryDecoratorTest.php
```diff
@@ -0,0 +1,133 @@
...
+        'image' => 'https://images.uitdatabank.be/da02d848-eb11-4bfa-a566-d8bd3b856990.jpeg',
+    ];
+
+    /**
+     * @test
+     */
+    public function it_should_replace_legacy_urls_when_enabled()
+    {
+        $id = '5624b810-c340-40a4-8f38-0393eca59bfe';
+        $mapping = [
```
LucWollants, line 86: I would make this replace both UDB2 and UDB3 media urls, now it does not test it handles multiple replacements.

tests/Offer/ReadModel/JSONLD/MediaUrlOfferRepositoryDecoratorTest.php
```diff
@@ -0,0 +1,133 @@
...
+            'udb3' => [
+                'enabled' => true,
+                'legacy_url' => 'https://io.uitdatabank.be/images/',
+                'new_url' => 'https://images.uitdatabank.be/',
+            ],
+        ];
+
+        $mediaUrlOfferRepositoryDecoratorDecorator = new MediaUrlOfferRepositoryDecorator(
```
LucWollants, line 94: This variable name has twice Decorator

src/Offer/ReadModel/JSONLD/MediaUrlOfferRepositoryDecorator.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function fetch(string $id, bool $includeMetadata = false): JsonDocument
+    {
+        $document = parent::fetch($id, $includeMetadata);
+        $document = $document->applyAssoc(
+            function (array $json) {
+                if (
+                    !isset($json['mediaObject'], $json['image']) ||
+                    !is_array($json['mediaObject']) ||
+                    !is_string($json['image'])
```
LucWollants, line 30: I would move this check `!is_string($json['image'])` to line 52.

tests/Offer/ReadModel/JSONLD/MediaUrlOfferRepositoryDecoratorTest.php
JonasVHG: Great suggestion. It made me find a bug in the mapping

LucWollants:
@JonasVHG I did a few extra minor tweaks. It's all good now. But best to only merge this once the config files are updated to avoid a crash.

LucWollants (approved):

PR #806, opened 2021-11-15
III-4227 Add endpoint to update available from

LucWollants:
### Added
- Added new `UpdateAvailableFrom` command
- Added new `AvailableFromUpdated` event
- Added new `AvailableFromDenormalizer` denormalizer
- Added new `UpdateAvailableFromRequestHandler` HTTP request handler
- Added new `UpdateAvailableFromHandler` command handler

### Changed
- Updated `Offer` aggregate to handle `UpdateAvailableFrom`
- Updated `OfferLDProjector` JSON-LD projector to handle `AvailableFromUpdated`

---
Ticket: https://jira.uitdatabank.be/browse/III-4227

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
     "predis/predis": "~1.0",
     "psr/http-server-middleware": "^1.0",
     "psr/log": "^1.0",
-    "publiq/stoplight-docs-uitdatabank": "dev-unreleased",
+    "publiq/stoplight-docs-uitdatabank": "dev-feature/III-4227-available-from",
     "rase/socket.io-emitter": "0.6.1",
     "respect/validation": "~1.1",
     "sentry/sdk": "^2.2",
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
     "predis/predis": "~1.0",
     "psr/http-server-middleware": "^1.0",
     "psr/log": "^1.0",
-    "publiq/stoplight-docs-uitdatabank": "dev-feature/III-4227-available-from",
+    "publiq/stoplight-docs-uitdatabank": "dev-unreleased",
     "rase/socket.io-emitter": "0.6.1",
     "respect/validation": "~1.1",
     "sentry/sdk": "^2.2",
--- a/src/Offer/Offer.php
+++ b/src/Offer/Offer.php
@@ -732,7 +732,7 @@ abstract class Offer extends EventSourcedAggregateRoot implements LabelAwareAggr
         }
 
         // It is required to use `==` instead of `===` to compare DateTime objects in PHP
-        if ($this->availableFrom !== null && $this->availableFrom == $availableFrom) {
+        if ($this->availableFrom == $availableFrom) {
             return;
         }
 
--- a/src/Offer/Offer.php
+++ b/src/Offer/Offer.php
@@ -68,6 +68,7 @@ use CultuurNet\UDB3\Offer\Events\Moderation\AbstractRejected;
 use CultuurNet\UDB3\Offer\ValueObjects\BookingAvailability;
 use CultuurNet\UDB3\PriceInfo\PriceInfo;
 use CultuurNet\UDB3\Title;
+use DateTimeImmutable;
 use DateTimeInterface;
 use Exception;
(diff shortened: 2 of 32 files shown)


src/Offer/Offer.php
```diff
@@ -722,6 +725,20 @@ protected function importWorkflowStatus(CultureFeed_Cdb_Item_Base $cdbItem): voi
...
     }
 
+    public function updateAvailableFrom(DateTimeInterface $availableFrom): void
+    {
+        if ($availableFrom < new DateTimeImmutable()) {
+            $availableFrom = new DateTimeImmutable();
+        }
+
+        // It is required to use `==` instead of `===` to compare DateTime objects in PHP
+        if ($this->availableFrom !== null && $this->availableFrom == $availableFrom) {
```
jdrieghe, line 735: Why do we still need the null check if we're already comparing with a `DateTimeInterface` instance?

jdrieghe:
```suggestion
        if ($this->availableFrom == $availableFrom) {
```

src/Offer/Offer.php
```diff
@@ -720,6 +725,20 @@ protected function importWorkflowStatus(CultureFeed_Cdb_Item_Base $cdbItem): voi
         $this->workflowStatus = $workflowStatus;
     }
 
+    public function updateAvailableFrom(DateTimeInterface $availableFrom): void
+    {
+        if ($availableFrom < new DateTimeImmutable()) {
+            $availableFrom = new DateTimeImmutable();
```
jdrieghe, line 731: As far as I can tell, there is no test to verify this behaviour.

composer.json
```diff
@@ -58,7 +58,7 @@
     "predis/predis": "~1.0",
     "psr/http-server-middleware": "^1.0",
     "psr/log": "^1.0",
-    "publiq/stoplight-docs-uitdatabank": "dev-unreleased",
+    "publiq/stoplight-docs-uitdatabank": "dev-feature/III-4227-available-from",
```
jdrieghe, line 61: Reminder to change this again before merging this PR.

src/Offer/Offer.php
LucWollants: You are right! Although the internal property `availableFrom` can start as `null`, the incoming parameter `availableFrom` can never be `null` because of the `DateTimeInterface` type hint.

src/Offer/Offer.php
LucWollants: Indeed no test for this because the scenario's work with an `assertEquals` and we don't have a way to mock the `now`. I will try to fix this...

LucWollants: I found a way to test this by storing the traceable event store and comparing the recorded event manually and making sure the time difference is small.

tests/Offer/CommandHandlers/UpdateAvailableFromHandlerTest.php
```diff
@@ -103,4 +108,45 @@ public function updateAvailableFromDataProvider(): array
...
+        $this->assertCount(1, $events);
+        $this->assertInstanceOf(AvailableFromUpdated::class, $events[0]);
+
+        /** @var AvailableFromUpdated $availableFromUpdated */
+        $availableFromUpdated = $events[0];
+        $this->assertEquals($eventId, $availableFromUpdated->getItemId());
+
+        $nowTimestamp = (new DateTimeImmutable())->getTimestamp();
+        $availableFromTimeStamp = $availableFromUpdated->getAvailableFrom()->getTimestamp();
+        $this->assertTrue($nowTimestamp - $availableFromTimeStamp < 1);
```
jdrieghe, line 150: You could use the `->diff()` method on `DateTimeImmutable` to get a `DateInterval` object to do the assertions, but I guess this works as well 👍

jdrieghe (approved):

tests/Offer/CommandHandlers/UpdateAvailableFromHandlerTest.php
LucWollants: True, but then I still had to interpret the `DateInterval` and convert that to a total amount in seconds.

PR #807, opened 2021-11-15
III-4227 Boy scouting

LucWollants:
### Changed
- Improved type hinting on `OfferLDProjector`, `EventLDProjector` and `PlaceLDProjector`
- Improved type hinting on `OfferLDProjectorTest`, `EventLDProjectorTest` and `PlaceLDProjectorTest`
- Used `Json::encode` which throws by default

---
Ticket: https://jira.uitdatabank.be/browse/III-4227

### DIFF
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -145,6 +145,9 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
         }
     }
 
+    /**
+     * @return array<string,string>
+     */
     private function getEventHandlers(): array
     {
         $events = [];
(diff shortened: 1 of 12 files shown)


src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -148,11 +145,7 @@ public function handle(DomainMessage $domainMessage)
         }
     }
 
-    /**
-     * @return string[]
-     *   An associative array of commands and their handler methods.
-     */
-    private function getEventHandlers()
+    private function getEventHandlers(): array
```
jdrieghe, line 148: The `string[]` doc block here was more specific, so maybe we should keep that?
I think we can be even more specific and mark it as `array<string,string>` in the doc block.

jdrieghe (approved):

jdrieghe (approved):

PR #824, opened 2021-12-02
III-4410 Create repository for news article table

LucWollants:
### Added
- Added value object for `NewsArticle` and a simple `NewsArticles` collection
- Added repository `NewsArticleRepository` and a concrete DBAL implementation `DBALNewsArticleRepository`

---
Ticket: https://jira.uitdatabank.be/browse/III-4410

### DIFF
--- a/src/Curators/DBALNewsArticleRepository.php
+++ b/src/Curators/DBALNewsArticleRepository.php
@@ -84,6 +84,17 @@ final class DBALNewsArticleRepository implements NewsArticleRepository
             ->execute();
     }
 
+    public function delete(UUID $id): void
+    {
+        $queryBuilder = $this->connection->createQueryBuilder();
+
+        $queryBuilder
+            ->delete('news_article')
+            ->where('id = :id')
+            ->setParameter(':id', $id->toString())
+            ->execute();
+    }
+
     private function createNewsArticle(array $newsArticleRow): NewsArticle
     {
         return new NewsArticle(
--- a/src/Curators/DBALNewsArticleRepository.php
+++ b/src/Curators/DBALNewsArticleRepository.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Curators;
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\Model\ValueObject\Translation\Language;
+use CultuurNet\UDB3\Model\ValueObject\Web\Url;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\FetchMode;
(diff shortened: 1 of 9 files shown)


src/Curators/NewsArticleNotFoundException.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Curators;
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use Exception;
+
+final class NewsArticleNotFoundException extends Exception
```
jdrieghe, line 10: Is there a reason why we still postfix exception classes with `...Exception`?
I think `NewsArticleNotFound()` is clear enough.

src/Curators/DBALNewsArticleRepository.php
```diff
@@ -51,6 +51,39 @@ public function getAll(): NewsArticles
         return $this->createNewsArticles($newsArticleRows);
     }
 
+    public function create(NewsArticle $newsArticle): void
+    {
+        $queryBuilder = $this->connection->createQueryBuilder();
```
jdrieghe, line 56: Isn't there an `->insert()` method on DBAL connections that can make this implementation a lot shorter?

src/Curators/DBALNewsArticleRepository.php
```diff
@@ -84,6 +84,17 @@ public function create(NewsArticle $newsArticle): void
             ->execute();
     }
 
+    public function delete(UUID $id): void
+    {
+        $queryBuilder = $this->connection->createQueryBuilder();
```
jdrieghe, line 118: Same here, I think there is a `->delete()` shorthand

src/Curators/NewsArticleNotFoundException.php
LucWollants: Good point, I changed it to `NewsArticleNotFound`

src/Curators/DBALNewsArticleRepository.php
LucWollants: There is indeed a shorthand `insert` on the `Connection` class but that is not escaped and are not safe for user input. So I always stick with the `QueryBuilder` methods and only use the shorthand ones inside tests.

See https://github.com/doctrine/dbal/blob/v2.9.3/lib/Doctrine/DBAL/Connection.php#L745

src/Curators/DBALNewsArticleRepository.php
LucWollants: Same issue with the `delete` method of `Connection` https://github.com/doctrine/dbal/blob/v2.9.3/lib/Doctrine/DBAL/Connection.php#L636

PR #826, opened 2021-12-02
III-4391 refactor OrganizerCreated Title

JonasVHG:
### Changed

- refactor `OrganizerCreated` to use primitives for Title

---
Ticket: https://jira.uitdatabank.be/browse/III-4391

### DIFF
--- a/src/Organizer/Events/OrganizerCreated.php
+++ b/src/Organizer/Events/OrganizerCreated.php
@@ -24,17 +24,17 @@ final class OrganizerCreated extends OrganizerEvent
     /**
      * @var string[]
      */
-    public $phones;
+    public array $phones;
 
     /**
      * @var string[]
      */
-    public $emails;
+    public array $emails;
 
     /**
      * @var string[]
      */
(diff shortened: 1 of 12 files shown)


JonasVHG:
The Issue III-4387 talks about also removing
CultuurNet\UDB3\Organizer\Events\OrganizerCreated

> CultuurNet\UDB3\Address\Address -> refactor to string $countryCode, string $locality, string postalCode, string $streetAddress properties
> CultuurNet\UDB3\Title -> refactor to string $title property
But since the constructor expects an array of addresses instead of a single address it might be good to decide what to do with this, before continuing.
https://github.com/cultuurnet/udb3-silex/blob/master/src/Organizer/OrganizerLDProjector.php#L169 makes it look like Organizers with multiple addresses are not useful.

LucWollants:
Lets focus only on `Title` for this PR and create a follow-up PR that only accepts one address.

LucWollants (approved):
Nice cleanup again

PR #850, opened 2021-12-10
III-4410 Add pagination

LucWollants:
### Added
- Added basic pagination with `page` query parameter
- Wrapped result inside `hydra:member` to mimic old API

---
Ticket: https://jira.uitdatabank.be/browse/III-4410

### DIFF
--- a/src/Curators/DBALNewsArticleRepository.php
+++ b/src/Curators/DBALNewsArticleRepository.php
@@ -38,19 +38,6 @@ final class DBALNewsArticleRepository implements NewsArticleRepository
         return $this->createNewsArticle($newsArticleRows[0]);
     }
 
-    public function getAll(): NewsArticles
-    {
-        $queryBuilder = $this->connection->createQueryBuilder();
-
-        $newsArticleRows = $queryBuilder
-            ->select('*')
-            ->from('news_article')
-            ->execute()
-            ->fetchAll(FetchMode::ASSOCIATIVE);
-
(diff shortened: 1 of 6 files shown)


src/Curators/DBALNewsArticleRepository.php
```diff
@@ -77,7 +64,14 @@ public function search(NewsArticleSearch $newsArticleSearch): NewsArticles
                 ->setParameter(':url', $newsArticleSearch->getUrl());
         }
 
+        if ($newsArticleSearch->getStartPage() > 1) {
+            $query = $query->setFirstResult(
```
jdrieghe, line 68: Is this zero-based? Should we add 1 if it's not?

LucWollants: It is 1-based. So I added a check inside the mutator method https://github.com/cultuurnet/udb3-silex/pull/850/files#diff-16b9e8bcb01691eaad8afdce848124a1d0cae3fd8d5350f674a637dbdddb9f66R29

LucWollants: I did understand the question wrong. The call `setFirstResult` is zero based but the pagination is one based.

For example: 2 items per page and page 3 is asked, this means that we need item (2 x 3) + 1 = 7. But because the DBAL offset is zero based we need item 6, so we can do 2 x 3.

jdrieghe (approved):

PR #854, opened 2021-12-13
III-4187 refactor enum Format

JonasVHG:
### Changed
 -  Changed `Format` to use new `Enum` class instead of deprecated one.
---


LucWollants (approved):
Good work! Make sure to read the extra info PHPStorm provides when doing the boy scouting.

PR #859, opened 2021-12-14
III-4187 refactor enum Privacy

JonasVHG:
### Changed

- Changed `Enum` of `Privacy`

---
Ticket: https://jira.uitdatabank.be/browse/III-4189


LucWollants:
@JonasVHG I'm gonna close this pull request. I had a closer look at the code and we should replace the `Label` class this is used in. So in the end we would end up having to do 2 refactors.

PR #864, opened 2021-12-23
III-4429 Patch main image on organizer

LucWollants:
### Added
- Added `MainImageUpdated` event
- Added `UpdateOrganizer` command (for now this only supports setting the main image)
- Added `UpdateOrganizerHandler` for the `UpdateOrganizer` command
- Added `UpdateOrganizerRequestHandler` to handle the PATCH request on organizers
- Added the method `updateOrganizer` on `Organizer` aggregate. This method will dispatch several dedicated events, but for now only `MainImageUpdated`
- Added the method `applyMainImageUpdated` on the `OrganizerLDProjector`

### Changed
- Adding an image to an organizer updates the organizer projection with a main image
- Removing the last image from an organizer removes the main image from the organizer projection
- Removing an image from an organizer which is also the main image resets the main image on the organizer projection 

---
Ticket: https://jira.uitdatabank.be/browse/III-4429


jdrieghe (approved):
👍

PR #865, opened 2022-01-10
IIII-4187 refactor enum JobStatus

JonasVHG:
### Changed

- Changed LegacyEnum for `JobStatus`

---
Ticket: https://jira.uitdatabank.be/browse/IIII-4187

### DIFF
--- a/src/Http/Jobs/JobStatus.php
+++ b/src/Http/Jobs/JobStatus.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Http\Jobs;
 
-use ValueObjects\Enum\Enum;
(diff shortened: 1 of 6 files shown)


src/Http/Jobs/JobStatus.php
```diff
@@ -4,20 +4,37 @@
...
+    {
+        return [
+            'waiting',
+            'running',
+            'failed',
+            'complete',
+        ];
+    }
+
+    public static function WAITING(): JobStatus
```
LucWollants, line 21: I would prefer to use the lower case function names here. Then we follow the naming convention of functions.

LucWollants (changes requested):
I would change the function casing. I also did a few extra boy scouting changes, so you will need to do a git pull first.

LucWollants (approved):

PR #866, opened 2022-01-10
III-4187 refactor enum EntityType

JonasVHG:
### Changed

- Refactored `EntityType` to stop using deprecated Enum

---

### DIFF
--- a/src/Http/JSONLD/EntityType.php
+++ b/src/Http/JSONLD/EntityType.php
@@ -17,24 +17,4 @@ class EntityType extends Enum
             'postaladdress',
         ];
     }
-
-    public static function EVENT(): EntityType
-    {
-        return new self('event');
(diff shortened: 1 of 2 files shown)


src/Http/JSONLD/EntityType.php
```diff
@@ -4,18 +4,37 @@
...
+    {
+        return [
+            'event',
+            'place',
+            'organizer',
+            'postaladdress',
+        ];
+    }
+
+    public static function EVENT(): EntityType
```
LucWollants, line 21: Same remark here, use the regular method casing.

LucWollants (changes requested):
Please also use the regular method name casing. I also did some extra boy scouting.

LucWollants (approved):

PR #867, opened 2022-01-10
III-4187 refactor enum RelationType

JonasVHG:
### Changed

- Refactored `RelationType` to replace deprecated Enum

---


LucWollants (changes requested):
Same request to use the conventional lower case method naming 😉

PR #870, opened 2022-01-11
III-4510 Refactor update organizer to update main image

LucWollants:
### Changed
- Refactored update organizer to update main image

---
Ticket: https://jira.uitdatabank.be/browse/III-4510


src/Organizer/Organizer.php
```diff
@@ -363,7 +324,35 @@ public function updateImage(
...
+
+        // If the organizer had no main image just set it as main.
+        // Although this should be a race condition.
+        if ($this->mainImageId === null) {
+            return true;
+        }
+
+        // Only set it as main when there is a difference.
+        // This is to prevent having events in the event store with no change.
+        return !$this->mainImageId->sameAs($mainImageId);
```
bertramakers, line 352: Nit: If we could store `$this->mainImageId` as `string|null` instead of `UUID|null` (in `applyImageAdded()`), this whole method could just be `return $this->hasImage($mainImageId) && $this->mainImageId !== $mainImageId;`

I'm a fan of the extra readability of these multiple steps, but one is not really necessary (only because `$this->mainImageId` is `UUID` and not just `string`), and these aggregates grow big very quickly so methods like this make them even bigger.

But functionally it works as intended and it is readable so just a nit!

bertramakers (approved):
Looks great, good work! ✅

src/Organizer/Organizer.php
LucWollants: Good remark! I did tweak the usage of `mainImageId` a bit.

bertramakers:
Perfect!

PR #871, opened 2022-01-11
III-4508 Make it possible to return either JSON or JSON-LD on /news-articles endpoints

bertramakers:
### Changed

- Updated `/news-articles` endpoints to support both JSON and JSON-LD as response content-type

---
Ticket: https://jira.uitdatabank.be/browse/III-4508

Note: This ticket also requires extra acceptance tests and updated API documentation, which will be done in other PRs


src/Curators/Serializer/NewsArticleNormalizer.php
```diff
@@ -9,15 +9,31 @@
 
 final class NewsArticleNormalizer implements NormalizerInterface
 {
+    private $jsonLd = false;
```
LucWollants, line 12: This property can get the `boolean` type.

src/Curators/Serializer/NewsArticleNormalizer.php
```diff
@@ -9,15 +9,31 @@
...
+
+    public function withJsonLd(bool $jsonLd): self
```
LucWollants, line 14: I think we only want to turn jsonLd "on", so maybe we can call this `asJsonLd`?

src/Http/Curators/CreateNewsArticleRequestHandler.php
```diff
@@ -48,9 +51,15 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
 
-        return new JsonResponse(
-            ['id' => $newsArticle->getId()->toString()],
-            StatusCodeInterface::STATUS_CREATED
-        );
+        $headers = new Headers($request);
+        $responseContentType = $headers->determineResponseContentType(['application/ld+json', 'application/json']);
+        $withJsonLd = $responseContentType === 'application/ld+json';
+
+        $normalized = (new NewsArticleNormalizer())->withJsonLd($withJsonLd)->normalize($newsArticle);
```
LucWollants, line 58: Maybe the construction of the `NewsArticleNormalizer` can move to a very simple factory class so we avoid some duplication?

src/Http/Request/Headers.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace CultuurNet\UDB3\Http\Request;
+
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use Negotiation\Accept;
+use Negotiation\Exception\Exception as NegotiationException;
+use Negotiation\Negotiator;
+use Psr\Http\Message\ServerRequestInterface;
+
+final class Headers
```
LucWollants, line 13: Nice abstraction!

src/Http/Request/Headers.php
```diff
@@ -0,0 +1,57 @@
...
+            return $possibleContentTypes[0];
+        }
+
+        $negotiator = new Negotiator();
+
+        $notAcceptable = ApiProblem::notAcceptable('Acceptable media types are: ' . implode(', ', $possibleContentTypes));
+
+        try {
+            /** @var Accept|null $mediaType */
+            $mediaType = $negotiator->getBest($acceptHeader, $possibleContentTypes);
```
LucWollants, line 46: Lately I am sometimes hesitant to add new dependencies, would this be something we could implement?

LucWollants (approved):
Nice work. I already approved because the remarks are minor.

src/Curators/Serializer/NewsArticleNormalizer.php
bertramakers: No real preference here but I'll change it!

src/Curators/Serializer/NewsArticleNormalizer.php
bertramakers: Right!

src/Http/Curators/CreateNewsArticleRequestHandler.php
bertramakers: Yes, I was also thinking how I could reduce the duplication yesterday but didn't find a good way because we also need the `$withJsonLd` boolean to determine the right `Response` class to return (`JsonResponse` or `JsonLdResponse`). But I think I found a good solution, I'll try it out.

src/Http/Request/Headers.php
bertramakers: I started off by implementing it myself but it has to follow the [rfc7231 spec for content negotiation](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3), which I'm no expert in. I agree about not adding too many dependencies, but this one seems pretty standardized (because of the spec it's based on) and it's abstracted away via the custom `Headers` class, so if we need to replace it sometime in the future it should be easy to just change the usages in that one custom class? (It's not like it would be spread out all over the codebase)

bertramakers: To give a complex example from the spec that's not trivial to implement ourselves:

```
Accept: text/plain; q=0.5, text/html,
             text/x-dvi; q=0.8, text/x-c
```

We don't need these elaborate features like the `q` right now, but an API client should be able to expect that we can at least parse it without crashing. Which I could not guarantee with a custom implementation right now, because there's so many things to take into account with this header.

bertramakers: Oh and if it helps, I picked this package because it has no dependencies on its own. So it's literally "just" one extra package. 😄

LucWollants: Ok, no waterfall of extra packages 😀

LucWollants (approved):
Nice improvements!

PR #872, opened 2022-01-12
III-4426 Make properties when adding an image on an organizer optional

LucWollants:
### Changed
- Made properties when adding an image on an organizer optional

---
Note: I left the `composer.json` conflict so I can set it do dev-unreleased when reviewed.

Ticket: https://jira.uitdatabank.be/browse/III-4426


bertramakers (approved):
Good improvement! 👍

PR #873, opened 2022-01-12
III-4509 Block new news article with existing url and about

bertramakers:
### Added

- Added check that no news article(s) already exist with the same `url` and `about` when creating a new news article

---
Ticket: https://jira.uitdatabank.be/browse/III-4509

### DIFF
--- a/src/Http/Curators/CreateNewsArticleRequestHandler.php
+++ b/src/Http/Curators/CreateNewsArticleRequestHandler.php
@@ -7,7 +7,9 @@ namespace CultuurNet\UDB3\Http\Curators;
 use Broadway\UuidGenerator\UuidGeneratorInterface;
 use CultuurNet\UDB3\Curators\NewsArticle;
 use CultuurNet\UDB3\Curators\NewsArticleRepository;
+use CultuurNet\UDB3\Curators\NewsArticleSearch;
 use CultuurNet\UDB3\Curators\Serializer\NewsArticleDenormalizer;
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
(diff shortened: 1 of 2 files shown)


src/Http/Curators/CreateNewsArticleRequestHandler.php
```diff
@@ -45,6 +47,26 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+        );
+
+        if (!$existingNewsArticles->isEmpty()) {
+            /** @var NewsArticle $first */
+            $first = $existingNewsArticles->getFirst();
+            $id = $first->getId()->toString();
+
+            throw ApiProblem::bodyInvalidDataWithDetail(
+                'A news article with the given url and about already exists. (' . $id . ') '
+                    . 'Do a GET /news-articles request with `url` and `about` parameters to find it programmatically.'
```
LucWollants, line 66: Is it not possible to get the news article with the `id`?

LucWollants (approved):
Looks good.

src/Http/Curators/CreateNewsArticleRequestHandler.php
bertramakers: Yes but this is to explain how a client can check themselves if an article with their `url` and `about` already exists _before_ doing a `POST` request. But it may not be that clear then 😄

LucWollants: Oh! I understand now 🙂

bertramakers: I'm going to leave it like this even if it's a bit confusing because it's super minor I think and updating it would require changes to this + unit tests + acceptance tests.

PR #874, opened 2022-01-13
III-4187 refactor enum permission

JonasVHG:
### Changed

- Refactor `Permission` to stop using deprecated Enum

---


src/Http/Management/UserPermissionsVoter.php
```diff
@@ -69,7 +66,7 @@ public function vote(TokenInterface $token, $object, array $attributes)
                 continue;
             }
 
-            $permission = Permission::fromNative($attribute);
+            $permission = $attribute;
 
             if (!in_array($permission, $userPermissions)) {
```
LucWollants, line 71: I think the type of `$permission` is `Permission` and the `$userPermissions` is array of `Permission`. So I think `in_array` will always be `false`?

src/Http/Role/EditRoleRestController.php
```diff
@@ -115,13 +115,13 @@ public function addPermission(string $roleId, string $permissionKey): Response
             throw new InvalidArgumentException('Required field roleId is missing');
         }
 
-        if (!in_array($permissionKey, array_keys(Permission::getConstants()))) {
+        if (!array_key_exists($permissionKey, Permission::getAllowedValues())) {
```
LucWollants, line 118: I think this array `Permission::getAllowedValues()` is not associative and has integer values as keys. So `array_key_exists` maybe always returns `true`?

LucWollants (changes requested):
Very good work and also some nice added value with the extra boy scouting.
I only have a few doubts about some types, better to add some extra tests for those and some extra type hints in the doc blocks.

LucWollants:
I also added @bertramakers as a reviewer because this change is part of the core of command handling, so an extra pair of eyes can't harm.

src/Http/Management/UserPermissionsVoter.php
bertramakers: +1 for unit test to be sure, but I don't see the problem here immediately to be honest. `in_array` can compare objects in arrays. There's even a third argument to enable/disable strict comparisons so it does or doesn't do type casting. (It type casts by default, but that should be no issue here)

src/Role/ValueObjects/Permission.php
```diff
@@ -4,32 +4,81 @@
...
+            'Organisaties bewerken',
+            'Gebruikers beheren',
+            'Labels beheren',
+            'Voorzieningen bewerken',
+            'Producties aanmaken',
+            'Films aanmaken',
+        ];
+    }
+
+    public static function getAllowedPermissions(): array
```
bertramakers, line 27: Small suggestion, maybe a name like `getAll()` or `getAllPermissions()` is better for this method. Because the "Allowed" of "getAllowedValues()" is generic and exists on all enums and is in regards to what values are allowed on this specific enum type (fixed list of possible values). While "getAllowedPermissions()" can get confusing  because it might seem like those are the permissions allowed for a specific user or role.

src/Http/Role/EditRoleRestController.php
```diff
@@ -233,4 +233,13 @@ private function getLabelId(string $labelIdentifier): ?UUID
...
     }
+
+    /**
+     * This function was added during the refactoring of Permission.
+     * Because the API expects Uppercase/Underscore names for Permission.
+     */
+    public static function getByUppercaseName(string $key): Permission
+    {
+        return new Permission(ucfirst(str_replace('_', ' ', strtolower($key))));
+    }
```
bertramakers, line 244: Wouldn't it be easier and less error-prone to keep the old uppercased names with underscores instead? Now we have to make sure we don't overlook a place where the conversion needs to happen. And it feels confusing in the long run. Someone might accidentally use the new casing in new endpoints which would be inconsistent. Was there a specific reason to change the casing of these permission values?

bertramakers: But the fact that this is copied in various classes again makes me wonder if this is the ideal approach. Maybe it should at least be copied to `Permission` itself if we keep it.

src/Role/ReadModel/Detail/Projector.php
```diff
@@ -128,4 +128,13 @@ function ($item) use ($permissionName) {
 
         $this->repository->save($document->withBody($json));
     }
+
+    /**
+     * This function was added during the refactoring of Permission.
+     * Because the API expects Uppercase/Underscore names for Permission.
+     */
+    private static function getUppercaseName(string $value): string
```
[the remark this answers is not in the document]
bertramakers: See my other comment as well https://github.com/cultuurnet/udb3-silex/pull/874/files#r785765302
(Didn't see this one at first)

src/Http/Role/EditRoleRestController.php
LucWollants: We wanted to create a follow-up ticket to refactor the usage of the uppercase values. Because only the UI needs the uppercase values inside the admin interface to manage roles. That was also the reason to keep the method `getUppercaseName` away from the `Permission` class.

bertramakers (commented):
Overall nice work removing this dependency on the old `Enum`. Nice boyscouting with the typehints as well. 👏 

Just a suggestion, when doing refactorings like this it's always interesting to start with one `Enum` and focus on that and then make a PR for just that one. Then the next one etc. Then we can give some feedback sooner on the approach (like the casing of the values etc). It also makes reviewing easier if we can go over multiple smaller PRs instead of one big PR. It's not because it's one ticket that you can't do multiple PRs.

src/Http/Role/EditRoleRestController.php
bertramakers: Assuming that only the UI uses these uppercase values for a second (not 100% sure), it will still require changes in a lot of UI repositories:

- udb3-frontend
- udb3-angular
- udb3-angular-app
- omd-angular-app

So like I asked in my first comment, was there a specific reason to change the casing of these permission values?

Technically it's a breaking API change if we change the casing. Best case we have to update 4 repositories, but I wouldn't be surprised if some other internal tool also uses them.

bertramakers: > Because only the UI needs the uppercase values inside the admin interface to manage roles.

The UI also uses it to actually check permissions, like which links to show in the sidebar.

LucWollants: When discussing this PR with Jonas I first made the advise to make the values lowercase because those lowercase values are stored inside the database. But I think I gave some bad advise. It seems better to keep the values as uppercase values. Inside the database layer we have more control on how things are handled compared to the parts that consume the exposed API.

One question stays, if we make the values uppercase, do we also make the method names uppercase? I guess we could keep the method names lowercase because that matches the method naming convention. But the values they return are uppercase.

src/Role/ReadModel/Detail/Projector.php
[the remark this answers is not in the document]
LucWollants: The methods are now part of `Permission` and well documented.

src/Http/Role/EditRoleRestController.php
LucWollants: I think the current approach is fine:
- The `Permission` class exposes lower case values
- The database and business layer work with the normal lower case values
- The API layer uses well documented methods that do a upper case conversion

src/Http/Management/UserPermissionsVoter.php
LucWollants: I noticed that this code is already covered with a unit test. It indeed works because strict comparison is not enabled.

LucWollants (approved):

LucWollants:
I tried to run the acceptance tests, but they still fail. I will have a look why.

src/Http/Management/UserPermissionsVoter.php
bertramakers: I think it would even work with strict comparison, since it's comparing `Permission` objects on both sides

LucWollants:
> I tried to run the acceptance tests, but they still fail. I will have a look why.

It turns out that:
- the `vote` method of `UserPermissionsVoter` gets a `Permission[]` array parameter
- the `vote` method of `PermissionsVoter` gets a `string[]` array parameter

LucWollants:
I also did some manual testing and everything related to roles seems to work again.

bertramakers:
> Just a suggestion, when doing refactorings like this it's always interesting to start with one Enum and focus on that and then make a PR for just that one. Then the next one etc. Then we can give some feedback sooner on the approach (like the casing of the values etc). It also makes reviewing easier if we can go over multiple smaller PRs instead of one big PR. It's not because it's one ticket that you can't do multiple PRs.

I just noticed that I was mistaken and this PR only focuses on the `Permission` enum. So this advice is not relevant, it was already done like I suggested! My bad, sorry about that! 🙃

PR #878, opened 2022-01-17
III-4187 refactor enum WebArchiveTemplate

JonasVHG:
### Changed

- Refactor `WebArchiveTemplate` for deprecated enum

---

### DIFF
--- a/src/EventExport/Format/HTML/WebArchive/WebArchiveTemplate.php
+++ b/src/EventExport/Format/HTML/WebArchive/WebArchiveTemplate.php
@@ -4,14 +4,25 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\EventExport\Format\HTML\WebArchive;
 
-use ValueObjects\Enum\Enum;
+use CultuurNet\UDB3\Model\ValueObject\String\Enum;
 
-/**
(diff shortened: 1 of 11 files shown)


LucWollants:
As discussed over Slack the updated type hinting can be done. But some return types should be changed to nullable objects. For example use `public function getSubtitle(): ?Subtitle`. Inside the constructor of `ExportEventsAsPDF`

LucWollants (changes requested):
As discussed over Slack the updated type hinting can be done. But some return types should be changed to nullable objects. For example use `public function getSubtitle(): ?Subtitle`. Inside the constructor of `ExportEventsAsPDF`

src/EventExport/Format/HTML/WebArchive/WebArchiveTemplate.php
```diff
@@ -4,14 +4,25 @@
...
-    public const MAP = 'map';
+    public static function getAllowedValues(): array
+    {
+        return [
+            'tips',
+            'map',
+        ];
+    }
+
+    public static function tips(): WebArchiveTemplate
```
LucWollants, line 19: I think these methods can be added as doc blocks because the name of the method would match the value.

LucWollants (approved):

PR #880, opened 2022-01-20
III-4516 Refactor copy event

bertramakers:
### Changed

- Refactored `POST /events/{eventId}/copies` to use a PSR-15 request handler, use JSON schema validation, accept both the deprecated and current calendar schema, and use a new `CopyEvent` command instead of a service class.

---
Ticket: https://jira.uitdatabank.be/browse/III-4516

Note: I will add the actual copy of the production link in a new PR. I wanted to do this refactoring first because otherwise I'd have to inject the production repository into the deprecated EventEditingService and update all its usages everywhere. And the refactoring was also necessary to document it correctly on Stoplight with the right errors and schema.


tests/Event/CommandHandlers/CopyEventHandlerTest.php
```diff
@@ -0,0 +1,95 @@
...
+use CultuurNet\UDB3\Language;
+use CultuurNet\UDB3\Model\ValueObject\Calendar\OpeningHours\OpeningHours;
+use CultuurNet\UDB3\Model\ValueObject\Calendar\PermanentCalendar;
+use CultuurNet\UDB3\Timestamp;
+use CultuurNet\UDB3\Title;
+use DateTimeImmutable;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class CopyEventHandlerTest extends TestCase
```
LucWollants, line 25: I guess it was not possible to use `CommandHandlerScenarioTestCase` here because two aggregates are involved when doing a copy?

bertramakers: Yeah I tried but the problem is indeed that the Scenario expects to start from an Aggregate that exists, and then allows you to compare the new events on that aggregate to the ones you'd expect. And in this case the Aggregate that "exists" is the original event, but we want to check the events on the new Aggregate.

tests/Http/Event/CopyEventRequestHandlerTest.php
```diff
@@ -0,0 +1,889 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidFactoryInterface;
+
+class CopyEventRequestHandlerTest extends TestCase
+{
+    use AssertApiProblemTrait;
+
+    private MockObject $uuidFactory;
```
LucWollants, line 45: The properties `$uuidFactory` and `$iriGenerator` can be replaced with a local variable according PHPStorm 🙂

bertramakers: Good point. I didn't pay attention to that inspection.

tests/Http/Event/CopyEventRequestHandlerTest.php
```diff
@@ -0,0 +1,889 @@
...
+        $this->assertEquals([$expectedCommand], $this->commandBus->getRecordedCommands());
+        $this->assertEquals(201, $response->getStatusCode());
+        $this->assertEquals($expectedJson, $response->getBody()->getContents());
+    }
+
+    public function validEventDataProvider(): array
+    {
+        return [
+            'single' => [
+                'data' => (object) [
```
LucWollants, line 96: TIL it is possible to give the data itself also meaningful names and not just the scenario.

tests/Http/Event/CopyEventRequestHandlerTest.php
```diff
@@ -0,0 +1,889 @@
...
+
+    /**
+     * @test
+     * @dataProvider validEventDataProvider
+     */
+    public function it_does_not_throw_when_given_valid_event_data($data, CopyEvent $expectedCommand): void
+    {
+        $response = $this->copyEventRequestHandler->handle(
+            (new Psr7RequestBuilder())
+                ->withBodyFromString(json_encode($data))
```
LucWollants, line 80: Maybe use the "safer" method `Json::encode`?

LucWollants (approved):
Nice work with the extensive tests for the request handler.

tests/Event/CommandHandlers/CopyEventHandlerTest.php
LucWollants: Thanks for clarifying!

tests/Http/Event/CopyEventRequestHandlerTest.php
bertramakers: Yeah good point, I just copied this from `UpdateCalendarRequestHandlerTest`. I'll change it there as well.

bertramakers: I updated all usages of `withBodyFromString()` where we used `json_encode()` inside it, to use the existing `withBodyFromArray()` instead. (Was probably added a bit later when we already had these usages of `withBodyFromString()` + `json_encode()`)

bertramakers: This is now making a lot of tests fail because they use an `object` instead of `array` for the JSON data 🙃 I'll fix it.

tests/Http/Event/CopyEventRequestHandlerTest.php
bertramakers: Yeah but just by coincidence. Because the test data has to be an array, and it doesn't matter to PHPUnit if the keys are sequential numbers (normal array with only values) or strings (associative array with keys). It only uses the values. But if you make it an associative array you can add some extra meaning for easier readability.

LucWollants (approved):
Ready to merge after final improvements.

PR #881, opened 2022-01-20
III-4187 refactor enum AudienceType

JonasVHG:
### Changed

- Changed the Legacy `AudienceType` which uses a deprecated `Enum` with Model `AudienceType`

### Removed

- Removed the Legacy `AudienceType` and corresponding Test

---


tests/Event/Events/AudienceUpdatedTest.php
```diff
@@ -5,36 +5,24 @@
...
-     * @var string
-     */
-    private $itemId;
+    private string $itemId;
 
-    /**
-     * @var AudienceType
-     */
-    private $audienceType;
+    private AudienceType $audienceType;
```
LucWollants, line 15: This can become a local variable instead of a property. Or even directly passed into `new Audience` on line 27.

src/Event/ValueObjects/Audience.php
```diff
@@ -5,46 +5,42 @@
...
  */
 final class Audience implements Serializable
 {
-    /**
-     * Store the Audience enum internally as a string to make sure that PHP encode works.
-     * @var string
-     */
-    private $audienceType;
-
+    private string $audienceType;
```
LucWollants, line 16: By changing the `Enum` class it is no longer needed to store the `AudienceType` as a `string`. Our own `Enum` class does support the `serialize` method.

LucWollants:
Very nice work with the boy scouting. The file `CdbXMLImporter` can have a few more type hint and doc block improvements.

src/Event/ValueObjects/Audience.php
JonasVHG: This one wasn't really clear because `src/Event/ValueObjects/Audience.php` is also a deprecated class, but because it doesn't use `marc-mabe/php-enum` I thought I would be to much out of scope the make many changes to this class.

bertramakers: I think @LucWollants means you can change this property to the type `AudienceType` instead of `string`. Converting it to a string was a necessary workaround to make some commands that used it serializable so we could put them in Resque.

I agree with you that we can keep `src/Event/ValueObjects/Audience.php` for now. It should be replaced in the long run but it's not necessary for this ticket.

LucWollants: Indeed, thanks for the extra clarification @bertramakers 
I think @JonasVHG change the type of the property https://github.com/cultuurnet/udb3-silex/pull/881/commits/ea6b3a47a5a1c5f3fa3aa7b026aa9647c5ed681e

LucWollants (approved):

PR #882, opened 2022-01-20
III-4187 refactor enum WorkflowStatus

JonasVHG:
### Changed
 - Replace deprecated `WorkflowStatus` with Model `WorkflowStatus`

### Removed

- Removed deprecated `WorkflowStatus`

---


src/Offer/Offer.php
```diff
@@ -730,11 +731,9 @@ public function delete(): void
 
     protected function importWorkflowStatus(CultureFeed_Cdb_Item_Base $cdbItem): void
     {
-        try {
-            $workflowStatus = WorkflowStatus::fromNative($cdbItem->getWfStatus());
-        } catch (\InvalidArgumentException $exception) {
-            $workflowStatus = WorkflowStatus::READY_FOR_VALIDATION();
-        }
+        $wfStatus = $cdbItem->getWfStatus();
+        $workflowStatus = $wfStatus ? new WorkflowStatus(str_replace('READYFORVALIDATION', 'READY_FOR_VALIDATION', strtoupper($wfStatus))) : WorkflowStatus::READY_FOR_VALIDATION();
```
JonasVHG, line 735: I'm not sure that this is the best way to handle udb2 imports that use  `readyforvalidation` while udb3 uses `READY_FOR_VALIDATION` but because the switch-off of UDB2 will happen soon. It might doable as a tempfix.

LucWollants: Maybe we could make a new static method called `fromCultureFeedWorkflowStatus` inside `WorkflowStatus` and that method contains a mapping from the CultureFeed values to the UDB3 values?

src/Event/ReadModel/JSONLD/EventLDProjector.php
```diff
@@ -260,7 +260,7 @@ protected function applyEventCreated(
             $jsonLD->creator = $author->toNative();
         }
 
-        $jsonLD->workflowStatus = WorkflowStatus::DRAFT()->getName();
+        $jsonLD->workflowStatus = WorkflowStatus::DRAFT()->toString();
```
LucWollants, line 263: If I understand correct the old code used the name of the constant which was in uppercase instead of the value which was in lowercase?

JonasVHG: Correct, in the new class everything is uppercase, so `toString()` works as a replacement.

src/Offer/Offer.php
JonasVHG: I was kinda hesitated to do that, because if the UDB2 import stuff is ever removed from UDB3(which I presume could safely be done, when the switch-off has happened). It would require more refactoring, although I agree that I probably looks cleaner in the `WorkflowStatus` class.

bertramakers: Sadly we won't be able to delete all UDB2-related code even after EntryAPI2 and SAPI2 have been switched off. Only the code that process _new_ imports from EntryAPI2, and the CdbXml Service which creates XML for events in UDB3 for SAPI2. But code like this, inside the Event/Place aggregates, needs to keep existing because we will forever have events like `EventImportedFromUdb2` or `PlaceImportedFromUdb2` in our event store which contain CdbXml. So we need to be able to parse that forever (for example here where we use it to determine the state of the event/place, or during replays when we rebuild the read models from the events in the event store). However we don't need to use Culturefeed for that forever, we could do it in a different way (later) to at least remove the dependency on `cultuurnet/culturefeed-php`.

LucWollants (approved):

src/Model/ValueObject/Moderation/WorkflowStatus.php
```diff
@@ -25,4 +25,14 @@ public static function getAllowedValues(): array
             'DELETED',
         ];
     }
+
+    /// This method is only needed to import UDB2 items, which use
+    /// lowercase values without underscores.
+    /// It should only be used in those contexts.
+    /// It could be removed after UDB2 imports are switched off.
```
LucWollants, line 32: Like @bertramakers mentioned here, this method can not be removed https://github.com/cultuurnet/udb3-silex/pull/882#discussion_r788935734

There are actually 3 distinct code flows in which an aggregate will be used in the UDB3 event sourced context:
1. When being updated through a command
2. When being loaded from the event store
3. When being replayed

The first flow typically works with a state updating method like `Offer::addLabel`. This validates internal state of the `Offer` aggregate and when all conditions are fulfilled it will persist an event `LabelAdded` by calling the `Offer::applyLabelAdded` method. It will also deliver the `LabelAdded` event to the event bus and projectors will handle the event to update the JSON representation.

The second flow that loads an aggregate calls `Offer::applyLadedAdded` directly with the persisted `LabelAdded` event.

The third flow when doing a replay takes the events from the event store and pushes them to the event listeners so that the projectors can update the JSON representation.

So `fromCultureFeedWorkflowStatus` will still be used even if the UDB2 imports are turned of (flow 1) because it will be used inside the aggregate loading (flow 2) and replay (flow 3)

src/Offer/Offer.php
LucWollants: Some extra context can be found here https://github.com/cultuurnet/udb3-silex/pull/882/files#r789481893

PR #885, opened 2022-01-21
III-4528 Be more flexible in ISO-8601 date formats

bertramakers:
### Changed

- Classes that validate / decode incoming JSON now parse datetimes in multiple possible ISO-8601 / RFC-3339 datetime formats. For example without second fractions, with milliseconds, and with microseconds. This fixes compatibility with older integrations that send use datetimes with milli/microseconds even though we never officially supported it, but because it happened to work by accident in the past.

---
Ticket: https://jira.uitdatabank.be/browse/III-4528

Note: There are more usages of `DateTime::createFromFormat()` and `DateTimeImmutable::createFromFormat()` then the ones I replaced in this PR, but I tried to change all those that handle datetime strings from "incoming" JSON. Classes/methods that decode datetime strings from e.g. the event store did not need to be replaced since we control the format those are stored in.

### DIFF
--- a/src/DateTimeInvalid.php
+++ b/src/DateTimeInvalid.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3;
+
+use InvalidArgumentException;
+
+final class DateTimeInvalid extends InvalidArgumentException
+{
+}
--- a/src/BookingInfo.php
+++ b/src/BookingInfo.php
@@ -139,12 +139,12 @@ final class BookingInfo implements JsonLdSerializableInterface
 
         $availabilityStarts = null;
         if ($data['availabilityStarts']) {
-            $availabilityStarts = DateTimeImmutable::createFromFormat(\DATE_ATOM, $data['availabilityStarts']);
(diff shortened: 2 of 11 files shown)


src/DateTimeInvalid.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3;
+
+use InvalidArgumentException;
+
+final class DateTimeInvalid extends InvalidArgumentException
```
LucWollants, line 9: Nit: maybe change this to `InvalidDateTime` to be more consistent with `InvalidArgumentException` and `InvalidAgeRangeException`?

LucWollants (approved):
Looks good! The extra comments inside `DateTimeFactory` are a real added value.

src/DateTimeInvalid.php
bertramakers: I went with `DateTimeInvalid` so it would be grouped together with `DateTimeFactory` in the list of files in the `src` directory. But if you want I can change it.

LucWollants: No that is perfect!

PR #889, opened 2022-01-24
III-4438 Remove Event Service

LucWollants:
### Changed
- Refactor `EventExportService` to use `DocumentRepository` instead of `EventServiceInterface`

### Removed
- Removed `ExternalEventService`
- Removed `LocalEventService`
- Removed `EventServiceInterface`

---
Ticket: https://jira.uitdatabank.be/browse/III-4438


bertramakers (approved):
Nice refactor and boyscouting 👍 Definitely an improvement to have less "service" classes that don't add any value, and to have this older code a bit cleaned up like using the `Json` class etc.

PR #890, opened 2022-01-24
III-4492 Fix url update error

bertramakers:
### Fixed

- Creating an organizer with an existing url now returns a better error
- Updating an organizer with an existing url now returns a better error
- Importing a new organizer with an existing url now returns a better error

---
Ticket: https://jira.uitdatabank.be/browse/III-4492

Note: It was not possible to also return an error for importing an existing organizer with a new url that is a duplicate, because the commands are dispatched on an async command bus and handled as jobs. The job for the `UpdateWebsite` command will then fail. But it's currently not possible to return an error response for that since by then the http response is already returned.


LucWollants (approved):
Indeed much better for an API integrator to get a nice warning!

One thing I often doubt about with our good acceptance tests is if we should also make a unit test to verify the API problem report. That would be some duplicated work but I would say better safe than sorry?

bertramakers:
I thought of that too but it's not easy because the unit tests use a `TraceableEventStore` and we cannot easily make it throw the `UniqueConstraintException` we need to check the error. So I thought it would not be worth the time investment in this case given that we already have acceptance tests that cover it as well.

LucWollants:
Ah yes that makes it rather complex to test. An acceptance test is indeed the most pragmatic solution in this case.

PR #894, opened 2022-01-28
III-4539 remove VO for domain and hostname

JonasVHG:
### Changed

- Replaced the Legacy VO for `Domain` and `Hostname`

---


bertramakers (approved):

src/Model/ValueObject/Web/Domain.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Model\ValueObject\Web;
+
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\IsString;
+
+abstract class Domain
```
LucWollants, line 9: Do we need this? I wonder if we use the IP somewhere, if not everything can just become a `Hostname`

JonasVHG: No ValueObjects\Web\IPAddress is not used in udb3-silex. I'll remove it.

LucWollants (approved):

PR #897, opened 2022-01-28
III-4539 remove VO Age

JonasVHG:
### Changed

- Replaced deprecated VO `Age`.

---

### DIFF
--- a/src/Offer/AgeRange.php
+++ b/src/Offer/AgeRange.php
@@ -62,8 +62,8 @@ class AgeRange
 
     public function __toString(): string
     {
-        $from = $this->from ? (string) $this->from->toInteger() : '';
-        $to = $this->to ? (string) $this->to->toInteger() : '';
+        $from = $this->from ? $this->from->toInteger() : '';
+        $to = $this->to ? $this->to->toInteger() : '';
 
         return $from . '-' . $to;
     }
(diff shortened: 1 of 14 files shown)


src/Offer/AgeRange.php
```diff
@@ -62,8 +62,8 @@ public function getTo(): ?Age
 
     public function __toString(): string
     {
-        $from = $this->from ? (string) $this->from : '';
-        $to = $this->to ? (string) $this->to : '';
+        $from = $this->from ? (string) $this->from->toInteger() : '';
```
LucWollants, line 65: It can be useful to add an `toString` on the `Age` class or even on the `IsInteger` class.

@bertramakers any reason why the `toString` method is "missing"?

bertramakers: The string concatenation in the return will automatically cast the `int` from `toInteger()` to a string so you can just leave this `(string)` out.

I think it's missing because it's generally not needed. The value is an `int` and if you want to concatenate it with a string PHP takes care of that, even in `strict_types` mode I think.

bertramakers: You can try it out here http://sandbox.onlinephpfunctions.com/code/5bf28fb3965fe36c554c8a0dabd3d6b797a594ed

LucWollants: That works!

LucWollants (approved):

PR #898, opened 2022-01-28
III-4402 Use organizer schema for validation

bertramakers:
### Added

- Added JSON schema validation to `ImportOrganizerRequestHandler`
- Added label permission check to `ImportLabelsHandler` (for Organizers)

### Removed

- Removed Respect validators/rules that were unused after refactor

---
Ticket: https://jira.uitdatabank.be/browse/III-4402


src/Http/Request/Body/MainLanguageValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,81 @@
...
+    public function parse(ServerRequestInterface $request): ServerRequestInterface
+    {
+        $data = $request->getParsedBody();
+        if (!is_object($data) ||
+            !isset($data->mainLanguage) ||
+            !is_string($data->mainLanguage) ||
+            $data->mainLanguage === ''
+        ) {
+            // Will be caught by JSON schema validation.
+            var_dump('no mainLanguage for some reason');
```
LucWollants: Maybe better to avoid a `var_dump` even if this code should not be executed?

LucWollants: Ok it was removed in a later commit.

LucWollants (approved):

PR #901, opened 2022-01-31
III-4411 post organizers

bertramakers:
### Changed

- Changed `POST /organizers` to work with the import logic, so we can create a complete organizer in the JSON-LD schema instead of the legacy schema

---
Ticket: https://jira.uitdatabank.be/browse/III-4411


src/Http/Organizer/LegacyOrganizerRequestBodyParser.php
```diff
@@ -21,20 +21,46 @@ public function parse(ServerRequestInterface $request): ServerRequestInterface
...
-                $data->mainLanguage => $data->address,
+        if ($mainLanguage && isset($data->name) && is_string($data->name)) {
+            $data->name = (object) [
+                $data->mainLanguage => $data->name,
             ];
         }
 
-        if (isset($data->contact)) {
+        if ($mainLanguage && isset($data->address->streetAddress)) {
+            $data->address->{$mainLanguage} = $data->address->{$mainLanguage} ?? (object) [];
```
LucWollants, line 36: This can maybe only be done once and removed out of the condition?

LucWollants (approved):
Nice work

src/Http/Organizer/LegacyOrganizerRequestBodyParser.php
bertramakers: No because then it sometimes happens that you just get an empty object. I did that at first and a test complained about it :)

bertramakers: I could extract this logic that's in in each `if` to a method thought but I wasn't sure how important this is since it's not really business logic but some data transformation. Plus it's for a legacy schema that shouldn't change anymore. So as long as tests cover it it I thought it's not very important?

PR #904, opened 2022-02-02
III-4539 remove vo UUID role

JonasVHG:
### Changed

- Refactor to remove deprecated VO `UUID`  in `Role` -related classes

---


tests/Http/Role/EditRoleRestControllerTest.php
```diff
@@ -67,8 +68,8 @@ class EditRoleRestControllerTest extends TestCase
 
     public function setUp()
     {
-        $this->roleId = (new UUID())->toNative();
-        $this->labelId = (new UUID())->toNative();
+        $this->roleId = (new UUID('5a359014-d022-48e4-98e2-173496e636fb'))->toString();
```
LucWollants, line 71: I think it's possible to just assign the string value directly?

LucWollants (approved):
Nice work! Only one small remark so I already approved.

Maybe a good candidate to refactor is now `LabelRolesWriteRepositoryInterface`?

PR #905, opened 2022-02-02
III-4363 Validate place import with JSON schema

LucWollants:
### Changed
- Changed place import validation to also use the JSON schema of a place

---
Ticket: https://jira.uitdatabank.be/browse/III-4363


src/Http/Place/CalendarValidationRequestBodyParser.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Place;
```
bertramakers, line 5: I would suggest to already put this in an `Offer` namespace instead. But only supporting `periodic` and `permanent` for now is fine. Then we can also add support for `single` and `multiple` later.

Places don't support `single` and `multiple`, but that is handled by the schema already. And then we don't need to have 2 `CalendarValidationRequestBodyParser` classes.

src/Http/Request/Body/MainLanguageValidatingRequestBodyParser.php
```diff
@@ -17,6 +18,14 @@ final class MainLanguageValidatingRequestBodyParser implements RequestBodyParser
         '/address',
     ];
 
+    private const PLACE_TRANSLATABLE_FIELDS = [
+        '/name',
+        '/description',
+        '/address',
+        '/bookingInfo/urlLabel',
+        '/priceInfo/[]/name',
```
bertramakers, line 26: I'd use JSONPath instead of JSON Pointers here. `/priceInfo/[]/name` is not a valid JSON pointer, because JSON pointers always only point to one specific node (like `/priceInfo/0/name`)

JSONPath on the other hand natively supports wildcards for array values, to target every value in an array.

The paths for these properties would be:
```
$.name
$.description
$.address
$.bookingInfo.urlLabel
$.priceInfo[*].name
```

You can get the data from the JSON using these paths with https://github.com/Galbar/JsonPath-PHP

It's an extra dependency but it feels more logical to use a standard approach here than a workaround to keep using JSON pointers in this class.

bertramakers: And unrelated, I think `status.reason` should also require at least a value for the `mainLanguage`? This is now not the case in the Respect rules but I think every translatable field should follow this rule?

src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -120,6 +123,11 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         $place = RequestBodyParserFactory::createBaseParser(
             new IdPropertyPolyfillRequestBodyParser($this->iriGenerator, $placeId),
             $this->importPreProcessingRequestBodyParser,
+            new JsonSchemaValidatingRequestBodyParser(JsonSchemaLocator::PLACE),
+            new CalendarValidationRequestBodyParser(),
+            new BookingInfoValidationRequestBodyParser(),
+            new PriceInfoValidationRequestBodyParser(),
+            MainLanguageValidatingRequestBodyParser::createForPlace(),
```
bertramakers, line 130: We also need a validator that checks that the `eventtype` term that is used is usable on places. That is currently done here in the Respect validator: https://github.com/cultuurnet/udb3-silex/blob/e301148e32d3a8a0f061f931b50a1240ce670ab6/src/Model/Import/Validation/Place/PlaceImportValidator.php#L31

src/Http/Place/CalendarValidationRequestBodyParser.php
LucWollants: Yes, the `Offer` namespace is the long-term location.

src/Http/Request/Body/MainLanguageValidatingRequestBodyParser.php
LucWollants: Indeed, it was a rather naive approach to modify the index and work around this new requirement 😀
The updated code now uses [Galbar/JsonPath-PHP](https://github.com/Galbar/JsonPath-PHP)

Two small tweaks were needed:
- when working with an indexed it is required to iterate over all the elements
- the errors use json pointers instead of a json path

But I think the tweaks are easy to understand and read.

src/Http/Place/ImportPlaceRequestHandler.php
LucWollants: The `ImportTermRequestBodyParser` now checks also if the `eventtype` is supported or not.

LucWollants:
@bertramakers I think I tackled most of the outstanding issues.

bertramakers (approved):
Perfect now, very nice work 👌

src/Http/Place/ImportPlaceRequestHandler.php
LucWollants: I created it, but it lacks context 😅

PR #908, opened 2022-02-03
III-4542 Fix imports of labels on organizers

bertramakers:
### Fixed

- Trying to add a private label to an organizer via import now just ignores that label instead of returning a 500
- Trying to remove a private label from an organizer via import now just keeps that label instead of removing it
- Visibility of labels added to an organizer via import is now always correct, even if they were in the wrong property (`labels`/`hiddenLabels`) according to their own visibility config in UDB3

---
Ticket: https://jira.uitdatabank.be/browse/III-4542

These changes fix failing tests added in https://github.com/cultuurnet/udb3-acceptance-tests/pull/70

Important! ⚠️ These bugs are likely also present in the import logic of labels on events/places. We'll also need to add these tests for events and places and fix any failing ones, ideally in the same way.


JonasVHG (approved):
Looks good.

PR #909, opened 2022-02-04
III-4539 remove vo UUID final

JonasVHG:
### Changed

- Replaced the last Legacy VO `UUID`'s

---


src/EventSourcing/DomainMessageBuilder.php
```diff
@@ -40,6 +40,13 @@ class DomainMessageBuilder
      */
     private $forReplay;
 
+    private UuidFactoryInterface $uuidFactory;
+
+    public function __construct(UuidFactoryInterface $uuidFactory)
```
LucWollants, line 45: You could work with a default parameter here to avoid lots of changes
```
    public function __construct(UuidFactoryInterface $uuidFactory = null)
    {
        if ($uuidFactory === null) {
           $uuidFactory = new UuidFactory(); 
        }
        
        $this->uuidFactory = $uuidFactory;
    }
```

LucWollants: Normally I would say to avoid using default parameters, but this is existing code and it was previously self-contained. Now the extra parameter gives some overhead.

LucWollants (changes requested):
I would give the `DomainMessageBuilder` constructor method a default parameter to avoid too many changes.

LucWollants (approved):

PR #910, opened 2022-02-08
Fix: allow urls with special characters

LucWollants:
### Fixed
- Fixed allow urls with special characters

### DIFF
--- a/src/Model/ValueObject/Web/Url.php
+++ b/src/Model/ValueObject/Web/Url.php
@@ -38,7 +38,7 @@ class Url
         $value = urlencode($value);
 
         // Revert meaningful characters.
-        // Taken from https://developers.google.com/maps/url-encoding
+        // Taken from https://developers.google.com/maps/url-encoding#special-characters
         $entities = ['%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%2B', '%24', '%2C', '%2F', '%3F', '%25', '%23', '%5B', '%5D'];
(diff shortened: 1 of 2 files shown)


src/Model/ValueObject/Web/Url.php
```diff
@@ -20,10 +20,31 @@ public function __construct($value)
...
+    {
+        // Take into account already encoded urls by first decoding them.
+        $value = urldecode($value);
+
+        // Encode the url but revert various semantic characters.
+        $value = urlencode($value);
+
+        // Revert meaningful characters.
+        $entities = ['%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%2B', '%24', '%2C', '%2F', '%3F', '%25', '%23', '%5B', '%5D'];
+        $replacements = ['!', '*', '\'', '(', ')', ';', ':', '@', '&', '=', '+', '$', ',', '/', '?', '%', '#', '[', ']'];
```
bertramakers, line 42: This feels like a workaround that might miss some edge cases. Maybe we could use https://www.php.net/manual/en/function.parse-url.php instead to divide the URL in meaningful parts and only decode-encode those that need to be encoded? Afterwards you can convert it into a string again with https://php.uz/manual/en/function.http-build-url.php
Using these functions and only encoding for example the paths and query parameter values feels safer I think?

Or alternatively, can you point to a resource that documents this approach as standardized?

LucWollants: Which is based on https://datatracker.ietf.org/doc/html/rfc3986#section-2.2

bertramakers: Perfect, if it's based on a standard it should be safe! Maybe just add a comment to point to these links for future reference.

bertramakers (approved):

PR #914, opened 2022-02-08
III-4539 remove vo url final

JonasVHG:
### Changed

- Replace the last depreacted `Url`'s
- 
---

### DIFF
--- a/src/Organizer/WebsiteNormalizer.php
+++ b/src/Organizer/WebsiteNormalizer.php
@@ -4,24 +4,26 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Organizer;
 
+use CultuurNet\UDB3\Model\ValueObject\Web\Url;
 use ValueObjects\Web\PortNumber;
-use ValueObjects\Web\Url;
+use ValueObjects\Web\Url as LegacyUrl;
 
 final class WebsiteNormalizer
 {
     public function normalizeUrl(Url $url): string
     {
-        $domain = $url->getDomain()->toNative();
+        $legacyUrl = LegacyUrl::fromNative($url->toString());
+        $domain = $legacyUrl->getDomain()->toNative();
         if (strpos($domain, 'www.') === 0) {
             $domain = substr($domain, strlen('www.'));
         }
 
-        $port = $url->getPort() instanceof PortNumber ? ':' . $url->getPort()->toNative() : '';
+        $port = $legacyUrl->getPort() instanceof PortNumber ? ':' . $legacyUrl->getPort()->toNative() : '';
 
-        $queryString = $url->getQueryString()->toNative();
-        $fragment = $url->getFragmentIdentifier()->toNative();
+        $queryString = $legacyUrl->getQueryString()->toNative();
+        $fragment = $legacyUrl->getFragmentIdentifier()->toNative();
 
-        $path = rtrim($url->getPath()->toNative(), '/');
+        $path = rtrim($legacyUrl->getPath()->toNative(), '/');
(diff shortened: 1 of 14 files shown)


src/Organizer/WebsiteNormalizer.php
```diff
@@ -4,24 +4,23 @@
...
         if (strpos($domain, 'www.') === 0) {
             $domain = substr($domain, strlen('www.'));
         }
 
-        $port = $url->getPort() instanceof PortNumber ? ':' . $url->getPort()->toNative() : '';
+        $port = $url->getPort() ? ':' . $url->getPort() : '';
 
-        $queryString = $url->getQueryString()->toNative();
-        $fragment = $url->getFragmentIdentifier()->toNative();
+        $queryString = $url->getQueryString() ? '?' . $url->getQueryString() : '';
```
LucWollants, line 20: It seems that the old `Url` implementation did always add `?` to the query and `#` to the fragment. See this snippet inside the old `Url` implementation:
```
        $query      = $queryString ? new QueryString(\sprintf('?%s', $queryString)) : new NullQueryString();
        $fragment   = $fragmentId ? new FragmentIdentifier(\sprintf('#%s', $fragmentId)) : new NullFragmentIdentifier();
```

Now the new methods return those fields without the prefix, that matches also the returned values from the built-in PHP method `parse_url`. So I think this new implementation is better, what do you prefer @bertramakers ?

src/Model/ValueObject/Web/Url.php
```diff
@@ -18,7 +18,7 @@ class Url
     public function __construct($value)
     {
         $this->guardString($value);
-        $this->guardRegexPattern('/\\Ahttp[s]?:\\/\\//', $value);
+        $this->guardRegexPattern('/\\Ahttp[s]?:\\/\\//', strtolower($value));
```
LucWollants, line 21: Maybe also convert the `$value` to lowercase?

tests/Model/ValueObject/Web/UrlTest.php
```diff
@@ -112,4 +112,63 @@ public function encodeUrlDataProvider(): array
...
+
+    /**
+     * @test
+     */
+    public function it_should_retrieve_a_path(): void
+    {
+        $withPath = new Url('https://www.publiq.be/articles#intro');
+        $withoutPath = new Url('https://www.publiq.be');
+
+        $this->assertEquals('/articles', $withPath->getPath());
```
LucWollants, line 147: Maybe add an extra test with several `/` parts?

src/Model/ValueObject/Web/Url.php
JonasVHG: I suspect that could interfere with the querystring/ UrlEncoding.  The question is: if/how we should accept stuff like `HtTps://WwW.pUBLIQ.Be` 
Existing tests https://github.com/cultuurnet/udb3-silex/blob/690cee1c1f6dc58a2d2730c68c708e222240db3f/tests/Organizer/WebsiteNormalizerTest.php#L60 suggest yes.

LucWollants: True! We discussed that yesterday and did not want to convert `%C3%A9` to `%c3%a9`

src/Organizer/WebsiteNormalizer.php
bertramakers: Without sounds good!

bertramakers: I don't consider the `?` part of the query parameter(s) or `#` part of the fragment. They just indicate where the query parameter(s) and/or fragment start.

LucWollants (approved):
Nice work!

PR #917, opened 2022-02-10
III-4564 Allow non encoded article urls

LucWollants:
### Changed
- Allowed non encoded article urls

---
Ticket: https://jira.uitdatabank.be/browse/III-4564


bertramakers (approved):
The extra unit test makes this very clear 👌

PR #920, opened 2022-02-10
III-4578 Fix PHP notice in event import with organizer

bertramakers:
### Fixed

- Fixed PHP notice when importing an event with an organizer, and the organizer contains more properties than just `@id`. (https://sentry.io/organizations/publiq-vzw/issues/3003000437/?project=5435528)

---
Ticket: https://jira.uitdatabank.be/browse/III-4578


JonasVHG (approved):
Looks Good

PR #922, opened 2022-02-11
III-4570 Fix hidden labels null value

bertramakers:
### Fixed

- Fixed `null` being added as a `hiddenLabel` when an event has a news article with a publisher that is not in the config list of publisher and their labels. If this happens an error will be logged in Sentry, but the code will continue without unintended side-effects.

---
Ticket: https://jira.uitdatabank.be/browse/III-4570


LucWollants (approved):
Looks good.

PR #925, opened 2022-02-14
III-4539 Remove value object Natural

LucWollants:
### Changed
- Refactored `ImageUploaderService` to work with `int`
- Refactored `EventBusForwardingConsumerFactory` to work with `int`
- Refactored `ReadServiceInterface` to work with `int`
- Refactored `ReadRepositoryInterface` to work with `int`
- Refactored `Query` to work with `int`
- Refactored `Entity` to work with `int`

Note: the class `Price` is not modified.

---

### DIFF
--- a/src/Media/ImageUploaderService.php
+++ b/src/Media/ImageUploaderService.php
@@ -13,7 +13,6 @@ use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
 use CultuurNet\UDB3\Model\ValueObject\MediaObject\CopyrightHolder;
 use League\Flysystem\FilesystemOperator;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
-use ValueObjects\Number\Natural;
 use ValueObjects\StringLiteral\StringLiteral;
 
 class ImageUploaderService implements ImageUploaderInterface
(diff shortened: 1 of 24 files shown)


src/Media/ImageUploaderService.php
```diff
@@ -28,17 +27,21 @@ class ImageUploaderService implements ImageUploaderInterface
...
 
     public function __construct(
         UuidGeneratorInterface $uuidGenerator,
         CommandBus $commandBus,
         FilesystemOperator $filesystem,
         string $uploadDirectory,
-        Natural $maxFileSize = null
+        int $maxFileSize = null
     ) {
+        if ($maxFileSize < 0) {
```
JonasVHG, line 41: There are 4 constructors in this PR. Is there any reason why you only check here if it is a positive integer?

JonasVHG (commented):
I just have a question about the constructors, the rest looks good.

src/Media/ImageUploaderService.php
LucWollants: Good remark! I made this a bit stricter and added some extra tests for a few cases.

JonasVHG (approved):

PR #926, opened 2022-02-14
III-4502 Import organizer description

bertramakers:
### Added

- Added support for adding/updating/deleting an organizer's description via `POST /organizers` and `PUT /organizers/{id}`

### Fixed

- Fixed bug that made it impossible to remove the description in the first language (discovered via the acceptance tests that just deleted every translation but the original one remained even though it should have been deleted).

---

Ticket: https://jira.uitdatabank.be/browse/III-4502


LucWollants (approved):
Nice improvement to keep the internal state of descriptions as a simple associative array of strings.

PR #935, opened 2022-02-18
III-4433 prevent local copy

JonasVHG:
### Added

- test local save if upload fails

### Changed

- Only make a local copy if upload fails

---
Ticket: https://jira.uitdatabank.be/browse/III-4433

### DIFF
--- a/src/Media/ImageStorage.php
+++ b/src/Media/ImageStorage.php
@@ -41,9 +41,9 @@ class ImageStorage
                 $source,
                 $this->mediaDirectory . '/' . $destination
             );
+            $this->localFilesystem->delete($source);
             throw $throwable;
         }
-
         $this->localFilesystem->delete($source);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Media/ImageStorage.php
```diff
@@ -28,19 +28,22 @@ public function __construct(
...
+                $this->localFilesystem->readStream($source),
+                [Config::OPTION_VISIBILITY => Visibility::PUBLIC]
+            );
+        } catch (\Throwable $throwable) {
+            // Move to the local file system in case upload fails
+            $this->localFilesystem->copy(
+                $source,
+                $this->mediaDirectory . '/' . $destination
+            );
+            $this->localFilesystem->delete($source);
```
LucWollants, line 44: The delete could go inside a `finally` block which will be executed always.

LucWollants: I think yes https://stackoverflow.com/questions/41616790/how-is-the-keyword-finally-meant-to-be-used-in-php/41616978

LucWollants: I tested it online http://sandbox.onlinephpfunctions.com/code/876bdbe8d689459af8d7a28d3bb116a307393acf

LucWollants (approved):
Nice improvement!

PR #937, opened 2022-02-21
III-4369 Refactor place POST endpoint

LucWollants:
### Changed
- Extend place import response with `placeId` and `url`

---
Ticket: https://jira.uitdatabank.be/browse/III-4369

### DIFF
--- a/src/Http/Place/ImportPlaceRequestHandler.php
+++ b/src/Http/Place/ImportPlaceRequestHandler.php
@@ -244,7 +244,11 @@ final class ImportPlaceRequestHandler implements RequestHandlerInterface
             }
         }
 
-        $responseBody = ['id' => $placeId];
+        $responseBody = [
+            'id' => $placeId,
+            'placeId' => $placeId,
+            'url' => $this->iriGenerator->iri($placeId),
+        ];
         if ($lastCommandId) {
             $responseBody['commandId'] = $lastCommandId;
         }
--- a/tests/Http/Place/ImportPlaceRequestHandlerTest.php
+++ b/tests/Http/Place/ImportPlaceRequestHandlerTest.php
@@ -206,7 +206,11 @@ final class ImportPlaceRequestHandlerTest extends TestCase
 
         $this->assertEquals(201, $response->getStatusCode());
         $this->assertEquals(
-            Json::encode(['id' => $placeId]),
+            Json::encode([
+                'id' => $placeId,
+                'placeId' => $placeId,
+                'url' => 'https://io.uitdatabank.dev/places/' . $placeId,
+            ]),
             $response->getBody()->getContents()
         );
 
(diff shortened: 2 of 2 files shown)


src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -159,12 +160,21 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             // publish the place after creating it.
             // Existing places should always keep their original status, so
             // only do this publish command for new places.
-            $placeAggregate->publish($publishDate);
+            // For now the trigger to do a publish is the imports path.
+            // In the future this needs to be fine-tuned with either default values
+            // or config for specific API partners.
```
bertramakers, line 165: Later it would be defined by an explicit `workflowStatus` property in the JSON. And if the property is not there, we would still check the path to decide the default (publish or not) for backward compatibility with existing integrators.

src/Http/Place/LegacyPlaceRequestBodyParser.php
```diff
@@ -0,0 +1,87 @@
...
+                $data->mainLanguage => $data->name,
+            ];
+        }
+
+        if (isset($data->type) && $data->type instanceof stdClass) {
+            $data->terms = [
+                (object) [
+                    'id' => $data->type->id,
+                    'label' => $data->type->id,
+                    'domain' => $data->type->id,
```
bertramakers, line 35: 1. I think you accidentally copy-pasted the `->id` for all three here
2. It's possible that these properties are not set in the request body, so I'd only set them if they are available

src/Http/Place/LegacyPlaceRequestBodyParser.php
```diff
@@ -0,0 +1,87 @@
...
+            unset($data->address->addressLocality);
+        }
+        if ($mainLanguage && isset($data->address->addressCountry)) {
+            $data->address->{$mainLanguage} = $data->address->{$mainLanguage} ?? (object) [];
+            $data->address->{$mainLanguage}->addressCountry = $data->address->addressCountry;
+            unset($data->address->addressCountry);
+        }
+
+        if (isset($data->calendar) && $data->calendar instanceof stdClass) {
+            $data->calendarType = $data->calendar->calendarType;
```
bertramakers, line 62: `calendarType` is not guaranteed to be in the request body

LucWollants: Should be fixed.

src/Http/Place/ImportPlaceRequestHandler.php
LucWollants: I added a reference to ticket https://jira.uitdatabank.be/browse/III-4609

src/Http/Place/LegacyPlaceRequestBodyParser.php
```diff
@@ -28,12 +28,20 @@ public function parse(ServerRequestInterface $request): ServerRequestInterface
         }
 
         if (isset($data->type) && $data->type instanceof stdClass) {
+            $terms = [
+                'id' => $data->type->id,
```
bertramakers, line 32: `type.id` is technically required but not guaranteed to be set since we don't actually validate the JSON before this step. The JSON schema validation only happens after this transformation of the legacy format.

PR #938, opened 2022-02-24
III-4367 Add PUT place endpoint

LucWollants:
### Added
- Added `PUT` place endpoint

### Removed
- Removed usage of `PlaceValidator`
- Removed `PlaceImportValidator`
- Removed `createPlace` and `createApprovedPlace` methods 

---
Ticket: https://jira.uitdatabank.be/browse/III-4367


bertramakers:
> Removed usage of PlaceValidator

Can we also remove the class `PlaceValidator` itself?

LucWollants:
> Can we also remove the class PlaceValidator itself? I see no functional reason why we should keep it if both the create and update endpoints work with the JSON schema validation.

That class is still used inside `ValidatePlaceJsonLdCommand`, I never used that command. I will ask in Slack if I can remove it.

bertramakers (approved):

PR #939, opened 2022-02-24
III-4594 Broken contactpoints problems

JonasVHG:
### Changed

- Changed the aggregate `Organizer` to store the ContactPoint as `StringArray`

### Fixed

- Fixed the comparison of `ContactPoint` when an old `ContactPoint` was broken, so that `ContactPoint` can be updated.

---
Ticket: https://jira.uitdatabank.be/browse/III-4594


bertramakers (approved):
Nice clean fix! 👌

PR #940, opened 2022-02-28
III-4567 Refactor place labels importing

LucWollants:
### Changed
- Refactored `ImportLabelsHandler` to take into account label visibility and privacy and the labels already present on the place or event
- Changed the `Offer::importLabels` method to handle a simple but correct list of labels
- Refactored the import of places to no longer use `ImportLabelVisibilityRequestBodyParser`
- Refactored the import of events to also work with the refactored `ImportLabelsHandler`

Note: In a follow-up PR it would even be possible to change the import label code from organizers to match the code from places and events closer. Once this is done `ImportLabelVisibilityRequestBodyParser` can be removed.

---
Ticket: https://jira.uitdatabank.be/browse/III-4567

### DIFF
--- a/app/Place/PlaceControllerProvider.php
+++ b/app/Place/PlaceControllerProvider.php
@@ -4,15 +4,12 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Silex\Place;
 
-use CultuurNet\UDB3\Http\Import\ImportLabelVisibilityRequestBodyParser;
 use CultuurNet\UDB3\Http\Import\ImportTermRequestBodyParser;
 use CultuurNet\UDB3\Http\Place\ImportPlaceRequestHandler;
 use CultuurNet\UDB3\Http\Place\UpdateMajorInfoRequestHandler;
 use CultuurNet\UDB3\Http\Place\EditPlaceRestController;
-use CultuurNet\UDB3\Http\Request\Body\CombinedRequestBodyParser;
 use CultuurNet\UDB3\Model\Import\Place\PlaceCategoryResolver;
 use CultuurNet\UDB3\Model\Serializer\Place\PlaceDenormalizer;
-use CultuurNet\UDB3\Silex\Labels\LabelServiceProvider;
 use Silex\Application;
 use Silex\ControllerCollection;
 use Silex\ControllerProviderInterface;
@@ -76,15 +73,7 @@ class PlaceControllerProvider implements ControllerProviderInterface, ServicePro
                 $app['place_repository'],
                 $app['uuid_generator'],
                 new PlaceDenormalizer(),
-                new CombinedRequestBodyParser(
-                    new ImportLabelVisibilityRequestBodyParser(
-                        $app[LabelServiceProvider::JSON_READ_REPOSITORY],
-                        $app[LabelServiceProvider::RELATIONS_READ_REPOSITORY]
-                    ),
-                    new ImportTermRequestBodyParser(
-                        new PlaceCategoryResolver()
-                    )
-                ),
+                new ImportTermRequestBodyParser(new PlaceCategoryResolver()),
                 $app['place_iri_generator'],
                 $app['imports_command_bus'],
                 $app['import_image_collection_factory'],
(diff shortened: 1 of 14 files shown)


src/Offer/CommandHandlers/ImportLabelsHandler.php
```diff
@@ -37,20 +49,88 @@ public function handle($command): void
...
 
-        /** @var Udb3ModelLabel $importLabel */
-        foreach ($command->getLabelsToImport() as $importLabel) {
+        $offer = $this->offerRepository->load($command->getItemId());
+
+        $labelsToImport = $command->getLabels();
+        $labelsToImport = $this->fixVisibility($labelsToImport);
+
+        $labelsOnOffer = new Labels();
+        foreach ($offer->getLabels()->asArray() as $labelOnOffer) {
```
bertramakers, line 58: Since `Offer::getLabels()` is new since this PR, maybe it would be best to do the conversion of a `LabelCollection` to `Labels` in there? That way you don't need to refactor how `$labels` is stored in `Offer` right now but the `getLabels()` method can already consistently return the new `Labels` VO. And it would make this relatively big `handle()` method a little bit smaller and easier to understand.

src/Offer/CommandHandlers/ImportLabelsHandler.php
```diff
@@ -37,20 +49,88 @@ public function handle($command): void
...
+            $labelsOnOffer = $labelsOnOffer->with(
+                new Label(
+                    new LabelName($labelOnOffer->getName()->toNative()),
+                    $labelOnOffer->isVisible()
+                )
+            );
+        }
+
+        $labelsToKeepOnOffer = $this->lockedLabelRepository->getLockedLabelsForItem($command->getItemId());
```
bertramakers, line 67: Would there be a way to also get this from the aggregate one way or another? I think this "locked labels" read model is just based on whether labels are added via `LabelAdded` or `LabelsImported`. So maybe we can also store that in the aggregate state and get it from there, and then remove this read model later (when we also remove `ImportLabelVisibilityRequestBodyParser` and it is no longer used.)

app/Place/PlaceControllerProvider.php
```diff
@@ -76,19 +73,10 @@ function (Application $app) {
...
-                new CombinedRequestBodyParser(
-                    new ImportLabelVisibilityRequestBodyParser(
-                        $app[LabelServiceProvider::JSON_READ_REPOSITORY],
-                        $app[LabelServiceProvider::RELATIONS_READ_REPOSITORY]
-                    ),
-                    new ImportTermRequestBodyParser(
-                        new PlaceCategoryResolver()
-                    )
-                ),
+                new ImportTermRequestBodyParser(new PlaceCategoryResolver()),
```
bertramakers, line 76: Just a suggestion but maybe keep it wrapped in `CombinedRequestBodyParser` so the next person who needs to add a `RequestBodyParser` here knows it exists and just adds the extra `RequestBodyParser` instead of e.g. adding an extra argument to the constructor for it. I don't think the decoration by `CombinedRequestBodyParser` will have a big performance hit.

bertramakers (approved):
As discussed we will pick up the suggestions above in a separate ticket instead. 👉 https://jira.uitdatabank.be/browse/III-4620

src/Offer/CommandHandlers/ImportLabelsHandler.php
bertramakers: Will be picked up in https://jira.uitdatabank.be/browse/III-4620

src/Offer/CommandHandlers/ImportLabelsHandler.php
bertramakers: Will be picked up in https://jira.uitdatabank.be/browse/III-4620

app/Place/PlaceControllerProvider.php
LucWollants: The `CombinedRequestBodyParser` is used again. This was also needed after merging aster and the changes for the `ImportPriceInfoRequestBodyParser`

src/Offer/CommandHandlers/ImportLabelsHandler.php
LucWollants: The method `Offer::getLabels` now returns a VO `Labels`

PR #943, opened 2022-03-04
III-4362 Add JSON schema validation to event import

LucWollants:
### Added
- Added JSON schema validation to event import

---
Ticket: https://jira.uitdatabank.be/browse/III-4362


src/Http/Offer/CalendarValidationRequestBodyParser.php
```diff
@@ -37,6 +37,24 @@ public function parse(ServerRequestInterface $request): ServerRequestInterface
                 );
                 break;
 
+            case 'single':
+            case 'multiple':
+                if (isset($data->subEvent)) {
```
bertramakers, line 42: I'd suggest to add `&& is_array($data->subEvent)` here. Because in the `foreach()` below the `$data->subEvent` is not guaranteed to be an array right now.

bertramakers (approved):
Only one small nit, but the related changes in https://github.com/cultuurnet/stoplight-docs-uitdatabank/pull/105 need some small tweaks I think and might influence the tests in here maybe.

PR #944, opened 2022-03-07
III-4620 Clean up technical debt in label import logic

bertramakers:
### Added

- Added `LabelImportPreProcessor` that contains all generic label import logic like fixing the visibility of the import labels, making sure private labels that the importer is not allowed to use are not added/removed, and making sure every import label has an aggregate

### Changed

- The `ImportLabelsHandler` classes for offers and organizers now use the new `LabelImportPreProcessor` internally
- `Offer::importLabels()` and `Organizer::importLabels()` now only accept one argument, the labels to import, and checks by itself which labels were added via the UI and should thus be kept.

### Removed

- Removed `ImportLabelVisibilityRequestBodyParser`
- Removed `RelationshipModelLockedLabelRepository`
- Removed `LockedLabelRepository` interface

---
Ticket: https://jira.uitdatabank.be/browse/III-4620


tests/Offer/CommandHandlers/ImportLabelsHandlerTest.php
```diff
@@ -16,19 +16,19 @@
 use CultuurNet\UDB3\Event\Events\LabelsImported;
 use CultuurNet\UDB3\Event\EventType;
 use CultuurNet\UDB3\Event\ValueObjects\LocationId;
-use CultuurNet\UDB3\Label;
+use CultuurNet\UDB3\Label as DeprecatedLabel;
```
LucWollants, line 19: Nit: often the prefix `Legacy` is used. Not that important because this is inside the scope of one file but could be handy when searching for `LegacyLabel`. Same applies for `LegacyLabelName`

LucWollants (approved):
Nice improvement, easy to read and easy to understand code for complex business logic.

tests/Offer/CommandHandlers/ImportLabelsHandlerTest.php
bertramakers: Right, fixed!

PR #945, opened 2022-03-08
III-4600 Add table for duplicate places

JonasVHG:
### Added

- Add table for storage of clusters of duplicate places
---
Ticket: https://jira.uitdatabank.be/browse/III-4600

### DIFF
--- a/app/Migrations/Version20220308071939.php
+++ b/app/Migrations/Version20220308071939.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Silex\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20220308071939 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable('duplicate_places');
+
+        $table->addColumn('cluster_id', Type::BIGINT)->setNotnull(true);
+        $table->addColumn('place_uuid', Type::GUID)->setLength(36)->setNotnull(true);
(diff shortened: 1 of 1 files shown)


app/Migrations/Version20220308071939.php
```diff
@@ -0,0 +1,29 @@
...
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable('duplicate_places');
+
+        $table->addColumn('cluster_id', Type::BIGINT)->setNotnull(true);
+        $table->addColumn('place_uuid', Type::GUID)->setLength(36)->setNotnull(true);
+        $table->addColumn('number_of_events', Type::BIGINT)->setNotnull(true);
+        $table->addColumn('wfstatus', Type::TEXT)->setNotnull(true);
+        $table->addColumn('is_canonical', Type::BOOLEAN)->setNotnull(true);
+        $table->addColumn('is_museumpas', Type::BOOLEAN)->setNotnull(true);
```
bertramakers, line 22: Do we need the data from these `is_museumpas`, `wfstatus`, and `number_of_events` columns? If so, I think it would be better not to store them here since this can be stale data (e.g. number of events can have changed since the last nightly data dump to Google Cloud, the analysis, and the dump back into this table).

In any case, they can probably be removed.

Also in general it's best to avoid unnecessary abbrevations like `wfstatus`.

LucWollants: The `number_of_events` was a good indication if a place could be deleted. If the event was created for example a week ago and there are still no events that use this place and it is a duplicate place then it is a good candidate to mark as deleted.

bertramakers: Can we not see that in the relationship table that we already have?

LucWollants: Good point!

PR #946, opened 2022-03-09
III-4609 Manual workflowStatus when creating/updating a place

bertramakers:
### Added

- Added support for `workflowStatus` property to explicitly set the desired workflow status when creating or updating a place

### Changed

- Auto-approval of places now happens via a process manager. This will also work for events, but events also still have auto-approval logic inside `EditEventRestController` and `EventDocumentImporter` for now.

---
Ticket: https://jira.uitdatabank.be/browse/III-4609


LucWollants (approved):
Very clean implementation

PR #950, opened 2022-03-18
Fix single and multiple calendar validation

LucWollants:
### Fixed
- Removed requirement for single and multiple calendars to have start and end date


LucWollants:
> Looks good, but maybe easier to just remove the validators if stuff like this keeps popping up?

The PR I am currently working on, to move the import logic inside the request handler, will remove the validation code. I wanted to be sure the code was fully refactored before removing the validator.

PR #951, opened 2022-03-21
III-4540 Refactor ImportEventRequestHandler to create commands

LucWollants:
### Changed
- Refactored `ImportEventRequestHandler` to create commands instead of using the `EventImporter`

### Removed
- Removed `EventValidator` and related VO validators
- Removed `EventDocumentImporter` and `DocumentImporterInterface`
- Removed `LabelPreProcessingDocumentImporter`
- Removed `TermPreProcessingDocumentImporter`
- Removed `createEvent` and `createApprovedEvent` methods from the legacy `EditEventRestController` and `EventEditingService `

---
Ticket: https://jira.uitdatabank.be/browse/III-4540


src/Http/Event/ImportEventRequestHandler.php
```diff
@@ -52,37 +99,173 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+                $calendar,
+                $theme,
+                $publishDate
+            );
+
+            if ($workflowStatus->sameAs(WorkflowStatus::READY_FOR_VALIDATION())) {
+                $eventAggregate->publish($publishDate);
+            }
+
+            // Events created by specific API partners should automatically be approved.
```
bertramakers, line 159: This is not needed here anymore but should be handled already by https://github.com/cultuurnet/udb3-silex/blob/master/src/Offer/ProcessManagers/AutoApproveForUiTIDv1ApiKeysProcessManager.php

bertramakers: It seems like I forgot to remove it from the `ImportPlaceRequestHandler` when I worked on https://github.com/cultuurnet/udb3-silex/pull/946 though. So it should also be removed there. And hopefully the acceptance tests for the auto-approval will still pass then 😅 I'd suggest to remove it from both request handlers and see if the tests are still green. If so it's a small change that can be committed in this PR (just removing this `if` statement below). If it requires additional changes because it ends up not working, we can do it in a separate PR.

src/Http/Event/ImportEventRequestHandler.php
```diff
@@ -52,37 +99,173 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+        $contactPoint = $eventAdapter->getContactPoint();
+        $commands[] = new UpdateContactPoint($eventId, $contactPoint);
+
+        $description = $eventAdapter->getDescription();
+        if ($description) {
+            $commands[] = new UpdateDescription($eventId, $mainLanguage, $description);
+        }
+
+        // Update the organizer only at the end, because it can trigger UiTPAS to send messages to another worker
+        // which might cause race conditions if we're still dispatching other commands here as well.
```
bertramakers, line 204: This comment is a bit contradictory because it says that the Organizer should only be updated at the very end, but there are still quite some commands after it.

bertramakers: In the old `EventDocumentImporter` it was also at the end.

bertramakers (commented):
So much legacy code removed 🤩

src/Http/Event/ImportEventRequestHandler.php
LucWollants: Oh nice catch, I must have done something wrong when porting the code. I will put this at the end for both the event and place imports.

src/Http/Event/ImportEventRequestHandler.php
LucWollants: As discussed this will be done in a new PR because events don't have an acceptance test for this, see ticket https://jira.uitdatabank.be/browse/III-4636

bertramakers (approved):

PR #954, opened 2022-03-24
Fix missing commandId (breaking the acceptance tests) when using synchronous imports

bertramakers:
These changes handle the import commands synchronous. As a workaround to not break the import tests (or have to change them to no longer wait for the jobs to finish), the imports return a "NIL" command id and this job is always "completed" according to the `/jobs` endpoint

---

Ticket: https://jira.uitdatabank.be/browse/III-4645

### DIFF
--- a/app/Event/EventControllerProvider.php
+++ b/app/Event/EventControllerProvider.php
@@ -92,7 +92,7 @@ class EventControllerProvider implements ControllerProviderInterface, ServicePro
                     new ImportTermRequestBodyParser(new EventCategoryResolver()),
                     new ImportPriceInfoRequestBodyParser($app['config']['base_price_translations'])
                 ),
-                $app['imports_command_bus'],
+                $app['event_command_bus'],
                 $app['import_image_collection_factory']
             )
         );
--- a/app/Event/EventControllerProvider.php
+++ b/app/Event/EventControllerProvider.php
@@ -92,7 +92,7 @@ class EventControllerProvider implements ControllerProviderInterface, ServicePro
                     new ImportTermRequestBodyParser(new EventCategoryResolver()),
                     new ImportPriceInfoRequestBodyParser($app['config']['base_price_translations'])
                 ),
-                $app['event_command_bus'],
+                $app['imports_command_bus'],
                 $app['import_image_collection_factory']
             )
         );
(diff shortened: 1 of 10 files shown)


bertramakers:
The unit tests are only failing because the import request handlers are now suddenly always returning a `commandId` with a NIL uuid, while in the past they did not return a `commandId` at all in the unit tests. But this NIL uuid is a temporary workaround so we don't break the acceptance tests that always expect a `commandId` at the moment. 

If we do go ahead with the synchronous command handling for the imports, we should update the acceptance tests to no longer use the `I wait for the command with id "%{commandId}" to complete` step, and remove this NIL uuid workaround.

LucWollants (approved):
I did run the acceptance tests several times and all is working from that perspective. So I already approved it. The tests are also much faster now 🎉

Maybe we can deploy a feature branch to the acceptance environment?

bertramakers:
I was thinking I could add a feature flag for it instead next week, and then we can actually merge this and deploy it as normal and have a way to quickly enable/disable it on an environment? Then we could even roll it out to production after some testing, but still change it again if there are performance issues.

LucWollants:
> I was thinking I could add a feature flag for it instead next week, and then we can actually merge this and deploy it as normal and have a way to quickly enable/disable it on an environment?

A feature flag seems a good option to enable/disable asynchronous/synchronous processing. But what would be the impact on the testing? I thinking running the tests with fixed synchronous behaviour will be best.

bertramakers:
The way I see it, the goal would be to always do the imports synchronously. So if we introduce a feature flag to switch them from async to sync (= feature flag is on), the goal would be to turn the feature flag on on all environments. And after a while we can remove the feature flag completely and just always handle them synchronously. So the import tests would also run synchronous.

But we don't know if this would have a big performance impact. Handling all these commands synchronously can result in a lot of event store queries because every command handler loads the aggregate from the event store and saves it again.

So my suggestion is to:

1. Add a feature flag to turn on synchronous import handling
2. Turn on the feature flag on local setups, acc, and maybe test. But keep it turned off on production for now.
3. Do some performance tests on acc/test
4. If performance seems okay, turn the feature flag on on production
5. If performance on production seems to be bad after all, we can always quickly turn off the feature flag
6. If performance on production also seems okay, we can just remove the feature flag after a while

If at any point performance is not okay, we can try to fix it of course. But then we can easily switch back to async command handling in the meantime.

bertramakers:
I've reverted the code to always use the `imports_command_bus` again (which will either be synchronous or async depending on the existing feature toggle), and added fixes in the tests to now expect the NIL uuid (because the command bus in the tests is always synchronous).

PR #957, opened 2022-03-25
III-4641 Allow location as plain string inside event

LucWollants:
### Changed
- Allowed location as plain string inside event

Note: I did not add this to the docs because it was never documented and it seems that only the angular UI passes it as a plain string.

---
Ticket: https://jira.uitdatabank.be/browse/III-4641


bertramakers (approved):

LucWollants:
As discussed I also added an extra acceptance test for this.

PR #967, opened 2022-03-31
III-4572 Handle empty time spans on event copy

LucWollants:
### Fixed
- Handled empty time spans also on event copy

---
Ticket: https://jira.uitdatabank.be/browse/III-4572

### DIFF
--- a/src/Http/Offer/LegacyTimeSpansParser.php
+++ b/src/Http/Offer/LegacyTimeSpansParser.php
@@ -16,22 +16,25 @@ final class LegacyTimeSpansParser
 
         // Rename timeSpans to subEvent
         if (isset($data->timeSpans) && is_array($data->timeSpans) && !isset($data->subEvent)) {
-            $data->subEvent = array_map(
-                function ($timeSpan) {
-                    // Rename start to startDate
-                    if ($timeSpan instanceof stdClass && isset($timeSpan->start)) {
-                        $timeSpan->startDate = $this->formatDateTime($timeSpan->start);
-                        unset($timeSpan->start);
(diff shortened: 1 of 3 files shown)


src/Http/Offer/LegacyTimeSpansParser.php
```diff
@@ -16,22 +16,25 @@ public function parse(stdClass $data): StdClass
...
-                    // Rename end to endDate
-                    if ($timeSpan instanceof stdClass && isset($timeSpan->end)) {
-                        $timeSpan->endDate = $this->formatDateTime($timeSpan->end);
-                        unset($timeSpan->end);
-                    }
-                    return $timeSpan;
-                },
-                $data->timeSpans
-            );
+            if (count($data->timeSpans) > 0) {
```
bertramakers, line 19: Nit: Can this `if` not be part of the `if` above it?

```
if (isset($data->timeSpans) && is_array($data->timeSpans) && !isset($data->subEvent) && count($data->timeSpans) > 0))
```

bertramakers (approved):

src/Http/Offer/LegacyTimeSpansParser.php
bertramakers: Looks good otherwise though!

LucWollants: I always wanted to do the `unset($data->timeSpans);`

bertramakers: Ah I missed that! Good call.

PR #968, opened 2022-03-31
III-3926 events with placetypes

JonasVHG:
### Changed

- allow a fallback for events with a `placeType` instead of an `eventType`

---
Ticket: https://jira.uitdatabank.be/browse/III-3926

### DIFF
--- a/src/Event/ReadModel/JSONLD/EventLDProjector.php
+++ b/src/Event/ReadModel/JSONLD/EventLDProjector.php
@@ -69,6 +69,7 @@ use CultuurNet\UDB3\Offer\ReadModel\JSONLD\OfferLDProjector;
 use CultuurNet\UDB3\Offer\ReadModel\JSONLD\OfferUpdate;
 use CultuurNet\UDB3\OrganizerService;
 use CultuurNet\UDB3\Place\LocalPlaceService;
+use CultuurNet\UDB3\Place\PlaceTypeResolver;
 use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
 use CultuurNet\UDB3\ReadModel\DocumentRepository;
 use CultuurNet\UDB3\ReadModel\JsonDocument;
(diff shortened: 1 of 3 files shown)


src/Event/ReadModel/JSONLD/EventLDProjector.php
```diff
@@ -316,7 +320,15 @@ protected function applyEventCopied(
         foreach ($eventJsonLD->terms as $term) {
             if ($term->domain === 'eventtype') {
                 $typeId = new StringLiteral($term->id);
-                $eventType = $this->eventTypeResolver->byId($typeId);
+                // This is a workaround to allow copies of events that
+                // have a placeType instead of an eventType.
+                // These events could also be cleaned up in the future
+                // @see https://jira.uitdatabank.be/browse/III-3926
```
bertramakers, line 326: Very clear! 👌

tests/Event/ReadModel/JSONLD/EventLDProjectorTest.php
```diff
@@ -194,6 +194,47 @@ public function it_handles_new_events_without_theme(): void
         $this->assertEquals($jsonLD, $body);
     }
 
+    /**
+     * @test
+     */
+    public function it_handles_new_events_with_an_incorrect_place_type(): void
+    {
+        $eventId = '1';
+        $eventCreated = new EventCreated(
```
bertramakers, line 203: Shouldn't this test `EventCopied` instead? 🤔

JonasVHG: Indeed, I've refactored the test

src/Event/ReadModel/JSONLD/EventLDProjector.php
JonasVHG: Thanks!

bertramakers (approved):
Thanks for the fix @JonasVHG ! 🥇

PR #969, opened 2022-04-01
III-4657 Handle unique constraint when updating an organizer

LucWollants:
### Fixed
- Handled unique constraint when updating an organizer

---
Ticket: https://jira.uitdatabank.be/browse/III-4657

### DIFF
--- a/src/Http/Organizer/ImportOrganizerRequestHandler.php
+++ b/src/Http/Organizer/ImportOrganizerRequestHandler.php
@@ -169,7 +169,11 @@ final class ImportOrganizerRequestHandler implements RequestHandlerInterface
             // It's not possible to catch the UniqueConstraintException that UpdateWebsite can cause here, since the
             // commands are handled async.
             /** @var string|null $commandId */
-            $commandId = $this->commandBus->dispatch($command);
+            try {
+                $commandId = $this->commandBus->dispatch($command);
+            } catch (UniqueConstraintException $e) {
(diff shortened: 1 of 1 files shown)


src/Http/Organizer/ImportOrganizerRequestHandler.php
```diff
@@ -169,7 +169,11 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             // It's not possible to catch the UniqueConstraintException that UpdateWebsite can cause here, since the
             // commands are handled async.
             /** @var string|null $commandId */
-            $commandId = $this->commandBus->dispatch($command);
+            try {
+                $commandId = $this->commandBus->dispatch($command);
+            } catch (UniqueConstraintException $e) {
+                throw ApiProblem::duplicateUrl($url->toString(), $e->getDuplicateValue());
```
bertramakers, line 175: I just realized why we didn't catch this exception here before. See the comment above (lines 169-170).
I think you can remove that comment now. But it means we should really disable the async import handling in the future once we get the performance issues resolved.

Intracto must've encountered this error in the short window that we had async import handling disabled on acc.

bertramakers (approved):
Already approving because the fix looks good but the comment above it should be removed before it gets merged.

src/Http/Organizer/ImportOrganizerRequestHandler.php
LucWollants: Good point! I did improve the comment and did mention the `synchronous_imports` config setting.

PR #970, opened 2022-04-01
III-4503 Organizer images imports

bertramakers:
### Added

- Image "import" functionality for organizers

### Changed

- `description`, `copyrightHolder`, `inLanguage` are now optional in mediaObject when "importing" events and places

---
Ticket: https://jira.uitdatabank.be/browse/III-4503

### DIFF
--- a/src/Http/Request/Body/ImagesPropertyPolyfillRequestBodyParser.php
+++ b/src/Http/Request/Body/ImagesPropertyPolyfillRequestBodyParser.php
@@ -26,16 +26,16 @@ final class ImagesPropertyPolyfillRequestBodyParser implements RequestBodyParser
     private const PLACES = 'mediaObject';
     private const ORGANIZERS = 'images';
 
-    private string $propertyName;
+    private string $imagesPropertyName;
     private IriGeneratorInterface $iriGenerator;
     private MediaObjectRepository $mediaObjectRepository;
 
     private function __construct(
-        string $propertyName,
+        string $imagesPropertyName,
         IriGeneratorInterface $imagesIriGenerator,
         MediaObjectRepository $mediaObjectRepository
     ) {
-        $this->propertyName = $propertyName;
+        $this->imagesPropertyName = $imagesPropertyName;
         $this->iriGenerator = $imagesIriGenerator;
         $this->mediaObjectRepository = $mediaObjectRepository;
     }
@@ -72,7 +72,7 @@ final class ImagesPropertyPolyfillRequestBodyParser implements RequestBodyParser
             return $request;
         }
 
-        $propertyName = $this->propertyName;
+        $propertyName = $this->imagesPropertyName;
         if (!isset($data->$propertyName) || !is_array($data->$propertyName)) {
             return $request;
         }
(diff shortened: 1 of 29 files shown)


src/Http/Request/Body/ImagesPropertyPolyfillRequestBodyParser.php
```diff
@@ -0,0 +1,173 @@
...
+    private string $propertyName;
+    private IriGeneratorInterface $iriGenerator;
+    private MediaObjectRepository $mediaObjectRepository;
+
+    private function __construct(
+        string $propertyName,
+        IriGeneratorInterface $imagesIriGenerator,
+        MediaObjectRepository $mediaObjectRepository
+    ) {
+        $this->propertyName = $propertyName;
```
LucWollants, line 38: Maybe change this to `imagePropertyName`?

src/Model/Serializer/ValueObject/MediaObject/ImageDenormalizer.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+        $this->imageIdParser = $imageIdParser;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        if (!$this->supportsDenormalization($data, $class, $format)) {
```
LucWollants, line 37: This is something we don't consistently verify. We should maybe add this to all denormalizers?

For example:
https://github.com/cultuurnet/udb3-silex/blob/master/src/Model/Serializer/ValueObject/Calendar/StatusDenormalizer.php#L21

src/Organizer/Organizer.php
```diff
@@ -340,6 +340,42 @@ public function updateImage(
         );
     }
 
+    public function importImages(Images $images): void
+    {
+        $currentImages = $this->images->toArray();
+        $importImages = $images->toArray();
+
+        $compareImages = static function (Image $a, Image $b) {
```
LucWollants, line 348: Nit: maybe use the following shorthand
```
$compareImages = static fn (Image $a, Image $b) => strcmp($a->getId()->toString(), $b->getId()->toString());
```

LucWollants (approved):
Very clean code

src/Http/Request/Body/ImagesPropertyPolyfillRequestBodyParser.php
bertramakers: Done!

src/Model/Serializer/ValueObject/MediaObject/ImageDenormalizer.php
bertramakers: It could help in case someone injects an incorrect implementation of `DenormalizerInterface` somewhere. But in that case we'll probably get some other kind of error as well that would point to the issue. I'd say feel free to add it in new denormalizers and update existing ones if you change them anyway :)

src/Organizer/Organizer.php
bertramakers: Yeah good suggestion. I just copied this from `Offer::importImages()` :)

PR #971, opened 2022-04-04
III-4614 Add routing key

LucWollants:
### Added
- Inject a closure to set the routing key of messages on exchange

---
Ticket: https://jira.uitdatabank.be/browse/III-4614


app/Metadata/MetadataServiceProvider.php
```diff
@@ -73,11 +72,6 @@ function () use ($context) {
             function (Request $request, Application $app) {
                 // If we're handling requests, the API used is usually the JSON-LD API.
                 $app['api_name'] = ApiName::JSONLD;
```
bertramakers, line 74: If you set this somewhere in `web/index.php` before `bootstrap.php` gets included, the whole `before()` middleware can be removed here. That might be clearer since it's the same approach as used in `bin/udb3.php` (set the API name directly in the PHP entry point)

PR #972, opened 2022-04-04
III-4602 mark as duplicate in bulk

JonasVHG:
### Added

- Added Command to mark Places as duplicate in Bulk

---
Ticket: https://jira.uitdatabank.be/browse/III-4602


src/Label/ReadModels/Relations/Repository/ReadRepositoryInterface.php
```diff
@@ -14,6 +15,11 @@ interface ReadRepositoryInterface
      */
     public function getLabelRelations(LabelName $labelName);
 
+    /**
+     * @return string[]
+     */
+    public function getLabelRelationsAsString(LabelName $labelName, RelationType $relationType): array;
```
LucWollants, line 21: Maybe more important then the `AsString` suffix is adding `ForType`, so I think `getLabelRelationsForType` is better. The return type will be clear from the header block.

src/Label/ReadModels/Relations/Repository/Doctrine/DBALReadRepository.php
```diff
@@ -33,6 +34,18 @@ public function getLabelRelations(LabelName $labelName)
         }
     }
 
+    public function getLabelRelationsAsString(LabelName $labelName, RelationType $type): array
```
LucWollants, line 37: I don't think this is tested on its own?

app/Console/MarkPlaceAsDuplicateInBulkCommand.php
```diff
@@ -0,0 +1,108 @@
...
+use CultuurNet\UDB3\Place\Canonical\DuplicatePlaceRepository;
+use CultuurNet\UDB3\Place\Commands\MarkAsDuplicate;
+use Psr\Log\LoggerAwareInterface;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Logger\ConsoleLogger;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+class MarkPlaceAsDuplicateInBulkCommand extends AbstractCommand
```
LucWollants, line 21: I would maybe change the name to something like `MarkPlacesAsDuplicate` with a good suffix indicating the duplicate places table.

app/Console/MarkPlaceAsDuplicateInBulkCommand.php
```diff
@@ -0,0 +1,108 @@
...
+        parent::__construct($commandBus);
+        $this->processManager = $processManager;
+        $this->duplicatePlaceRepository = $duplicatePlaceRepository;
+        $this->canonicalService = $canonicalService;
+    }
+
+    public function configure(): void
+    {
+        $this->setName('place:mark-as-duplicate-bulk');
+        $this->setDescription('Marks multiple Places as duplicate of another one, implicitly making that one canonical');
```
LucWollants, line 44: We have to make clear what the source will be

src/Place/Canonical/PlaceCluster.php
```diff
@@ -0,0 +1,28 @@
...
+
+    private array $placeIds;
+
+    public function __construct(int $clusterId, array $placeIds)
+    {
+        $this->clusterId = $clusterId;
+        $this->placeIds = $placeIds;
+    }
+
+    public function getClusterId(): int
```
LucWollants, line 19: At first glance this extra class looked like a good abstraction, but now after inspecting the code the method `getClusterId` is nowhere used and `getPlacesIds` just returns an array of strings. So I think the method `DuplicatePlaceRepository::getCluster` could just return a string array.

src/Place/Canonical/DuplicatePlaceRepository.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\Canonical;
+
+interface DuplicatePlaceRepository
+{
+    public function getClusterIds(): array;
```
LucWollants, line 9: I think a doc block with return type `string[]` can be useful

app/Console/MarkPlaceAsDuplicateInBulkCommand.php
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        $clusterIds = $this->duplicatePlaceRepository->getClusterIds();
+
+        foreach ($clusterIds as $clusterId) {
+            $placeCluster = $this->duplicatePlaceRepository->getCluster($clusterId);
+            $canonicalId = $this->canonicalService->getCanonical($placeCluster->getPlacesIds());
+            $duplicateIds = array_diff($placeCluster->getPlacesIds(), [$canonicalId]);
+
+            if (!$this->askConfirmation($input, $output, count($duplicateIds))) {
```
LucWollants, line 69: I think asking this in a loop of clusters can be annoying 😅. As a work around you could just give the amount as clusters or add a extra method to get all places.

app/Console/MarkPlaceAsDuplicateInBulkCommand.php
```diff
@@ -0,0 +1,108 @@
...
+
+            foreach ($duplicateIds as $duplicateId) {
+                try {
+                    $this->commandBus->dispatch(
+                        new MarkAsDuplicate(
+                            $duplicateId,
+                            $canonicalId
+                        )
+                    );
+                    $logger->info('Successfully marked place as duplicate');
```
LucWollants, line 81: I would be more explicit here and say which place is marked as duplicate and which place is the canonical.

app/Console/MarkPlaceAsDuplicateInBulkCommand.php
```diff
@@ -0,0 +1,108 @@
...
+                } catch (CannotMarkPlaceAsCanonical | CannotMarkPlaceAsDuplicate $e) {
+                    $logger->error($e->getMessage());
+                    return 1;
```
LucWollants, line 84: Not sure we want to exit this command which will run in bulk an not on a single place.

tests/Place/Canonical/CanonicalServiceTest.php
```diff
@@ -0,0 +1,173 @@
...
+    private DBALDuplicatePlaceRepository $duplicatePlaceRepository;
+
+    private DocumentRepository $documentRepository;
+
+    private CanonicalService $canonicalService;
+
+    public function setUp(): void
+    {
+        $this->documentRepository = new InMemoryDocumentRepository();
+        $labels_relations = new Table('labels_relations');
```
LucWollants, line 31: Casing of variables should always be `camelCasing`, so it should be `labelsRelations`

tests/Place/Canonical/CanonicalServiceTest.php
```diff
@@ -0,0 +1,173 @@
...
+            ]
+        );
+
+        $this->assertEquals('901e23fe-b393-4cc6-9307-8e3e3f2ea77f', $canonicalId);
+    }
+
+    /**
+     * @test
+     */
+    public function it_will_get_the_oldest_place_if_equel_nr_of_events(): void
```
LucWollants, line 161: Small type, should be `equal`

src/Place/Canonical/CanonicalService.php
```diff
@@ -0,0 +1,88 @@
...
+
+class CanonicalService
+{
+    private string $museumpasLabel;
+
+    private RepositoryInterface $eventRelationsRepository;
+
+    private ReadRepositoryInterface $labelRelationsRepository;
+
+    private DocumentRepository $documentRepository;
```
LucWollants, line 21: This name can be more explicit, I think this are places?

tests/Place/Canonical/CanonicalServiceTest.php
```diff
@@ -0,0 +1,173 @@
...
+use Doctrine\DBAL\Types\Type;
+use PHPUnit\Framework\TestCase;
+
+class CanonicalServiceTest extends TestCase
+{
+    use DBALTestConnectionTrait;
+
+    private DBALDuplicatePlaceRepository $duplicatePlaceRepository;
+
+    private DocumentRepository $documentRepository;
```
LucWollants, line 24: This can become a local variable inside the constructor.

app/Console/MarkPlacesAsDuplicateFromTableCommand.php
```diff
@@ -40,7 +40,7 @@ public function __construct(
     public function configure(): void
     {
         $this->setName('place:mark-as-duplicate-bulk');
-        $this->setDescription('Marks multiple Places as duplicate of another one, implicitly making that one canonical');
+        $this->setDescription('Marks multiple Places as duplicate of another one, based on the autofilled table');
```
LucWollants, line 38: I think using the name of the table `duplicate_places` is even more clear.

app/Console/MarkPlaceAsDuplicateInBulkCommand.php
```diff
@@ -48,12 +48,14 @@ public function configure(): void
             InputOption::VALUE_NONE,
             'Skip confirmation.'
         );
+        $this->addOption('dry-run', null, InputOption::VALUE_NONE, 'Execute the mark-places-as-duplicate as a dry run.');
```
LucWollants: On line 43 the old name is still used.

src/Place/Canonical/CanonicalService.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->museumpasLabel = $museumpasLabel;
+        $this->eventRelationsRepository = $eventRelationsRepository;
+        $this->labelRelationsRepository = $labelRelationsRepository;
+        $this->placeRepository = $placeRepository;
+    }
+
+    public function getCanonical(array $placeIds): string
+    {
+        $placesWithMuseumpas = $this->getPlacesWithMuseumPasInCluster($placeIds);
+        if (count($placesWithMuseumpas) === 1) {
```
LucWollants, line 38: I think we can throw a custom exception if `placesWithMuseumpas` is bigger than 1. The other code can stay.

src/Place/Canonical/Exception/MuseumPassNotUniqueInCluster.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\Canonical\Exception;
+
+class MuseumPassNotUniqueInCluster extends \Exception
+{
+    public function __construct(int $amount)
```
LucWollants, line 9: It could be useful to also pass and report the cluster id.

LucWollants (approved):
Nice work! Thanks for the extra effort to fix the code review remarks.

src/Place/Canonical/Exception/MuseumPassNotUniqueInCluster.php
JonasVHG: Problem here would be that we got rid of the `PlaceCluster` and just give an array of `uuid` to `getCanonical()`, so we don't know the clusterId. I think we could log the clusterId in `MarkPlacesAsDuplicateFromTableCommand` so we can see if this(or another error) happens during the process of a cluster.

LucWollants: Good point, that makes me think that `$cluster = $this->duplicatePlaceRepository->getCluster($clusterId);` from the CLI needs to be done inside the `CanonicalService`. The more logic we can put in the service the better.

JonasVHG: I've made the changes. It think it would also be possible to remove `$clusterIds = $this->canonicalService->getClusterIds();` in `MarkPlacesAsDuplicateFromTableCommand` and just return 1  Associative Arrays with all the canonical places as keys instead of Cluster by Cluster. But since that might take a bit hit on the database, it might be prudent to wait till the datateam has updated their script to no longer use deleted & rejected places?

PR #973, opened 2022-04-04
III-4659 Fix deleting main image from an organizer

LucWollants:
### Fixed
- Fixed deleting the main image from an organizer by using correct object type

---
Ticket: https://jira.uitdatabank.be/browse/III-4659


tests/Organizer/OrganizerLDProjectorTest.php
```diff
@@ -564,6 +564,26 @@ public function it_handles_removing_an_image(): void
...
+        $this->mockGet($organizerId, 'organizer_with_two_images.json');
+
+        $domainMessage = $this->createDomainMessage(
+            new ImageRemoved(
+                $organizerId,
+                '03789a2f-5063-4062-b7cb-95a0a2280d92',
+            )
+        );
+
+        $this->expectSave($organizerId, 'organizer_with_main_image_removed.json');
```
bertramakers, line 582: Nit: A better file name might be `organizer_with_one_image.json`.
Because the organizer still has a main image, and outside of this test the link with `organizer_with_two_images.json` will not be clear and it might be confusing that it's called "with main image removed" if it still has a mainImage (but a different one)

bertramakers: It's the `_and_removed_main_image` or `_with_main_image_removed` that I think is unclear. The organizer still has a main image, it was just changed because the image that was the main image was removed.

I'd suggest to structure & name the test data files like this:

`organizer_with_two_images.json`:
```
{
  "images": [
    // image 1,
    // image 2
  ],
  "mainImage": // image 2
}
```

`organizer_with_one_image.json`:
```
{
  "images": [
    // image 1
  ],
  "mainImage": // image 1
}
```

Then in your test you start with `organizer_with_two_images.json`, remove image 2, and assert that you get `organizer_with_one_image.json`.

Of course if there is already a `organizer_with_image.json` it would be best to use that instead of adding another `organizer_with_one_image.json`.

LucWollants: But how to make the distinction between one where the main image was deleted and another where the main image was not deleted?

LucWollants: You are right that the filename should not describe the test, but for removing an image we have two scenarios that result in two different files with only one image.

One test to remove the image that is also the main image will have another JSON result than the test that removes the image that is not the main image. So 2 files are needed with one image, but one with image A and another with image B.

LucWollants: But I found a solution to change the main image in the new test. After that I am able to reuse the same test file which i also renamed.

PR #984, opened 2022-04-15
III-4665 Catch publish exceptions

LucWollants:
### Changed
- Catched publish exceptions

---
Ticket: https://jira.uitdatabank.be/browse/III-4665

### DIFF
--- a/src/Http/Place/ImportPlaceRequestHandler.php
+++ b/src/Http/Place/ImportPlaceRequestHandler.php
@@ -229,7 +229,6 @@ final class ImportPlaceRequestHandler implements RequestHandlerInterface
             try {
                 $commandId = $this->commandBus->dispatch($command);
             } catch (NotAllowedToPublish $notAllowedToPublish) {
-
             }
             $lastCommandId = $commandId ?? null;
         }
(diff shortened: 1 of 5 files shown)


src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -224,13 +225,12 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
         foreach ($commands as $command) {
-            /** @var string|null $commandId */
-            $commandId = $this->commandBus->dispatch($command);
-            if ($commandId) {
-                $lastCommandId = $commandId;
+            try {
+                $commandId = $this->commandBus->dispatch($command);
+            } catch (NotAllowedToPublish $notAllowedToPublish) {
             }
+            $lastCommandId = $commandId ?? null;
```
bertramakers, line 233: Nice, I didn't realize you could also do this to check possibly undefined variables

bertramakers (approved):

src/Http/Place/ImportPlaceRequestHandler.php
LucWollants: Some sugar since PHP 7.0: https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op

LucWollants: But in this case healthy for the code 😀

bertramakers: I knew about the `??` operator but not that you could also use it for undefined variables, I thought it only worked for arrays/objects like `$arr['key'] ?? : null`. But makes sense if it's basically a shorthand for `isset()`

PR #990, opened 2022-04-26
III-4677 Add endpoint to update attendance mode

lucwollants:
### Added
- Added new endpoint to update attendance mode. This endpoint takes into account the related location of the event.

---
Ticket: https://jira.uitdatabank.be/browse/III-4677

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -5068,12 +5068,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/cultuurnet/udb3-json-schemas.git",
-                "reference": "28582269bedf4286f26582e36a87d1c351e27ffd"
+                "reference": "39fb511ffff67c47ada80de2e5db0a4668e4f1d1"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/cultuurnet/udb3-json-schemas/zipball/28582269bedf4286f26582e36a87d1c351e27ffd",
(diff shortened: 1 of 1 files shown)


src/Http/Event/UpdateAttendanceModeRequestHandler.php
```diff
@@ -0,0 +1,94 @@
...
+        RepositoryInterface $eventRelationsRepository,
+        AttendanceModeWithLocation $attendanceModeWithLocation
+    ): void {
+        if ($attendanceModeWithLocation->getLocationId() === null &&
+            !$attendanceModeWithLocation->getAttendanceMode()->sameAs(AttendanceMode::online())) {
+            $location = $eventRelationsRepository->getPlaceOfEvent($eventId);
+
+            if ($location === null || $location === LocationId::VIRTUAL_LOCATION) {
+                throw ApiProblem::bodyInvalidData(
+                    new SchemaError('/location', 'The event has no related location, please provide a location')
```
bertramakers, line 89: Small nit, but:

```suggestion
                    new SchemaError('/location', 'A location is required when changing an online event to mixed or offline')
```

might be a bit clearer?

bertramakers:
Note: I've changed the target of this PR to `feature/III-4667-polyfill-virtual-location` since that was merged into this.

bertramakers (approved):
Looks very nice, my only comment is a nit so already approving because it can be merged either way

PR #992, opened 2022-04-28
III-4661 correct id for sameAs

JonasVHG:
### Fixed

- Use correct ID in `sameAs` when copying an event

---
Ticket: https://jira.uitdatabank.be/browse/III-4661


LucWollants:
As discussed we will also add a decorator that always sets the sameAs so a replay is not needed to fix this for existing events.

bertramakers:
Normally we add polyfills like this to the existing `NewPropertyPolyfillOfferRepository` decorator, so all polyfills that can be removed after the next complete replay are in 1 place. (Since then they should not be needed anymore)

LucWollants:
> Normally we add polyfills like this to the existing `NewPropertyPolyfillOfferRepository` decorator, so all polyfills that can be removed after the next complete replay are in 1 place. (Since then they should not be needed anymore)

I proposed to use another class because this change fixes a bug. But I guess we can reuse the `NewPropertyPolyfillOfferRepository` decorator and make clear it's fixing a bug based on method name.

@JonasVHG is that clear for you or we need to discuss this in detail?

src/SameAsInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3;
+
+interface SameAsInterface
```
LucWollants, line 7: This time I would not make an interface, there is only going to be one implementation.

src/SameAs.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3;
+
+class SameAs implements SameAsInterface
```
LucWollants, line 7: Maybe this needs an extra suffix or prefix to make it more clear it has to do with uitinvlaanderen?

src/Offer/ReadModel/JSONLD/SameAsOfferRepositoryDecorator.php
```diff
@@ -0,0 +1,44 @@
...
+use CultuurNet\UDB3\SameAsInterface;
+
+final class SameAsOfferRepositoryDecorator extends DocumentRepositoryDecorator
+{
+    private SameAsInterface $sameAs;
+
+    public function __construct(DocumentRepository $documentRepository, SameAsInterface $sameAs)
+    {
+        parent::__construct($documentRepository);
+        $this->sameAs = $sameAs;
```
LucWollants, line 19: Feel free to inline the construction of the sameAsGenerator, no need to inject it.

JonasVHG: Since I've moved it to NewPropertyPolyfillOfferRepository it's no longer relevant.

LucWollants (approved):
Nice work.

PR #993, opened 2022-04-28
III-4678 Update attendance mode when setting location

LucWollants:
### Changed
- Update attendance mode when setting real location on online event
- Prevent using virtual location on the location endpoint

---
Ticket: https://jira.uitdatabank.be/browse/III-4678

### DIFF
--- a/src/Http/ApiProblem/ApiProblem.php
+++ b/src/Http/ApiProblem/ApiProblem.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Http\ApiProblem;
 
-use CultuurNet\UDB3\Http\Docs\Stoplight;
 use CultuurNet\UDB3\Offer\OfferType;
 use Exception;
 
@@ -407,14 +406,5 @@ final class ApiProblem extends Exception
         );
         $e->setExtraProperties(['label' => $labelName]);
         return $e;
-    }
-
-    public static function useAttendanceMode(string $detail): self
-    {
-        return self::pathParameterInvalid(
-            $detail
-            . ' For more information check the documentation of the update attendance mode endpoint.'
-            . ' See: ' . Stoplight::ATTENDANCE_MODE_UPDATE
-        );
     }
 }
--- a/src/Http/ApiProblem/ApiProblem.php
+++ b/src/Http/ApiProblem/ApiProblem.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Http\ApiProblem;
 
+use CultuurNet\UDB3\Http\Docs\Stoplight;
 use CultuurNet\UDB3\Offer\OfferType;
 use Exception;
 
@@ -406,5 +407,14 @@ final class ApiProblem extends Exception
         );
         $e->setExtraProperties(['label' => $labelName]);
         return $e;
+    }
+
+    public static function useAttendanceMode(string $detail): self
+    {
+        return self::pathParameterInvalid(
+            $detail
+            . ' For more information check the documentation of the update attendance mode endpoint.'
+            . ' See: ' . Stoplight::ATTENDANCE_MODE_UPDATE
+        );
     }
 }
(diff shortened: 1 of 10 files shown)


src/Http/Event/UpdateLocationRequestHandler.php
```diff
@@ -33,6 +33,12 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         $eventId = $routeParameters->getEventId();
         $locationId = $routeParameters->get('locationId');
 
+        if ((new LocationId($locationId))->isVirtualLocation()) {
+            throw ApiProblem::pathParameterInvalid(
+                'Instead of passing the virtual location, please update the attendance mode.'
```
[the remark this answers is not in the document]
LucWollants: The main reason is that we don't have to expose/handle the virtual location. An API integrator never has to use the virtual location, not with the import/create event endpoint and not with the update attendance mode endpoint.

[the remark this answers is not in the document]
LucWollants: Also the virtual location is something we partly invented to make sure every event has a location.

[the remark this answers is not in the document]
bertramakers: > an extra safety to make sure everything related to attendance mode goes through the attendance mode endpoint

But then should we also not throw an error when trying to set a real location on an online event, instead of automatically setting the attendanceMode to `offline`? Which is also a guess, could be that the integrator intended `mixed`.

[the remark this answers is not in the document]
LucWollants: All errors is fine for me, I will also discuss with @corneelwille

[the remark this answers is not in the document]
LucWollants: The code is updated to throw an exception

src/Http/Event/ImportEventRequestHandler.php
```diff
@@ -152,13 +152,18 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
         } else {
             if ($workflowStatus->sameAs(WorkflowStatus::READY_FOR_VALIDATION())) {
                 $commands[] = new Publish($eventId, $publishDate);
             }
 
             $commands[] = new UpdateTitle($eventId, $mainLanguage, $title);
             $commands[] = new UpdateType($eventId, $type->getId());
+            // The attendance mode needs to be updated before the location can be changed.
+            // For example passing a real location to an online event is not allowed.
+            $commands[] = new UpdateAttendanceMode($eventId, $event->getAttendanceMode());
```
bertramakers, line 166: Do we check that the new location exists in the validation layer? If not, I think it's possible that we update the attendanceMode from online to offline/mixed, but then the `UpdateLocation()` after it might fail and the event might become offline/mixed but with a virtual location?

Is it possible to combine the update of the attendance mode and the update of the location in a single command, so if either one fails the other does not get changed?

LucWollants: Validating the existence of the location is indeed missing here. We could take the same approach as in `UpdateLocationRequestHandler` and verify the existence of the location before doing any commands?

I think the check for and existing location is useful also out of the context of updating the attendance mode.

bertramakers: Sounds good!

LucWollants: I added validation to check the existence of the provided location.

src/Http/ApiProblem/ApiProblem.php
```diff
@@ -407,4 +408,13 @@ public static function labelNotAllowed(string $labelName): self
         $e->setExtraProperties(['label' => $labelName]);
         return $e;
     }
+
+    public static function useAttendanceMode(string $detail): self
+    {
+        return self::pathParameterInvalid(
```
bertramakers, line 414: If/when we refactor the "update location" endpoint to use a JSON body (so an integrator can also use the complete URL of a location, instead of just the UUID), this type will be incorrect and I think it will be easy to glance over because it's now hidden inside `useAttendanceMode` 🤔

LucWollants: I inlined the usage of `pathParameterInvalid` again and made a private method to create the message.

bertramakers (approved):
Perfect now 🤩  thanks for the optimizations 👌

PR #994, opened 2022-04-29
III-4592 Fix duplicate labels with different visibility on organizers/places/events

bertramakers:
### Fixed

- When an event/place/organizer has a label with an incorrect or outdated visibility in its internal state (due to the visibility being stored in LabelAdded or being derived from the XML in Imported/UpdatedFromUDB2 events), the `importLabels()` method no longer crashes on the same label but with a new/correct visibility (as it is always set correctly by `LabelImportPreProcessor`). Instead the label with the incorrect/old visibility is removed (resulting in a `LabelRemoved` event), and the same label but with the new/correct visibility is added (resulted in a `LabelAdded` event). This was the event listeners downstream will also update the visibility to the correct one.
- When an event/place/organizer is imported or updated via XML and the XML contains an invalid keyword visibility, the keyword is now projected as a label with the correct visibility.
- For existing events/places/organizers a polyfill is provided so that keywords that appear both in `labels` and `hiddenLabels` because they were imported/updated from XML multiple times (with different keyword visibility) are filtered out of the incorrect property.
- A clearer API problem is returned when an event/place/organizer is imported via JSON if they have the same label in `labels` and `hiddenLabels`.

---
Ticket: https://jira.uitdatabank.be/browse/III-4592

Todo:

- [x] Fix `importLabels()` crashing when importing a label that is already on the event/place/organizer but with an invalid/outdated visibility inside its aggregate state
- [x] Fix incorrect label visibility on event/place/organizer JSON-LD in projector (will only work for new XML imports - which are not happening anymore - or after a replay for existing events/places/organizers)
- [x] Fix duplicate labels on event/place/organizer JSON-LD (same label both in `labels` and `hiddenLabels`) by adding a decorator/polyfill that looks for these duplicates and checks what the correct visibility is
  - [x] For organizers
  - [x] For events + places (offers)
- [x] Add validation/error for JSON-LD imports with same label both in `labels`/`hiddenLabels`

### DIFF
--- a/app/Event/EventJSONLDServiceProvider.php
+++ b/app/Event/EventJSONLDServiceProvider.php
@@ -26,7 +26,7 @@ use CultuurNet\UDB3\Offer\Popularity\PopularityRepository;
 use CultuurNet\UDB3\Offer\ReadModel\JSONLD\CdbXMLItemBaseImporter;
 use CultuurNet\UDB3\Offer\ReadModel\JSONLD\CuratorEnrichedOfferRepository;
 use CultuurNet\UDB3\Offer\ReadModel\JSONLD\MediaUrlOfferRepositoryDecorator;
-use CultuurNet\UDB3\Offer\ReadModel\JSONLD\NewPropertyPolyfillOfferRepository;
+use CultuurNet\UDB3\Offer\ReadModel\JSONLD\PropertyPolyfillOfferRepository;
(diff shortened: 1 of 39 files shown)


LucWollants (approved):
The business logic is complex and there are plenty of edge cases, but the code is easy to read and understand.

I also ran all acceptance tests and they are all green.

PR #997, opened 2022-05-02
III-4685 prevent identical tariffs for imports

JonasVHG:
### Added

- Add a check on duplicate `Tariff` names for imports

---
Ticket: https://jira.uitdatabank.be/browse/III-4685


tests/Model/ValueObject/Price/TariffsTest.php
```diff
@@ -68,4 +68,66 @@ public function it_should_find_only_unique_values(): void
...
+
+        $tariffs = new Tariffs($tariff1, $tariff2, $tariff3);
+
+        $this->assertFalse($tariffs->hasDuplicates());
+    }
+
+    /**
+     * @test
+     */
+    public function it_should_find_same_tariff_names_in_another_language(): void
```
LucWollants, line 88: I think this name is a bit misleading, because it does not find names in another language.

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,41 @@
...
+use CultuurNet\UDB3\Model\Serializer\ValueObject\Price\PriceInfoDenormalizer;
+use CultuurNet\UDB3\Model\ValueObject\Price\PriceInfo;
+use Psr\Http\Message\ServerRequestInterface;
+
+class PriceInfoValidatingRequestBodyParser implements RequestBodyParser
+{
+    public function parse(ServerRequestInterface $request): ServerRequestInterface
+    {
+        $data = $request->getParsedBody();
+        if (!isset($data->priceInfo) || !is_array($data->priceInfo)) {
```
bertramakers, line 20: Nit: Since this class only gets executed after `JsonSchemaValidatingRequestBodyParser`, the `priceInfo` property should always be an array if present. So this if could be reduced to just:
```
if (!isset($data->priceInfo)) {
    return $request; 
}
```

(In some other RequestBodyParsers we also check the type and if other subproperties are set and the correct type etc because those are called _before_ `JsonSchemaValidatingRequestBodyParser` to transform old/legacy JSON request bodies to the ones expected by the JSON schema.)

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,41 @@
...
+        $priceInfo = $priceInfoDenormalizer->denormalize(
+            Json::decodeAssociatively(Json::encode($data->priceInfo)),
+            PriceInfo::class
+        );
+
+        if ($priceInfo->getTariffs()->hasDuplicates()) {
+            throw ApiProblem::bodyInvalidData(
+                new SchemaError(
+                    '/priceInfo',
+                    'Tariff names should be unique.'
```
bertramakers, line 34: Would be cool if we could have a more specific jsonPointer here that points to the tariff name that is a duplicate. (E.g. `/priceInfo/2/name` if the 3rd tariff's name is a duplicate of an earlier one)

Also nice would be to include multiple schemaErrors if there are multiple duplicates, so the integrator doesn't need to fix them one by one.

src/Http/Event/ImportEventRequestHandler.php
```diff
@@ -111,6 +112,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             new AgeRangeValidatingRequestBodyParser(),
             new CalendarValidatingRequestBodyParser(),
             new BookingInfoValidatingRequestBodyParser(),
+            new PriceInfoValidatingRequestBodyParser(),
```
bertramakers, line 115: @LucWollants @JonasVHG maybe we should start looking into combining some of these into a class like `OfferValidatingRequestBodyParser` so we don't have to maintain 2 lists of property validators (here + `ImportPlaceRequestHandler`). Can be in another PR later though.

src/Model/ValueObject/Price/Tariffs.php
```diff
@@ -5,11 +5,39 @@
...
+use CultuurNet\UDB3\Model\ValueObject\Translation\Language;
 
 class Tariffs extends Collection
 {
     public function __construct(Tariff ...$tariffs)
     {
         parent::__construct(...$tariffs);
     }
+
+    public function hasDuplicates(): bool
```
bertramakers, line 17:
```suggestion
    public function hasDuplicateNames(): bool
```
Small nit but this might be clearer, as a duplicate can also be interpreted as same name + same price + same currency.

bertramakers (commented):
Thanks @JonasVHG, overall approach looks great! 👌 Just some small suggestions.

src/Http/Event/ImportEventRequestHandler.php
LucWollants: Yes, that can certainly help!

tests/Http/Offer/PriceInfoValidatingRequestBodyParserTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $actual = (new PriceInfoValidatingRequestBodyParser())->parse($request)->getParsedBody();
+
+        $this->assertEquals($priceInfo, $actual);
+    }
+
+    /**
+     * @test
+     */
+    public function it_throws_on_same_names(): void
```
LucWollants, line 46: I don't see a throw here?

tests/Http/Event/ImportEventRequestHandlerTest.php
```diff
@@ -4392,6 +4392,104 @@ public function it_throws_if_priceInfo_has_no_main_language(): void
...
+                    'id' => '1.50.0.0.0',
+                ],
+            ],
+            'location' => [
+                '@id' => 'https://io.uitdatabank.dev/places/5cf42d51-3a4f-46f0-a8af-1cf672be8c84',
+            ],
+            'calendarType' => 'permanent',
+            'priceInfo' => [
+                [
+                    'category' => 'base',
```
LucWollants, line 4416: Maybe the base category should have another index to make sure that case is also handled?

src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -114,6 +115,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             new JsonSchemaValidatingRequestBodyParser(JsonSchemaLocator::PLACE),
             new CalendarValidatingRequestBodyParser(),
             new BookingInfoValidatingRequestBodyParser(),
+            new PriceInfoValidatingRequestBodyParser(),
```
LucWollants, line 118: This is not tested I think?

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /*
+     * @return SchemaError[]
+     */
+    private function getSchemaErrors(array $priceInfo): array
+    {
+        $errors = [];
+        $priceMatrix = [];
+
+        foreach ($priceInfo as $index => $price) {
```
LucWollants, line 41: Maybe change this to `foreach ($priceInfos as $index => $priceInfo)` for readability?

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,55 @@
...
+            foreach ($price['name'] as $lang => $priceLang) {
```
LucWollants, line 42: I think the correct naming is `foreach ($priceInfo['name'] as $language => $name)`

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
LucWollants: Should be fixed now

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
LucWollants: @JonasVHG this is maybe something that can still be done inside this pull request?

src/Model/ValueObject/Price/Tariffs.php
LucWollants: Method is no longer used.

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
LucWollants: This is changed.

src/Http/Event/ImportEventRequestHandler.php
LucWollants: @JonasVHG this is maybe something that can still be done inside this pull request?

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -33,20 +33,20 @@ public function parse(ServerRequestInterface $request): ServerRequestInterface
...
+    private function getSchemaErrors(array $priceInfos): array
     {
         $errors = [];
         $priceMatrix = [];
 
-        foreach ($priceInfo as $index => $price) {
-            foreach ($price['name'] as $lang => $priceLang) {
-                if (isset($priceMatrix[$lang]) && in_array($priceLang, $priceMatrix[$lang], true)) {
+        foreach ($priceInfos as $index => $priceInfo) {
+            foreach ($priceInfo['name'] as $language => $priceLang) {
```
LucWollants, line 42: I think you forgot to change `$priceLang` to `$name`?

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -33,20 +33,20 @@ public function parse(ServerRequestInterface $request): ServerRequestInterface
     /*
      * @return SchemaError[]
      */
-    private function getSchemaErrors(array $priceInfo): array
+    private function getSchemaErrors(array $priceInfos): array
     {
         $errors = [];
         $priceMatrix = [];
```
LucWollants, line 39: After changing the variable names, I wonder this is not a `$nameMatrix` instead of a `$priceMatrix`?

src/Http/Event/ImportEventRequestHandler.php
JonasVHG: @LucWollants , I've tried it but but I'm afraid that it causes too much code changes so I assume that it would be better in a different PR.

src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $errors = [];
+        $nameMatrix = [];
+
+        foreach ($priceInfos as $index => $priceInfo) {
+            foreach ($priceInfo['name'] as $language => $priceLang) {
+                if (isset($nameMatrix[$language]) && in_array($priceLang, $nameMatrix[$language], true)) {
+                    $errors[] = new SchemaError(
+                        '/priceInfo' . '/' . $index . '/name/' . $language,
+                        'Tariff name "' . $priceLang . '" should be unique.'
```
bertramakers, line 47:
```suggestion
                        'Tariff name "' . $priceLang . '" must be unique.'
```
Just a small nit but maybe clearer since SHOULD is often seen as a recommendation, and MUST as a requirement in technical docs.

tests/Http/Offer/PriceInfoValidatingRequestBodyParserTest.php
bertramakers: It does `$this->assertCallableThrowsApiProblem()` on the last line of the test. But maybe that was changed after your comment, I haven't reviewed the whole history of commits just the overall diff.

bertramakers (commented):
I leave it up to @LucWollants to approve after he has checked that all his comments have been resolved with the latest changes, but for me this looks great! Just added a small nitpick but nothing super important.

src/Http/Event/ImportEventRequestHandler.php
LucWollants: Done inside https://github.com/cultuurnet/udb3-silex/pull/1014

LucWollants (approved):

PR #999, opened 2022-05-04
III-4697 Fix bug that causes `labels` and/or `hiddenLabels` to become an object

bertramakers:
### Fixed

- Fixed bug that causes `labels` and/or `hiddenLabels` to become an object, by calling `array_values()` on the filtered array so the array becomes sequential instead of associative again.

---
Ticket: https://jira.uitdatabank.be/browse/III-4697


LucWollants:
Code looks good, can we also validate this in acceptance test as extra safety? But new resources wont need the polyfill fix I guess?

LucWollants (approved):

bertramakers:
> Code looks good, can we also validate this in acceptance test as extra safety? But new resources wont need the polyfill fix I guess?

Yeah for new resources it will never occur since they don't need this polyfill. And also the bug that this polyfill fixes was related to XML imports specifically, which we cannot do in the acceptance tests.

bertramakers:
But I added & pushed the unit tests first to clearly show that they were failing and that the bug is indeed located there 😃

JonasVHG (approved):

PR #1000, opened 2022-05-04
Hotfix/III-4661 fix for events without translation

JonasVHG:
### Fixed

- Fixed `sameAs` polyfill for events without translations

---
Ticket: https://jira.uitdatabank.be/browse/III-4661

### DIFF
--- a/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
+++ b/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
@@ -176,7 +176,7 @@ final class PropertyPolyfillOfferRepository extends DocumentRepositoryDecorator
 
         $urlParts = explode('/', $json['@id']);
         $id = array_pop($urlParts);
-        $name = $json['name']['nl'] ?? $json['name'];
+        $name = $json['name']['nl'] ?? ($json['name'][$json['mainLanguage']] ?? $json['name']);
 
         $json['sameAs'] = (new SameAsForUitInVlaanderen())->generateSameAs($id, $name);
 
--- a/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
+++ b/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
@@ -176,7 +176,7 @@ final class PropertyPolyfillOfferRepository extends DocumentRepositoryDecorator
 
         $urlParts = explode('/', $json['@id']);
         $id = array_pop($urlParts);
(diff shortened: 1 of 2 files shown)


src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
```diff
@@ -176,8 +176,9 @@ private function polyfillBrokenSameAs(array $json): array
 
         $urlParts = explode('/', $json['@id']);
         $id = array_pop($urlParts);
+        $name = $json['name']['nl'] ?? $json['name'];
```
LucWollants, line 179: I think there can be events with translated names, but for example only `en`? Should we try use the main language to define the language index. Only when the main language fails we use the name if it is a string.

bertramakers: I don't mind checking specifically for `nl` and using that if available, but if no `nl` name is found it will now use `$json['name']` which is not a string but an associative arrays with other keys like `fr`, `en`, ...

bertramakers: I also seem to remember that the name actually does not matter very much to make the link work and I've seen links that just use `x` as slug (since it's the id in the URL that actually determines what gets shown and the slug is just for SEO). So if there's no `nl` maybe we can just use that?

JonasVHG: I've changed the code the prefer Dutch(since UiTinVlaanderen is in Dutch). And provide fallbacks for events without Dutch translation. I'll test with UiV will do with an event that doesn't have a Dutch translation.

bertramakers: Still, the last fallback is `?? $json['name']` which is not a string. What you can do if there's no `nl` and for some weird reason the main language is also not set on `$json['name']` (which should not happen), is just grab the first value from the array. So something like `current($json['name'])`

JonasVHG: Correct, I was confused because in the past there were events without translations, but that was fixed with a previous replay.

bertramakers: Ah yep that should be fixed by now indeed

bertramakers (approved):
Looks perfect to me now, thanks for the extensive tests @JonasVHG !

PR #1002, opened 2022-05-05
III-4693 Support deleting an onlineUrl

LucWollants:
### Added
- Added `DeleteOnlineUrl` command
- Added `OnlineUrlDeleted` event
- Added `DeleteOnlineUrlHandler` to handle `DeleteOnlineUrl` command
- Added HTTP handler `DeleteOnlineUrlRequestHandler`

### Changed
- Extended `EventLDProjector` with apply method for `OnlineUrlDeleted`
- Extended `Event` aggregate with `deleteOnlineUrl` method
- Handled empty onlineUrl on import as a delete
- Delete `onlineUrl` when setting `attendanceMode` to `offline`

---
Ticket: https://jira.uitdatabank.be/browse/III-4693

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -7,6 +7,7 @@ namespace CultuurNet\UDB3\Event;
 use Broadway\EventSourcing\EventSourcedAggregateRoot;
 use CultuurNet\UDB3\Event\Events\AttendanceModeUpdated;
 use CultuurNet\UDB3\Event\Events\AvailableFromUpdated;
+use CultuurNet\UDB3\Event\Events\OnlineUrlDeleted;
 use CultuurNet\UDB3\Event\Events\OnlineUrlUpdated;
 use CultuurNet\UDB3\Event\Events\ThemeRemoved;
 use CultuurNet\UDB3\Event\Events\VideoAdded;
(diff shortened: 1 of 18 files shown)


src/Event/Event.php
```diff
@@ -383,6 +384,10 @@ public function updateAttendanceMode(AttendanceMode $attendanceMode): void
         if ($this->attendanceMode !== $attendanceMode->toString()) {
             $this->apply(new AttendanceModeUpdated($this->eventId, $attendanceMode->toString()));
         }
+
+        if (!empty($this->onlineUrl) && $this->attendanceMode === AttendanceMode::offline()->toString()) {
+            $this->apply(new OnlineUrlDeleted($this->eventId));
+        }
```
bertramakers, line 390: Nit: You could drop the `!empty($this->onlineUrl)` and call `$this->deleteOnlineUrl()` instead of `$this->apply(new OnlineUrlDeleted($this->eventId))`

bertramakers, line 390: Nit: You could drop the `!empty($this->onlineUrl)` and call `$this->deleteOnlineUrl()` instead of `$this->apply(new OnlineUrlDeleted($this->eventId))`

bertramakers (approved):

src/Event/Event.php
LucWollants: I intentionally never call mutator methods inside other mutator methods to keep them loosely coupled. But it indeed also works like you propose.

LucWollants: This seems to be the same comment?

PR #1004, opened 2022-05-06
III-4648 Polyfill typicalAgeRange for events

LucWollants:
### Added
- Polyfill `typicalAgeRange` on events and places
- Always project `typicalAgeRange` so when executing a replay the polyfill can be removed

### Changed
- Extend `PropertyPolyfillOfferRepository` with `OfferType`
- Change `XMLImporter` to set all ages by default
- Only polyfill `attendanceMode` on events

---
Ticket: https://jira.uitdatabank.be/browse/III-4648

### DIFF
--- a/src/Event/ReadModel/JSONLD/EventLDProjector.php
+++ b/src/Event/ReadModel/JSONLD/EventLDProjector.php
@@ -275,6 +275,8 @@ class EventLDProjector extends OfferLDProjector implements
 
         $jsonLD->attendanceMode = AttendanceMode::offline()->toString();
 
+        $jsonLD->typicalAgeRange = '-';
+
         $defaultAudience = new Audience(AudienceType::everyone());
         $jsonLD->audience = $defaultAudience->serialize();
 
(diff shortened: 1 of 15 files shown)


src/Event/ReadModel/JSONLD/EventLDProjector.php
```diff
@@ -275,6 +275,8 @@ protected function applyEventCreated(
 
         $jsonLD->attendanceMode = AttendanceMode::offline()->toString();
 
+        $jsonLD->typicalAgeRange = '-';
```
bertramakers, line 278: It should also be set for events imported/updated from UDB2 I think?

bertramakers: \+ Places imported/updated from UDB2 then

LucWollants: Good point, I modified the XMLImporter to set all ages by default

bertramakers: Looks good, I only think that places created via `PlaceImportedFromUDB2` should also get `"typicalAgeRange": "-"`?
It seems like the `CultuurNet\UDB3\Place\ReadModel\JSONLD\CdbXMLImporter` does not set any typicalAgeRange at first glance, probably because Actors did not have age info in UDB2. But we still want every place to have a `typicalAgeRange`, so we should also set the default for those.

LucWollants: There is also an importer for places, I missed that one 🤦‍♂️

bertramakers (approved):

PR #1006, opened 2022-05-09
III-4689 Cannot publish approved offer

bertramakers:
### Fixed

- When trying to publish an offer that is already approved, no error gets returned. Only for rejected or deleted.

---
Ticket: https://jira.uitdatabank.be/browse/III-4689


LucWollants (approved):
Very clean code. The `ConvertsToApiProblem` is a very useful interface.

PR #1007, opened 2022-05-09
III-4705 Fix term validation for imports when no domain is set

bertramakers:
### Fixed

- Fixed term validation in imports when no term `domain` is set, while still ignoring the legacy domains to avoid errors when an integrator sends us our own JSON-LD (with some modifications) that might still contain legacy domains

---
Ticket: https://jira.uitdatabank.be/browse/III-4705


bertramakers:
For more info see my comment in the ticket

LucWollants (approved):
Very clear!

PR #1009, opened 2022-05-09
III-4709 Make sure ConvertsToApiProblem exceptions do not get logged in Sentry

bertramakers:
### Fixed

- `ConvertsToApiProblem` exceptions do not get logged to Sentry anymore, except if the status code of their ApiProblem is not 4XX.
- `ApiProblem` exceptions still do not get logged to Sentry (like before), except if their status code is not 4XX.

---
Ticket: https://jira.uitdatabank.be/browse/III-4709


bertramakers:
I left the comments for now because I myself had to think for a second about what `isBadRequestException` now implied exactly 😄

PR #1017, opened 2022-05-16
III-4700 Change Publish command permission to aanbodBewerken

LucWollants:
### Fixed
- Changed `Publish` command permission to `aanbodBewerken`

---
Ticket: https://jira.uitdatabank.be/browse/III-4700

### DIFF
--- a/tests/Event/Commands/Moderation/PublishTest.php
+++ b/tests/Event/Commands/Moderation/PublishTest.php
@@ -11,6 +11,7 @@ final class PublishTest extends TestCase
 {
     /**
      * @test
+     * @bugfix https://jira.uitdatabank.be/browse/III-4700
      */
     public function it_has_permission_aanbodBewerken(): void
     {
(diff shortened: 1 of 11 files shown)


tests/Event/Commands/Moderation/PublishTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace CultuurNet\UDB3\Event\Commands\Moderation;
+
+use CultuurNet\UDB3\Role\ValueObjects\Permission;
+use PHPUnit\Framework\TestCase;
+
+final class PublishTest extends TestCase
+{
+    /**
+     * @test
```
bertramakers, line 13: Maybe reference that this is a `@bugfix` and add a `@see` link to the ticket so we don't forget why this was overridden later.

bertramakers, line 13: Maybe reference that this is a `@bugfix` and add a `@see` link to the ticket so we don't forget why this was overridden later.

LucWollants: Screenshot 2022-05-17 at 15 36 23

bertramakers: Sure! 👍

PR #1020, opened 2022-05-18
III-4732 Use `url/not-found` for all 404s

bertramakers:
### Changed

- `ApiProblem::invalidQueryParameter()` to use `urlNotFound()` internally
- Renamed `UpdateLocationNotSupported` to `AttendanceModeNotSupported` and gave it a dedicated custom error type with a 400 status code, as it's a domain error and not a URL error (404) (Also had to change the details a bit to make them better fit the error type "attendance-mode-not-supported")

### Removed

- `ApiProblem::invalidPathParameter()` and replaced usages with `urlNotFound()`
- `ApiProblem::userNotFound()` and replaced usages with `urlNotFound()` (This custom error type was not supported and should be replaced with "URL not found" in this case anyway.)
- `ApiProblem::invalidEmailAddress()` and replaced usages with `urlNotFound()` (This custom error type was not supported and should be replaced with "URL not found" in this case anyway.)
- `ApiProblem::tokenNotSupported()` and replaced usages with `unauthorized()` (This custom error type was not supported and could be replaced with "Unauthorized" in these cases anyway.)

---
Ticket: https://jira.uitdatabank.be/browse/III-4732


src/Event/Event.php
```diff
@@ -312,7 +312,7 @@ public function updateLocation(LocationId $locationId): void
 
         if (!$locationId->isNilLocation() && $this->attendanceMode === AttendanceMode::online()->toString()) {
             throw new UpdateLocationNotSupported(
-                'Instead of passing the real location for this online event, please update the attendance mode to offline or mixed.'
+                'Cannot update the location of an online event to a real location. Set the attendanceMode to mixed or offline first.'
```
LucWollants, line 315: Maybe `real location` should be changed to `physical location`. I think I have seen that used inside the API docs.

LucWollants (approved):
Very easy to review because of the granular and well-commented commits.
Furthermore a consistent functional use of 404 HTTP errors.

src/Event/Event.php
bertramakers: Yeah I made the same remark in another PR now that I think about it, not sure where though. But doesn't matter, physical location is probably clearer because "real" can mean many things, could also be some kind of property on the API for all the integrator knows.

PR #1024, opened 2022-05-18
III-4726 block update duplicate tarrif

JonasVHG:
### Added

- Added validation for updates of PriceInco

---
Ticket: https://jira.uitdatabank.be/browse/III-4726

### DIFF
--- a/src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
+++ b/src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
@@ -16,12 +16,14 @@ class PriceInfoValidatingRequestBodyParser implements RequestBodyParser
     {
         $data = $request->getParsedBody();
 
-        if (!isset($data->priceInfo)) {
+        if (!isset($data->priceInfo) && !is_array($data)) {
             return $request;
         }
 
+        $priceInfo = $data->priceInfo ?? $data;
+
         $errors = $this->getSchemaErrors(
-            Json::decodeAssociatively(Json::encode($data->priceInfo))
+            Json::decodeAssociatively(Json::encode($priceInfo))
         );
 
         if (count($errors) > 0) {
--- a/src/Http/Offer/UpdatePriceInfoRequestHandler.php
+++ b/src/Http/Offer/UpdatePriceInfoRequestHandler.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Offer;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Event\Commands\UpdatePriceInfo as EventUpdatePriceInfo;
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Http\Request\Body\DenormalizingRequestBodyParser;
+use CultuurNet\UDB3\Http\Request\Body\RequestBodyParserFactory;
+use CultuurNet\UDB3\Http\Request\RouteParameters;
+use CultuurNet\UDB3\Http\Response\NoContentResponse;
+use CultuurNet\UDB3\Model\Serializer\ValueObject\Price\PriceInfoDenormalizer;
+use CultuurNet\UDB3\Model\ValueObject\Price\PriceInfo;
+use CultuurNet\UDB3\Offer\OfferType;
+use CultuurNet\UDB3\Place\Commands\UpdatePriceInfo as PlaceUpdatePriceInfo;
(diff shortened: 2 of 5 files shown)


src/Http/Offer/PriceInfoValidatingRequestBodyParser.php
```diff
@@ -16,12 +16,14 @@ public function parse(ServerRequestInterface $request): ServerRequestInterface
     {
         $data = $request->getParsedBody();
 
-        if (!isset($data->priceInfo)) {
+        if (!isset($data->priceInfo) && !is_array($data)) {
```
LucWollants, line 19: I think this requires a comment to indicate that either a full event or place is passed with the price info as property or either the price info as the root object.

src/Http/Offer/UpdatePriceInfoRequestHandler.php
```diff
@@ -0,0 +1,76 @@
...
+                );
+            } else {
+                $updatePriceInfo = new PlaceUpdatePriceInfo(
+                    $offerId,
+                    LegacyPriceInfo::fromUdb3ModelPriceInfo($priceInfo)
+                );
+            }
+            $this->commandBus->dispatch($updatePriceInfo);
+        } catch (ApiProblem $apiProblem) {
+            throw $apiProblem;
```
LucWollants, line 71: I don't think it makes sense to catch an `ApiProblem` and then throw it again.

src/Http/Offer/UpdatePriceInfoRequestHandler.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $parser = RequestBodyParserFactory::createBaseParser(
+            new JsonSchemaValidatingRequestBodyParser(
+                JsonSchemaLocator::getSchemaFileByOfferType(
+                    $offerType,
+                    JsonSchemaLocator::EVENT_PRICE_INFO_PUT,
+                    JsonSchemaLocator::PLACE_PRICE_INFO_PUT
+                )
+            ),
+            new PriceInfoValidatingRequestBodyParser(),
```
LucWollants, line 48: Now that I see this in a bigger context I notice that `PriceInfoValidatingRequestBodyParser` is not a good name.
Because this is just a part of the validation, namely checking for duplicate names.
The other validation rules are checked with `JsonSchemaValidatingRequestBodyParser`

tests/Http/Offer/UpdatePriceInfoRequestHandlerTest.php
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * @test
+     * @dataProvider offerTypeDataProvider
+     */
+    public function it_dispatches_an_update_command_if_no_duplicate_names_are_found(string $offerType): void
+    {
+        $request = (new Psr7RequestBuilder())
+            ->withRouteParameter('offerType', $offerType)
+            ->withRouteParameter('offerId', 'a91bc028-c44a-4429-9784-8641c9858eed')
+            ->withBodyFromString('[{"category":"base","name":{"nl":"Basistarief"},"price":10,"priceCurrency":"EUR"},{"category":"tariff","name":{"nl":"Jongeren"},"price":5,"priceCurrency":"EUR"},{"category":"tariff","name":{"nl":"Senioren"},"price":5,"priceCurrency":"EUR"}]')
```
LucWollants, line 51: This is not readable nor maintanable. It should be created from an array.

tests/Http/Offer/UpdatePriceInfoRequestHandlerTest.php
```diff
@@ -0,0 +1,125 @@
...
+use CultuurNet\UDB3\PriceInfo\PriceInfo;
+use CultuurNet\UDB3\PriceInfo\Tariff;
+use CultuurNet\UDB3\StringLiteral;
+use CultuurNet\UDB3\ValueObject\MultilingualString;
+use Money\Currency;
+use Money\Money;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class UpdatePriceInfoRequestHandlerTest extends TestCase
```
LucWollants, line 25: This file is missing a lot of tests to validate the price info. All the validation rules from `PriceInfoDataValidator` that are validated here `PriceInfoDataValidatorTest` should now be verified here.

A good example to see tests for the validation of new request handlers is `UpdateCalendarRequestHandlerTest`

When the endpoint is using the new approach `PriceInfoDataValidator` can even be removed.

LucWollants (approved):
The new file `UpdatePriceInfoRequestHandler` is not yet used. So I will already merge this PR.

PR #1025, opened 2022-05-18
III-4268 refactor PriceInfo

JonasVHG:
### Added

- new Common `UpdatePrice` for both Place & Event

### Removed

- The separate `UpdatePrice` for Place & Event

### Changed

- The `EventHasTicketSalesCommandValidator` to know the difference for an `UpdatePrice` on an `Event` and on a `Place`

---
Ticket: https://jira.uitdatabank.be/browse/III-4268


src/Offer/Commands/UpdatePriceInfo.php
```diff
@@ -0,0 +1,23 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\Commands;
+
+use CultuurNet\UDB3\PriceInfo\PriceInfo;
+
+final class UpdatePriceInfo extends AbstractCommand
+{
+    protected PriceInfo $priceInfo;
```
LucWollants, line 11: I wonder if this could not just be using `CultuurNet\UDB3\Model\ValueObject\Price\PriceInfo`
That can also be used inside the new command handler where it gets converted to the old format before sending it to the aggregate.

app/UiTPAS/UiTPASCommandValidatorServiceProvider.php
```diff
@@ -19,7 +19,8 @@ public function register(Application $app)
             function (Application $app) {
                 return new EventHasTicketSalesCommandValidator(
                     $app['uitpas'],
-                    LoggerFactory::create($app, LoggerName::forService('uitpas', 'ticket-sales'))
+                    LoggerFactory::create($app, LoggerName::forService('uitpas', 'ticket-sales')),
+                    $app['place_repository']
```
LucWollants, line 23: I would always make the order of arguments depending on the "importance" of a parameter and not add newer parameters just to the end.

src/Offer/CommandHandlers/UpdatePriceInfoHandler.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public function handle($command): void
+    {
+        if (!($command instanceof UpdatePriceInfo)) {
+            return;
+        }
+
+        $offer = $this->offerRepository->load($command->getItemId());
+        $offer->updatePriceInfo(
```
LucWollants, line 27: Here a conversion could happen from the Model PriceInfo to the Legacy PriceInfo

src/Offer/DefaultOfferEditingService.php
```diff
@@ -252,7 +253,7 @@ public function updatePriceInfo(string $id, PriceInfo $priceInfo)
         $this->guardId($id);
 
         return $this->commandBus->dispatch(
```
LucWollants, line 255: I don't think this dispatch is needed? Even the method `updatePriceInfo` can be removed from the `OfferEditingServiceInterface` interface.

src/UiTPAS/Event/CommandHandling/Validation/EventHasTicketSalesCommandValidator.php
```diff
@@ -43,6 +49,13 @@ public function validate($command)
 
         $eventId = $command->getItemId();
 
+        try {
+            $this->placeRepository->load($eventId);
```
LucWollants, line 53: This is now a double negotiation, so maybe it's better to inject the event repository?
When the event is not found a return can be done.

src/Offer/CommandHandlers/UpdatePriceInfoHandler.php
JonasVHG: Same as previous comment, I think we shouldn't make this PR too big.

src/Offer/Commands/UpdatePriceInfo.php
JonasVHG: Having started on this, I think this would be best handled in another PR as to not make PR's too big.

PR #1026, opened 2022-05-23
III-4727 Finetune command to update attendance mode of events

LucWollants:
### Changed
- Ask for physical location UUID when changing attendance mode to offline
- Set nil location when changing attendance mode to online

---
Ticket: https://jira.uitdatabank.be/browse/III-4727

### DIFF
--- a/app/Console/UpdateEventsAttendanceMode.php
+++ b/app/Console/UpdateEventsAttendanceMode.php
@@ -6,6 +6,8 @@ namespace CultuurNet\UDB3\Silex\Console;
 
 use Broadway\CommandHandling\CommandBus;
 use CultuurNet\UDB3\Event\Commands\UpdateAttendanceMode;
+use CultuurNet\UDB3\Event\Commands\UpdateLocation;
+use CultuurNet\UDB3\Event\ValueObjects\LocationId;
 use CultuurNet\UDB3\Model\ValueObject\Online\AttendanceMode;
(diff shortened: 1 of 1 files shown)


app/Console/UpdateEventsAttendanceMode.php
```diff
@@ -45,24 +47,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $query = $this->askForQuery($input, $output);
         $attendanceMode = $this->askAttendanceMode($input, $output);
 
+        $locationId = new LocationId(LocationId::NIL_LOCATION);
+        if (!$attendanceMode->sameAs(AttendanceMode::online())) {
+            $locationId = $this->askForPhysicalLocation($input, $output);
```
JonasVHG, line 52: Wouldn't it be prudent to prevent users from giving the online uuid `00000000-0000-0000-0000-000000000000`? Now this command is able to create events with "attendanceMode": "offline" and "locationId": "00000000-0000-0000-0000-000000000000".

LucWollants: Good remark! That is indeed guarded inside the request handler https://github.com/cultuurnet/udb3-silex/blob/814416b08c4980f554b2eea3fde4762d132c94a4/src/Http/Event/UpdateAttendanceModeRequestHandler.php#L78 but not inside the CLI command

LucWollants: This is fixed now.

PR #1036, opened 2022-05-31
III-4751 Allow copyright property to be empty when serializing ImageUpdated

LucWollants:
### Changed
- Converted `CopyrightHolder` property to a native `string` to allow empty values when serializing `ImageUpdated` from the event store
- Modified `description` property to a native `string`

---
Ticket: https://jira.uitdatabank.be/browse/III-4751


bertramakers (approved):
🐛🔨

PR #1038, opened 2022-05-31
III-4718 Fix "document not found" errors in Sentry and logs when updating label visibility

bertramakers:
### Fixed

- Fixed 2 errors being logged in Sentry for every related document of a label that gets its visibility updated, caused by the ItemVisibilityProjector not taking the RelationType into account and always trying to update every related document id both as an event, place, and organizer (of which only 1 exists for the given id).

---
Ticket: https://jira.uitdatabank.be/browse/III-4718


LucWollants (approved):
Super clan code and very easy to review because of the small and comprehensive commits.

PR #1040, opened 2022-06-02
III-4743 MainImage not updated when Removing an UDB2 image from offer

JonasVHG:
### Fixed

- The `Offer` gets a new `MainImage` when it has an old UDB2 image as `MainImage` and it is removed.

---
Ticket: https://jira.uitdatabank.be/browse/III-4743

### DIFF
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -422,7 +422,7 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
         );
 
         // Unset the main image if it matches the removed image
-        if (isset($offerLd->image) && strpos($offerLd->{'image'}, $imageId)) {
+        if (isset($offerLd->image)) {
             unset($offerLd->{'image'});
         }
 
--- a/tests/Offer/ReadModel/JSONLD/OfferLDProjectorTest.php
+++ b/tests/Offer/ReadModel/JSONLD/OfferLDProjectorTest.php
@@ -881,6 +881,56 @@ class OfferLDProjectorTest extends TestCase
         );
     }
 
+    /**
+     * @test
+     */
+    public function it_should_keep_the_udb2_image_main_when_deleting_a_new_one(): void
+    {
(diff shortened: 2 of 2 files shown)


src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -422,7 +422,7 @@ protected function applyImageRemoved(AbstractImageRemoved $imageRemoved): ?JsonD
         );
 
         // Unset the main image if it matches the removed image
-        if (isset($offerLd->image) && strpos($offerLd->{'image'}, $imageId)) {
+        if (isset($offerLd->image)) {
```
LucWollants, line 425: By doing the `unset` of the main image always I think that the main image will be reset to the first image. Even when the main image was not removed.

I think this can be tested by having three images, the main image being the third image. Then a removal of the second image will result in the first image becoming the main image. But this should stay the last image.

tests/Offer/ReadModel/JSONLD/OfferLDProjectorTest.php
```diff
@@ -830,6 +830,169 @@ public function it_should_make_the_oldest_image_main_when_deleting_the_current_m
...
+            'https://foo.bar/media/5ae74e68-20a3-4cb1-b255-8e405aa01ab9.png',
+            $eventBody->image
+        );
+    }
+
+
+    /**
+     * @test
+     */
+    public function it_should_keep_the_main_image_when_deleting_a_previous_in_a_list(): void
```
LucWollants, line 888: Maybe rephrase the last part to `when_deleting_an_image_with_a_lower_index`?

LucWollants (approved):
Looks good now!

I think it's worth investigating how much effort it is to also write an acceptance test for this.

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
JonasVHG: This is now fixed (for both udb2 & udb3 images) by checking the url instead of the id.

PR #1042, opened 2022-06-03
III-4716 remove labels with newlines in projection

JonasVHG:
### Changed

- replace labels with newline in in `AbstractLabelsImported` & `AbstractLabelsEvent`

### Added

- Remove labels with newlines in the JSON projection

---
Ticket: https://jira.uitdatabank.be/browse/III-4716

### DIFF
--- a/src/Offer/Events/AbstractLabelEvent.php
+++ b/src/Offer/Events/AbstractLabelEvent.php
@@ -46,7 +46,13 @@ abstract class AbstractLabelEvent extends AbstractEvent implements LabelEventInt
 
         return new static(
             $data['item_id'],
-            new Label($data['label'], $data['visibility'])
+            new Label(self::trimLineEndings($data['label']), $data['visibility'])
         );
+    }
+
+    private static function trimLineEndings(string $labelName): string
+    {
+        $labelName = preg_replace('/\\r\\n/', ' ', $labelName);
+        return preg_replace('/\\n/', ' ', $labelName);
     }
 }
--- a/src/Offer/Events/AbstractLabelsImported.php
+++ b/src/Offer/Events/AbstractLabelsImported.php
@@ -34,7 +34,7 @@ abstract class AbstractLabelsImported extends AbstractEvent implements LabelsImp
         $labels = new Labels();
         foreach ($data['labels'] as $label) {
             $labels = $labels->with(new Label(
-                new LabelName($label['label']),
+                new LabelName(self::trimLineEndings($label['label'])),
                 $label['visibility']
             ));
         }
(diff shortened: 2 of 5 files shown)


src/Offer/Events/AbstractLabelEvent.php
```diff
@@ -46,7 +46,13 @@ public static function deserialize(array $data): AbstractLabelEvent
 
         return new static(
             $data['item_id'],
-            new Label($data['label'], $data['visibility'])
+            new Label(self::trimLineEndings($data['label']), $data['visibility'])
         );
     }
+
+    private static function trimLineEndings(string $labelName): string
```
LucWollants, line 53: Maybe `trim` is not the best name, because a `trim` in PHP code typically manipulates the start and/or end of a string.

src/Offer/Events/AbstractLabelsImported.php
```diff
@@ -34,7 +34,7 @@ public static function deserialize(array $data): AbstractLabelsImported
         $labels = new Labels();
         foreach ($data['labels'] as $label) {
             $labels = $labels->with(new Label(
-                new LabelName($label['label']),
+                new LabelName(self::trimLineEndings($label['label'])),
```
LucWollants, line 37: I don't think this is tested?

src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
```diff
@@ -241,6 +241,29 @@ function (array $json) {
...
+    private function removeNewLineLabels(JsonDocument $jsonDocument): JsonDocument
+    {
+        return $jsonDocument->applyAssoc(
+            function (array $json) {
+                $removeNewLineLabels = static function (array $json, string $propertyName): array {
+                    if (!isset($json[$propertyName]) || !is_array($json[$propertyName])) {
+                        return $json;
+                    }
+                    $json[$propertyName] = array_values(
+                        array_filter($json[$propertyName], fn ($label) => !str_contains($label, '\\n'))
```
LucWollants, line 253: Maybe it is also possible that the string contains an Apple Style newline `\r`?

LucWollants (changes requested):
The overall approach is fine, but the code to check for newlines is scattered over several files and implemented in several ways.

Maybe it's worth checking this regex `preg_replace('/\s+/', ' ', $string)` ?
See https://stackoverflow.com/questions/3760816/remove-new-lines-from-string-and-replace-with-one-empty-space

src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
bertramakers: `\r` was only used before Mac OS X though. `\r\n` is possible of course since that is still used on Windows, but that also contains `\n`.

LucWollants:
> To be able to do that, the labels should still contain the newlines in the events from the event store.

I think we need to brainstorm a bit more about the best approach then. Because now almost all label related events derive from `AbstractLabelEvent`. The event `AbstractLabelEvent` stores a `Label`. The label contains both the `LabelName` and `visibility`.

If we would change `AbstractLabelEvent` to use native types there will be suddenly two properties to maintain inside `AbstractLabelEvent`. A `string` for the name and a `bool` for the visibility. That change requires a big amount of code changes.

PR #1044, opened 2022-06-09
III-4783 refactor regex labelname

JonasVHG:
### Changed

- Changed the regex from `LabelName` to accept newlines in labels.

---
Ticket: https://jira.uitdatabank.be/browse/III-4783

### DIFF
--- a/tests/Model/ValueObject/Taxonomy/Label/LabelNameTest.php
+++ b/tests/Model/ValueObject/Taxonomy/Label/LabelNameTest.php
@@ -67,6 +67,18 @@ class LabelNameTest extends TestCase
                 '--',
                 true,
             ],
+            [
+                PHP_EOL,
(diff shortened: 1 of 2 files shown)


bertramakers (approved):

tests/Model/ValueObject/Taxonomy/Label/LabelNameTest.php
```diff
@@ -67,6 +67,22 @@ public function labelNameDataProvider(): array
                 '--',
                 true,
             ],
+            [
+                PHP_EOL,
```
LucWollants, line 71: I wonder if this should not just be `\r\n` or `\n` as a string because `PHP_EOL` depends on the OS.

LucWollants (approved):

PR #1046, opened 2022-06-10
III-4778 Remove mark as duplicate code

LucWollants:
### Removed
- Removed `MarkPlaceAsDuplicateCommand` CLI command
- Removed  `MarkPlacesAsDuplicateFromTableCommand` CLI command
- Removed `DispatchMarkedAsDuplicateEventCommand` CLI command
- Removed `LocationMarkedAsDuplicateProcessManager` process manager
- Removed `MarkAsDuplicateCommandHandler` command handler
- Removed the `Place::markAsDuplicateOf` aggregate method
- Removed the `Place::markAsCanonicalFor` aggregate method
- Removed the `MarkAsDuplicate` command

### Changed
- Avoid processing `MarkedAsDuplicate` event
- Avoid processing `MarkedAsCanonical` event

---
Ticket: https://jira.uitdatabank.be/browse/III-4778


src/Place/ReadModel/History/HistoryProjector.php
```diff
@@ -137,12 +135,6 @@ public function handle(DomainMessage $domainMessage): void
             case $event instanceof MajorInfoUpdated:
                 $this->projectMajorInfoUpdated($domainMessage);
                 break;
-            case $event instanceof MarkedAsCanonical:
-                $this->projectMarkedAsCanonical($domainMessage);
-                break;
```
JonasVHG, line 142: Open question(since I'm not really familiar with this part of the code). But since the eventstore already has MarkAsDuplicate and MarkAsCanonical Events, would it be safe to remove them here?

LucWollants: It is safe to remove the handling of `MarkedAsDuplicate` and `MarkedAsCanonical` in every event listener. It is even safe to remove the `apply` method from an aggregate. The only thing that is still required to be around are the events.

You can check that on `EventSourcedAggregateRoot::handle`. If the method does not exist, it is not handled.

bertramakers (approved):

PR #1049, opened 2022-06-13
III-4784 Refactor Labelevents

JonasVHG:
### Changed

- Refactor Label Events

---
Ticket: https://jira.uitdatabank.be/browse/III-4784

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -500,12 +500,12 @@ class Event extends Offer implements UpdateableWithCdbXmlInterface
 
     protected function createLabelAddedEvent(LegacyLabel $label): LabelAdded
     {
-        return new LabelAdded($this->eventId, $label);
+        return new LabelAdded($this->eventId, $label->getName()->toNative(), $label->isVisible());
     }
 
     protected function createLabelRemovedEvent(LegacyLabel $label): LabelRemoved
     {
-        return new LabelRemoved($this->eventId, $label);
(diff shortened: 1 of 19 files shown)


LucWollants (changes requested):
I think this will still crash because the method `AbstractLabelsImported:: deserialize` is used to read an event from the event store and and that still uses `LabelName`. The proposed approach is to just store 2 arrays of strings. One array with visible labels and one array with invisible labels.

LucWollants:
There could still be an issue with the internal state of the Offer aggregate. This aggregate stores the labels as a legacy `LabelCollection` which for now supports new lines. But once the legacy `LabelCollection` will be replaced the internal state has to work with primitive types to support new lines.

src/Organizer/Events/LabelsImported.php
```diff
@@ -53,11 +64,16 @@ public static function deserialize(array $data): LabelsImported
     public function serialize(): array
```
LucWollants, line 64: I don't think it's needed to change this method. The internal property is still a VO `Labels`. You can also directly access this property and don't need to use the `getLabels` accessor method. So I reverted this change.

LucWollants:
This now has 2 extra PRs:
- https://github.com/cultuurnet/udb3-silex/pull/1052 (Should be ready)
- https://github.com/cultuurnet/udb3-silex/pull/1051 (Still work in progress)

PR #1050, opened 2022-06-14
III-4779 Filter out excluded labels when asking for suggestions

LucWollants:
### Changed
- Filtered out excluded labels when asking for suggestions

---
**Note**: the initial approach was to create a table with excluded labels; that table would be used for a subquery in a `WHERE NOT IN` clause. But it turns out that Doctrine DBAL can also handle a simple array as an input parameter for the `notIn` expression. For now, I just used the existing `InMemoryExcludedLabelsRepository`.

---

Ticket: https://jira.uitdatabank.be/browse/III-4779

### DIFF
--- a/src/Label/ReadModels/JSON/Repository/Doctrine/DBALReadRepository.php
+++ b/src/Label/ReadModels/JSON/Repository/Doctrine/DBALReadRepository.php
@@ -152,6 +152,19 @@ final class DBALReadRepository extends AbstractDBALRepository implements ReadRep
 
         if ($query->isSuggestion()) {
             $queryBuilder->andWhere('name REGEXP \'^[a-zA-Z\d_\-]{2,50}$\'');
+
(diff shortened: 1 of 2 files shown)


src/Label/ReadModels/JSON/Repository/Doctrine/DBALReadRepository.php
```diff
@@ -152,6 +152,19 @@ private function createSearchQuery(Query $query): QueryBuilder
...
+                $queryBuilder->andWhere(
+                    $queryBuilder->expr()->notIn(
+                        SchemaConfigurator::UUID_COLUMN,
+                        array_map(
+                            fn (string $label) => '"' . $label . '"',
+                            $excludedLabels
+                        )
+                    )
+                );
+            }
```
bertramakers, line 167: Smart! 👍 💯

bertramakers (approved):

JonasVHG (approved):

PR #1054, opened 2022-06-16
III-4613 Add script to remove facilities from places

LucWollants:
### Added
- Added CLI command `RemoveFacilitiesFromPlace` to remove the facilities from a single place or a list of places matching the query `terms.id:3.`

### Changed
- Prevent updating the facilities with an empty list if the offer has no facilities. This will prevent a redundant and empty `FacilitiesUpdated` inside the event store

---
Ticket: https://jira.uitdatabank.be/browse/III-4613

### DIFF
--- a/app/Console/RemoveFacilitiesFromPlace.php
+++ b/app/Console/RemoveFacilitiesFromPlace.php
@@ -16,7 +16,7 @@ use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 final class RemoveFacilitiesFromPlace extends AbstractCommand
 {
-    private const FACILITY_QUERY = 'terms.id:3.';
+    private const FACILITY_QUERY = 'terms.id:3.*';
 
     private ResultsGeneratorInterface $searchResultsGenerator;
 
--- a/app/Console/RemoveFacilitiesFromPlace.php
+++ b/app/Console/RemoveFacilitiesFromPlace.php
@@ -1,0 +1,93 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


JonasVHG (approved):

app/Console/RemoveFacilitiesFromPlace.php
```diff
@@ -0,0 +1,93 @@
...
+use CultuurNet\UDB3\Search\ResultsGeneratorInterface;
+use CultuurNet\UDB3\Search\SearchServiceInterface;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+final class RemoveFacilitiesFromPlace extends AbstractCommand
+{
+    private const FACILITY_QUERY = 'terms.id:3.';
```
bertramakers, line 19: Are you sure this will work? I get 0 results on production for this `q`, even with `disableDefaultFilters=true`: 

https://search.uitdatabank.be/places/?disableDefaultFilters=true&q=terms.id:3.

bertramakers: It does work if you add a wildcard at the end it seems: [https://search.uitdatabank.be/places/?disableDefaultFilters=true&q=terms.id:3.*](https://search.uitdatabank.be/places/?disableDefaultFilters=true&q=terms.id:3.*)

LucWollants: Indeed! I forgot the wildcard.

PR #1055, opened 2022-06-17
III-4773 Only validate ticket sales for update and delete of organizer

LucWollants:
### Added
- Added new `DeleteOrganizer` offer command
- Added new `UpdateOrganizer` offer command
- Added new `ChangeNotAllowedByTicketSales` exception
- Added new `DeleteOrganizerHandler` command handler
- Added new `UpdateOrganizerHandler` command handler
- **Added new `EventHasTicketSalesGuard` guard to also take actual change into account**

### Changed
- Changed `Offer` to expose `organizerId`

### Removed
- Removed `EventHasTicketSalesCommandValidator` and related classes

---
Ticket: https://jira.uitdatabank.be/browse/III-4773


JonasVHG (approved):

src/Offer/Commands/DeleteOrganizer.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    private string $organizerId;
+
+    public function __construct(string $itemId, string $organizerId)
+    {
+        parent::__construct($itemId);
+        $this->organizerId = $organizerId;
+    }
+
+    public function getOrganizerId(): string
```
bertramakers, line 17: Does `DeleteOrganizer` even need the organizer id? It could just delete the current organizer relationship?

src/UiTPAS/Validation/EventHasTicketSalesGuard.php
```diff
@@ -0,0 +1,80 @@
...
+            // The command is for a place which can not have ticket sales.
+            return;
+        }
+
+        $equalOrganizers = $event->getOrganizerId() === $command->getOrganizerId();
+        if ($command instanceof UpdateOrganizer && $equalOrganizers) {
+            // A change only happens when a different organizer is updated
+            return;
+        }
+        if ($command instanceof DeleteOrganizer && !$equalOrganizers) {
```
bertramakers, line 56: If `DeleteOrganizer` would not have a specific organizer id, you could just check if `$event->getOrganizerId()` is null or not (if not null it's a change).

src/UiTPAS/Validation/EventHasTicketSalesGuard.php
```diff
@@ -0,0 +1,80 @@
...
+            // A change only happens when the existing organizer is deleted
+            return;
+        }
+
+        try {
+            $hasTicketSales = $this->uitpas->eventHasTicketSales($eventId);
+        } catch (\Exception $exception) {
```
bertramakers, line 63: I think if we catch `Throwable` here instead of `Exception`, we could fix https://jira.uitdatabank.be/browse/III-4721 at the same time.

(Sentry calls it an `ErrorException` but I think it's actually just an `Error`, which is not an `Exception`)

src/Offer/CommandHandlers/DeleteOrganizerHandler.php
```diff
@@ -0,0 +1,38 @@
...
+        $this->organizerChangeAllowedBasedOnTicketSales = $organizerChangeAllowedBasedOnTicketSales;
+    }
+
+    public function handle($command): void
+    {
+        if (!($command instanceof DeleteOrganizer)) {
+            return;
+        }
+
+        $this->organizerChangeAllowedBasedOnTicketSales->guard($command);
```
bertramakers, line 32: Maybe we should do the checks if there is a change in these new command handlers and not in `EventHasTicketSalesGuard`, since it is dependent on the command? That way `EventHasTicketSalesGuard` does not need to know about every command that needs to do this check, it could just check if there are ticket sales or not.

bertramakers (approved):
Very clear and I think it works as intended now! Just some minor suggestions.

src/Offer/CommandHandlers/DeleteOrganizerHandler.php
LucWollants: Would it be safe to always call `eventHasTicketSales` even for places? Then the amount of code from `EventHasTicketSalesGuard` can be reduced and there won't be much duplication in the command handlers.

src/Offer/Commands/DeleteOrganizer.php
LucWollants: I think the derived event `OrganizerDeleted` would still expose the method `getOrganizerId`? In that case, it's an easy change because the aggregate can add the id of the current organizer.

bertramakers: Events may always contain more info than strictly necessary to record the change.

src/Offer/CommandHandlers/DeleteOrganizerHandler.php
bertramakers: I meant to move [this check](https://github.com/cultuurnet/udb3-silex/pull/1055/files#diff-289ee1aff94dbd0d438f353e6b00b25bc0d75f221b309ddd0217f66112bc1752R51-R55) to UpdateOrganizerHandler, and [this check](https://github.com/cultuurnet/udb3-silex/pull/1055/files#diff-289ee1aff94dbd0d438f353e6b00b25bc0d75f221b309ddd0217f66112bc1752R56-R59) to DeleteOrganizerHandler.
That way `EventHasTicketSalesGuard` doesn't kneed to know in which contexts it will be called, and we can re-use it in other command handlers later if ever needed.

bertramakers: > Would it be safe to always call eventHasTicketSales even for places?

I think if you'd call it for a place id it will result in an exception, but those are caught and handled as if there are no ticket sales so it could work. We should test it (manually) to be sure though.

src/Offer/Commands/DeleteOrganizer.php
LucWollants: As discussed through Slack this will be handled in a dedicated PR that will also add an API call to remove an organizer on event/place/organizer without having to pass an organizer id.

src/Offer/CommandHandlers/DeleteOrganizerHandler.php
LucWollants: Good proposal, this should be fixed now.

PR #1076, opened 2022-07-04
III-4828 improve Organizer Id Validation

JonasVHG:
### Added

- Added an `UpdateOrganizerRequestHandler`

### Removed

- Legacy Paths for `UpdateOrganizer`
- Test for LegacyPath `UpdateOrganizer`

---
Ticket: https://jira.uitdatabank.be/browse/III-4828

### DIFF
--- a/app/Event/EventControllerProvider.php
+++ b/app/Event/EventControllerProvider.php
@@ -54,7 +54,6 @@ class EventControllerProvider implements ControllerProviderInterface, ServicePro
 
         $controllers->put('/{cdbid}/booking-info/', 'event_editing_controller:updateBookingInfo');
         $controllers->put('/{cdbid}/contact-point/', 'event_editing_controller:updateContactPoint');
-        $controllers->put('/{cdbid}/organizer/{organizerId}/', 'event_editing_controller:updateOrganizer');
         $controllers->delete('/{cdbid}/organizer/{organizerId}/', 'event_editing_controller:deleteOrganizer');
         $controllers->put('/{cdbid}/typical-age-range/', 'event_editing_controller:updateTypicalAgeRange');
         $controllers->delete('/{cdbid}/typical-age-range/', 'event_editing_controller:deleteTypicalAgeRange');
--- a/src/Http/Offer/UpdateOrganizerRequestHandler.php
+++ b/src/Http/Offer/UpdateOrganizerRequestHandler.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Offer;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Http\Request\RouteParameters;
+use CultuurNet\UDB3\Http\Response\NoContentResponse;
+use CultuurNet\UDB3\Offer\Commands\UpdateOrganizer;
+use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
+use CultuurNet\UDB3\ReadModel\DocumentRepository;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+final class UpdateOrganizerRequestHandler implements RequestHandlerInterface
+{
+    private CommandBus $commandBus;
+    private DocumentRepository $organizerDocumentRepository;
+
+    public function __construct(CommandBus $commandBus, DocumentRepository $organizerDocumentRepository)
+    {
+        $this->commandBus = $commandBus;
+        $this->organizerDocumentRepository = $organizerDocumentRepository;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $offerId = $routeParameters->getOfferId();
+        $organizerId = $routeParameters->get('organizerId');
+
+        try {
+            $this->organizerDocumentRepository->fetch($organizerId);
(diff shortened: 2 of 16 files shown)


app/Event/EventControllerProvider.php
```diff
@@ -54,7 +54,6 @@ public function connect(Application $app): ControllerCollection
 
         $controllers->put('/{cdbid}/booking-info/', 'event_editing_controller:updateBookingInfo');
         $controllers->put('/{cdbid}/contact-point/', 'event_editing_controller:updateContactPoint');
-        $controllers->put('/{cdbid}/organizer/{organizerId}/', 'event_editing_controller:updateOrganizer');
```
bertramakers, line 57: This one should probably also be removed? https://github.com/cultuurnet/udb3-silex/blob/master/app/Place/PlaceControllerProvider.php#L37

bertramakers: + The place equivalent of the editing service

src/Http/Offer/UpdateOrganizerRequestHandler.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->commandBus = $commandBus;
+        $this->organizerDocumentRepository = $organizerDocumentRepository;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $offerId = $routeParameters->getOfferId();
+        $organizerId = $routeParameters->get('organizerId');
```
bertramakers, line 33:
```suggestion
        $organizerId = $routeParameters->getOrganizerId();
```
(The method already exists)

src/Http/Offer/UpdateOrganizerRequestHandler.php
```diff
@@ -0,0 +1,45 @@
...
+
+        try {
+            $this->organizerDocumentRepository->fetch($organizerId);
+        } catch (DocumentDoesNotExist $e) {
+            throw ApiProblem::urlNotFound('Organizer with id "' . $organizerId . '" does not exist.');
```
bertramakers, line 38: 👍

src/Http/OfferRestBaseController.php
```diff
@@ -82,30 +82,6 @@ public function deleteTypicalAgeRange(string $cdbid)
...
-    {
-        $this->editor->updateOrganizer($cdbid, $organizerId);
-
-        return new NoContent();
-    }
-
-    /**
-     * @deprecated
-     */
-    public function updateOrganizerFromJsonBody(Request $request, string $cdbid): Response
```
bertramakers, line 95: This method is still used for an even older route

bertramakers: See https://github.com/cultuurnet/udb3-silex/blob/master/app/Event/EventControllerProvider.php#L77
That route is actually better in terms of URL because the organizer id should not be part of the URL, but it uses the wrong HTTP method sadly. 😄

tests/Event/EventEditingServiceTest.php
```diff
@@ -30,15 +30,22 @@ final class EventEditingServiceTest extends TestCase
      */
     private $readRepository;
 
+    /**
+     * @var DocumentRepository|MockObject
+     */
+    private $organizerDocumentRepository;
```
LucWollants, line 36: There is no need for a property, the mock creation can be inlined inside the construction of `EventEditingService`

tests/Offer/DefaultOfferEditingServiceTest.php
```diff
@@ -36,6 +36,11 @@ class DefaultOfferEditingServiceTest extends TestCase
      */
     private $offerRepository;
 
+    /**
+     * @var DocumentRepository|MockObject
+     */
+    private $organizerDocumentRepository;
```
LucWollants, line 42: Also no need here for the property. I see that this is done with several mocks in this class.

tests/Place/DefaultPlaceEditingServiceTest.php
```diff
@@ -48,6 +48,11 @@ class DefaultPlaceEditingServiceTest extends TestCase
      */
     protected $readRepository;
 
+    /**
+     * @var DocumentRepository|MockObject
+     */
+    private $organizerDocumentRepository;
```
LucWollants, line 54: Inlining is also possible here.

bertramakers:
Wouldn't it be better to move the controller + service methods for the "update organizer from JSON" endpoint to a PSR-15 request handler? We'll have to keep supporting that older endpoint for the foreseeable future, and we'll need to refactor every endpoint to PSR-15 request handlers for Silex EOL. So might as well do it now?

LucWollants (approved):

bertramakers:
Okay I was looking at the changes per commit and it seemed like we were making changes to the `updateOrganizerFromJsonBody` controller method without refactoring it, but I see now in the complete diff that it was just to revert it back to its original state and it wasn't really changed for now. Can be done in a follow-up PR indeed.

bertramakers (approved):

bertramakers:
> I guess the approach would be to split the code flow inside the handle method of the request handler by using the ServerRequestInterface and check if we need to follow the POST or PUT flow?

That or a separate request handler, either way works. For the body we'll need JSON schema validation of course. And in an ideal world the endpoint would be `PUT /events/{eventId}/organizer` with a JSON body instead of either `PUT /events/{eventId}/organizer/{organizerId}` (correct method but putting the data in the URL) or `POST /events/{eventId}/organizer` (data correctly in the body but incorrect method because the request should be idempotent). So if we could achieve that with little extra effort by doing it in one handler that would be the best, but nice to have at this point.

JonasVHG:
I had started on moving the endpoint, but I agree that it would be better to do it in a later, separate PR.

PR #1081, opened 2022-07-20
III-4821 Reduce projected-to-jsonld messages

bertramakers:
### Changed

- Refactored our `SimpleEventBus` to a `MiddlewareEventBus`
- For HTTP requests, the `EventProjectedToJSONLD`, `PlaceProjectedToJSONLD` and `OrganizerProjectedToJSONLD` messages on the event bus are now intercepted until the request is handled. At that point the ones with unique payloads are re-published. This way we avoid publishing multiple redundant messages to the AMQP exchange, and excessive Redis writes due to related documents being updated twice or more.

### Fixed

- Fixed some messages not being flagged as "replayed" during a replay, due to the event bus being instantiated multiple times.

---
Ticket: https://jira.uitdatabank.be/browse/III-4821


src/EventBus/Middleware/CallbackOnFirstPublicationMiddleware.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace CultuurNet\UDB3\EventBus\Middleware;
+
+use Broadway\Domain\DomainEventStream;
+use Closure;
+use CultuurNet\UDB3\EventBus\EventBusMiddleware;
+
+final class CallbackOnFirstPublicationMiddleware implements EventBusMiddleware
+{
+    private int $publicationCount = 0;
```
LucWollants, line 13: I wonder if a "state" boolean would make more sense here?

src/EventBus/Middleware/ReplayFlaggingMiddleware.php
```diff
@@ -14,11 +14,17 @@ final class ReplayFlaggingMiddleware implements EventBusMiddleware
 {
     private static bool $replaying = false;
 
+    /**
+     * Static so that if multiple instances of the middleware are created accidentally, they are all in sync.
+     */
     public static function startReplayMode(): void
```
LucWollants, line 20: Would it also work to make the `ReplayFlaggingMiddleware` a singleton?

bertramakers: That was the original approach with the `ReplayFlaggingEventBus` decorator, but that was actually bugged because it was created in an `extend()` _around_ a shared service (= a singleton in pimple). And the extend was always called again, so everytime the event bus was injected somewhere a new `ReplayFlaggingEventBus` decorator was created. I tested that by doing a `var_dump()` in the constructor of the `ReplayFlaggingEventBus` and then running the replay command, and it was being instantiated about 5 times.

So I figured it was better to bake the singleton-behavior into the class itself then depend on a correct bootstrapping, since we will also be moving to another service container later and I've seen similar bugs pop up when SAPI3 was ported to a different service container. (It's easy to use `add()` instead of `addShared()` or something similar by accident)

bertramakers: \+ The static methods make it easy to access it from any part of the code (for example the replay command) without having to get it from the service container or event bus in some way

src/EventBus/Middleware/CallbackOnFirstPublicationMiddleware.php
bertramakers: I don't really have an opinion on this, I guess a boolean might be safer if we do a super large amount of publications (maybe in a long-running CLI command). Can always be refactored in a follow-up PR :)

PR #1084, opened 2022-08-03
III-4822 Embed related documents on the fly in JSON-LD of events/places

bertramakers:
### Added

- Added `DocumentRepositoryInterface` decorator for events/places that looks up and embed the documents for the `location` and `organizer` properties on the fly when the event/place is loaded from Redis
- Added `RelatedDocumentProjectedToJSONLDDispatcher` that listens to `PlaceProjectedToJSONLD` and `OrganizerProjectedToJSONLD` and publishes `EventProjectedToJSONLD` and `PlaceProjectedToJSONLD` messages to the event bus for the related events/places of the updated place/organizer.
- Added `TraceableEventBus` copied from Broadway but with an added method to get the recorded DomainMessages (instead of just the events).
- Added in-memory implementations of the event & place relations repositories for easier testing compared to mocked repositories.

### Changed

- Boyscouting to add types to relation repositories, renamed them for easier distinction and usage, renamed their service container entries to class/interface names, and moved their service container entries to service providers

### Removed

- Removed `RelatedEventLDProjector` so events' JSON-LD is no longer updated if a related place or organizer is updated
- Removed `RelatedPlaceLDProjector` so places' JSON-LD is no longer updated if a related organizer is updated

---
Ticket: https://jira.uitdatabank.be/browse/III-4822


src/Offer/ProcessManagers/RelatedDocumentProjectedToJSONLDDispatcher.php
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    private function handlePlaceProjectedToJSONLD(PlaceProjectedToJSONLD $placeProjectedToJSONLD): void
+    {
+        $placeId = $placeProjectedToJSONLD->getItemId();
+
+        // In theory the event relations repository should only return unique event ids for a given place id, but to be
+        // safe pass it through array_unique() so we don't needlessly dispatch duplicate messages.
+        $eventIds = $this->eventRelationsRepository->getEventsLocatedAtPlace($placeId);
```
LucWollants, line 74: The table `event_relations` has a primary key on `event` so it should be unique. Although it doesn't harm to call `array_unique`

bertramakers: Yes but that is only one implementation, and the implementation could in theory change. And like you said it doesn't harm :)

LucWollants (approved):
Very clean solution and easy to review 👌

bertramakers:
I'll first add an acceptance test for the scenario we discussed in Slack before merging this:

- Create an organizer
- Create a place
- Create an event
- Link the event to the place
- Link the place to the organizer
- Update the name of the organizer
- Fetch the event detail and check that the organizer name is updated in `location.organizer.name`

JonasVHG (approved):

bertramakers:
I've added the acceptance tests here: https://github.com/cultuurnet/udb3-acceptance-tests/pull/133

PR #1085, opened 2022-08-04
III-4849 Handle locations with missing or null id

LucWollants:
### Fixed
- Handle locations with missing id or null id better in HTML exporter

---
Ticket: https://jira.uitdatabank.be/browse/III-4849


bertramakers (commented):
This technically fixes it, but I don’t think the result is correct now. If I’m not mistaken this will export every event with a location that had no `@id` as if it has an online location. But there are events that were imported from UDB2 and even relatively recently still from EntryAPI2 that have a physical location with a real address, but no `location.@id` because it was not required in UDB2 / EntryAPI2 to link to a separate location resource. You could just include a location name and address.

So I think the actual fix is just to check if the event had a `location.@id` before accessing it to check if the `@id` is the one for online locations to fix the `TypeError`, but not to export events without `location.@id` as if they have an online location.

bertramakers:
Code itself is perfect now but the test names are a bit outdated and they also use an online address which might be confusing later. A real use case is a physical address and "real" name on the location without `@id`

LucWollants:
> Code itself is perfect now but the test names are a bit outdated and they also use an online address which might be confusing later. A real use case is a physical address and "real" name on the location without `@id`

Those tests did indeed make no sense, I think everything is fine now.

JonasVHG (approved):

bertramakers (approved):

PR #1089, opened 2022-08-05
III-4851 Optimize AMQP publishing during/after replay

bertramakers:
### Added

- Added `--no-amqp-messages-after-replay` option to replay command, to disable the publication of "ProjectedToJSONLD" messages (1 per replayed event/place/organizer) that happens at the end of the replay

### Changed

- During a replay, no AMQP messages are published
- After a replay, one ProjectedToJSONLD message is published for every event/place/organizer that was replayed. One message will also be published for every related event/place in the case of places/organizers, via the existing `RelatedDocumentProjectedToJSONLDDispatcher` event bus listener.

### Removed

- Removed `--disable-publishing` option from replay command
- Removed `--disable-related-offer-subscribers` option from replay command
- Removed `--subscriber` option from replay command
- Removed `ConfigWriter` class and test

---
Ticket: https://jira.uitdatabank.be/browse/III-4851


app/Console/ReplayCommand.php
```diff
@@ -28,6 +28,7 @@ class ReplayCommand extends AbstractCommand
     public const OPTION_START_ID = 'start-id';
     public const OPTION_DELAY = 'delay';
     public const OPTION_CDBID = 'cdbid';
+    public const OPTION_DISABLE_AMQP_PUBLICATION_POST_REPLAY = 'disable-amqp-publication-post-replay';
```
LucWollants, line 33: Maybe a name that fits the intercepting mechanism is `disable-amqp-publication-of-intercepted-messages`?

LucWollants (approved):

app/Console/ReplayCommand.php
bertramakers: The intercepting felt like an implementation detail to me. When discussing the functionality with Kristof we always talked about it in terms of "dispatching 1 message per replayed or related event/place/organizer, after the replay". I didn't even think of the "intercepting" as an easy implementation for that at first, I even started an implementation that stored the replayed ids and then queried the related repositories etc 😄 So to summarize, I'm not sure the "intercepting" part would be clear for someone who uses the command without knowing the code details.

In hindsight the "post-replay" could maybe be changed to "after-replay" for clarity though.

bertramakers: I have renamed it to `--no-amqp-publication-after-replay`, that is also more consistent with the standard `--no-ansi` and `--no-interaction` options that Symfony Console provide

PR #1095, opened 2022-08-18
III-4881 Refactor productions write controller to use PSR7 interfaces

LucWollants:
### Added
- Added new trait `AssertJsonResponseTrait` to assert JSON responses
- Added new request handler `CreateProductionRequestHandler` with PSR7 interfaces
- Added new request handler `AddEventRequestHandler` with PSR7 interfaces
- Added new request handler `RemoveEventRequestHandler` with PSR7 interfaces
- Added new request handler `MergeProductionsRequestHandler` with PSR7 interfaces
- Added new request handler `RenameProductionRequestHandler` with PSR7 interfaces
- Added new request handler `SkipEventsRequestHandler` with PSR7 interfaces

### Removed
- Removed the now obsolete `ProductionsWriteController`

---

Ticket: https://jira.uitdatabank.be/browse/III-4881


app/Event/ProductionControllerProvider.php
```diff
@@ -48,16 +42,52 @@ function (Application $app) {
...
         );
 
+        $app[CreateProductionRequestHandler::class] = $app->share(
+            fn (Application $app) => new CreateProductionRequestHandler(
+                $app['event_command_bus'],
+                new CreateProductionValidator()
+            )
+        );
+
+        $app[AddEventRequestHandler::class] = $app->share(
```
bertramakers, line 52: Maybe also mention "Production" in this class name (and `RemoveEventRequestHandler`)? For example:
```suggestion
        $app[AddEventToProductionRequestHandler::class] = $app->share(
```
It's already in the Productions namespace, but it may be clearer when looking at it without knowing the namespace + the other request handlers also have it in their name (e.g. `RenameProductionRequestHandler` instead of `RenameRequestHandler`)

bertramakers: Just a nit though

src/Http/Productions/AddEventRequestHandler.php
```diff
@@ -0,0 +1,40 @@
...
+    public function __construct(CommandBus $commandBus)
+    {
+        $this->commandBus = $commandBus;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $eventId = $routeParameters->getEventId();
+        $productionId = $routeParameters->get('productionId');
```
bertramakers, line 29: We could add a `getProductionId()` for this, since it's used in multiple RequestHandlers

bertramakers (approved):

app/Event/ProductionControllerProvider.php
LucWollants: Good point, I also was doubting this so I made it consistent now.

bertramakers (approved):

PR #1101, opened 2022-08-22
III-4866 refactor EditPlaceRestController to 2 handlers-1/2-updateAddress()

JonasVHG:
### Added

- RequestHandler for `updateAddress()` & tests
- test for new RequestHandler

### Changed

- make `PlaceControllerProvider` use `RequestHandler` for `updateAddress()`
- replace usages of `DefaultPlaceEditingService` with `DefaultOfferEditingService`

### Removed

- remove `updateAddress()` & `placeContext()` from `EditPlaceRestController`
- remove `updateAddress()` from `PlaceEditingServiceInterface`
- remove `DefaultPlaceEditingService` & tests
- tests for `EditPlaceRestController`

---
Ticket: https://jira.uitdatabank.be/browse/III-4866


src/Http/Place/UpdateAddressRequestHandler.php
```diff
@@ -0,0 +1,57 @@
...
+        $address = (new AddressJSONDeserializer())->deserialize(
+            new StringLiteral($request->getBody()->getContents())
+        );
+
+        try {
+            $this->placeDocumentRepository->fetch($placeId);
+        } catch (DocumentDoesNotExist $e) {
+            throw new EntityNotFoundException(
+                sprintf('Place with id: %s not found.', $placeId)
+            );
```
bertramakers, line 48: @LucWollants @JonasVHG maybe it would be good to have an improvement ticket (to pick up later) to add a check like this to every event/place endpoint? So we don't have to repeat this in every request handler, but do have this check everywhere. I think it's now missing in a lot of places and it seems very useful to catch mistakes.
And also for other resources maybe?
I can check if it's possible to do this easily in the new router (#1102). Sometimes routers have a feature to look up a wildcard like an id first before the request is sent to a request handler.

src/Place/DefaultPlaceEditingService.php
```diff
@@ -1,22 +0,0 @@
...
-declare(strict_types=1);
-
-namespace CultuurNet\UDB3\Place;
-
-use CultuurNet\UDB3\Address\Address;
-use CultuurNet\UDB3\Language;
-use CultuurNet\UDB3\Offer\DefaultOfferEditingService;
-use CultuurNet\UDB3\Place\Commands\UpdateAddress;
-
-class DefaultPlaceEditingService extends DefaultOfferEditingService implements PlaceEditingServiceInterface
```
bertramakers, line 12: Good riddance! 🙌

bertramakers (approved):
Perfect! 👍

src/Http/Place/UpdateAddressRequestHandler.php
LucWollants: I created [III-4899](https://jira.uitdatabank.be/browse/III-4899) for this.

tests/Http/Place/UpdateAddressRequestHandlerTest.php
```diff
@@ -0,0 +1,255 @@
...
+                '{
+                    "streetAddress": "Veldstraat 11",
+                    "postalCode": "9000",
+                    "addressLocality": "Gent",
+                    "addressCountry": "BE"
+                }'
+            )
+            ->build('PUT');
+
+        $this->updateAddressRequestHandler->handle($updateAddressRequest);
```
LucWollants, line 72: Although the response is `NoContentResponse` I would still check it with `AssertJsonResponseTrait`

tests/Http/Place/UpdateAddressRequestHandlerTest.php
```diff
@@ -0,0 +1,255 @@
...
+                }'
+            )
+            ->build('PUT');
+
+        $this->expectException(InvalidArgumentException::class);
+
+        $this->updateAddressRequestHandler->handle($updateAddressRequest);
+    }
+
+    public function it_throws_on_empty_values(): void
```
LucWollants, line 236: Very good to add these checks already, but because the name of the method is plural `values` it's maybe better to also check this for the other properties?

src/Http/Place/UpdateAddressRequestHandler.php
bertramakers: Ah now that I see the `DocumentDoesNotExist` in the title I just remembered, we already have a solution for this: This exception is caught in the error handler and converted to an API problem. So I think we can just leave out the check in the request handlers. We also have the same for `AggregateNotFoundException`, so if a command is dispatched and the command handler loads the aggregate but gets that exception, we also get a nice API problem right now: https://github.com/cultuurnet/udb3-silex/blob/master/app/Error/WebErrorHandlerProvider.php#L92-L114

So I think we can keep that for now and just remove the check in this request handler (and others that we refactor)?

tests/Http/Place/UpdateAddressRequestHandlerTest.php
```diff
@@ -0,0 +1,255 @@
...
+                '{
+                    "streetAddress": "Veldstraat 11",
+                    "postalCode": "9000",
+                    "addressLocality": "Gent",
+                    "addressCountry": "BE"
+                }'
+            )
+            ->build('PUT');
+
+        $updateAddressRequest2 = $this->psr7RequestBuilder
```
LucWollants, line 170: Nit: I personally don't like adding number suffixes to variable names because that requires understanding the full scope. Maybe adding a language suffix also works, for example: `updateAddressInDutchRequest` and `updateAddressInFrenchRequest` are more

LucWollants (approved):
I already approved because my remarks are minor.

src/Http/Place/UpdateAddressRequestHandler.php
bertramakers: Also `EntityNotFoundException` is not converted to a nice API problem response right now I think, so this extra check in the request handler is currently worse than the error handling that we have. So I would suggest to remove it from this request handler in this PR

LucWollants: @JonasVHG the ticket is not needed and the `fetch` with `catch` can be removed.

tests/Http/Place/UpdateAddressRequestHandlerTest.php
```diff
@@ -47,8 +47,7 @@ protected function setUp(): void
         );
 
         $this->updateAddressRequestHandler = new UpdateAddressRequestHandler(
-            $this->commandBus,
-            $placeDocumentRepository
```
LucWollants: The `$placeDocumentRepository` can be removed completely I think.

PR #1102, opened 2022-08-22
III-4896 Add PSR router

bertramakers:
### Added

- PSR router that gets used if a request cannot be matched by the Silex router
- Support for OPTIONS requests on routes registered on the PSR router (but we still also need a catch-all OPTIONS route in Silex to support CORS on the Silex routes for now). The necessary CORS headers are still added to every response by the middleware in Silex, even for routes matched by the new router. This will be moved in a later ticket once all routes are moved to the new router. (We cannot do it earlier since we also need the CORS headers on the routes in the Silex router)

### Changed

- Registered the `GetDetailRequestHandler` for events and places on the new router as a proof of concept
- Move path rewriting to a `LegacyPathRewriter` class
- Renamed `LegacyRoutesServiceProvider` to `CatchAllRouteServiceProvider` as it is also used to dispatch requests to the new PSR router (until we can by-pass the Silex router completely once all routes, middlewares etc are moved)

---
Ticket: https://jira.uitdatabank.be/browse/III-4896

All acceptance tests are green, and the CORS requests still work in the UI, also for event details. I also tested the OPTIONS request via curl like this (make sure to change the event URL):

```
curl -i 'https://io.uitdatabank.dev/event/6c1c3991-0e8f-480d-9c02-16eadb62dae2' -X OPTIONS -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:103.0) Gecko/20100101 Firefox/103.0' -H 'Accept: */*' -H 'Accept-Language: en,en-US;q=0.7,nl;q=0.3' -H 'Accept-Encoding: gzip, deflate, br' -H 'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: authorization,x-api-key' -H 'Referer: https://old.uitdatabank.dev/' -H 'Origin: https://old.uitdatabank.dev' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache'
```

Notes:

- We currently cannot register request handlers without making them a callable by wrapping them in an array and adding the `handle` method as a second array value. I opened an issue for this in the router's repo: https://github.com/thephpleague/route/issues/324
- We currently cannot do the path rewriting in a middleware because middlewares are not called for unmatched routes. I've commented on an open issue for that: https://github.com/thephpleague/route/issues/319. But for now the current solution that rewrites the path of the request before it gets dispatched on the router works fine


LucWollants (approved):
I like the test for path rewrites!

PR #1103, opened 2022-08-23
III-4869 Refactor patch offer rest controller to PSR7 interface

LucWollants:
### Added
- Added new `PatchOfferRequestHandler` with PSR7 interface

### Removed
- Removed the obsolete `PatchOfferRestController`

---
Ticket: https://jira.uitdatabank.be/browse/III-4869


app/Offer/OfferControllerProvider.php
```diff
@@ -60,6 +61,8 @@ public function connect(Application $app): ControllerCollection
 
         $controllers->put('/{offerType}/{offerId}/organizer/{organizerId}/', UpdateOrganizerRequestHandler::class);
 
+        $controllers->patch('/{offerType}/{offerId}', PatchOfferRequestHandler::class);
```
bertramakers, line 64:
```suggestion
        $controllers->patch('/{offerType}/{offerId}/', PatchOfferRequestHandler::class);
```

We always add a trailing slash, and I think it's actually needed to make the request work with both a trailing slash and without a trailing slash.

src/Http/Offer/PatchOfferRequestHandler.php
```diff
@@ -5,62 +5,60 @@
...
-    public function handle(Request $request, string $cdbid): Response
+    public function handle(ServerRequestInterface $request): ResponseInterface
     {
+        $routeParameters = new RouteParameters($request);
+        $offerType = $routeParameters->getOfferType();
+        $offerId = $routeParameters->getOfferId();
+
         $domainModel = $this->parseDomainModelNameFromRequest($request);
-        $commandClass = 'CultuurNet\UDB3\\' . $this->offerType->toString() . '\Commands\Moderation\\' . $domainModel;
+        $commandClass = 'CultuurNet\UDB3\\' . $offerType->toString() . '\Commands\Moderation\\' . $domainModel;
```
bertramakers, line 35: I know it's not new code but I'm curious, why are some backslashes escaped and others not in this string? Is the escaping even necessary in single quotes?

bertramakers (approved):
Already approving because my comments are very small. Nice work! 👍

src/Http/Offer/PatchOfferRequestHandler.php
LucWollants: Otherwise if we do `\'` then it will escape the single quote, so we need `\\'` to avoid escaping

bertramakers: Ah yes I see now! Thanks for explaining :)

PR #1104, opened 2022-08-24
III-4866 refactor EditPlaceRestController to 2 handlers-2/2-getEvents()

JonasVHG:
### Fixed

---
Ticket: https://jira.uitdatabank.be/browse/III-4866


app/Place/PlaceControllerProvider.php
```diff
@@ -64,12 +64,10 @@ public function connect(Application $app): ControllerCollection
 
     public function register(Application $app): void
     {
-        $app['place_editing_controller'] = $app->share(
```
LucWollants, line 67: The `place_editing_controller` is still used for the endpoints that are not yet replaced. The class it uses should now be `OfferRestBaseController`

LucWollants (approved):

PR #1106, opened 2022-08-24
III-4901 Replace `symfony/security` with another solution

bertramakers:
### Removed

- Removed `symfony/security`, which will not work with the new router / PSR-7 requests. It is replaced with a custom solution (building on existing code that we already had to parse/validate JWTs and API keys), which I think makes it a lot easier to see at a glance how requests are authenticated.

### Changed

- Cleaned up some services related to authentication, like `current_user_id`, `current_user_is_god_user`, `jwt`, etc ...

---
Ticket: https://jira.uitdatabank.be/browse/III-4901


bertramakers:
I did not add a unit test for the new `RequestAuthenticator` because I think the tests in https://github.com/cultuurnet/udb3-acceptance-tests/pull/141 should cover it, and it seems more valuable to test it with integration tests instead of unit tests in this case

LucWollants (approved):
Perfect! It's very easy to understand what is happening by reading the code now.

JonasVHG (approved):

PR #1112, opened 2022-08-25
III-4876 Refactor label read rest controller to use PSR-7 interfaces

LucWollants:
### Added
- Added `labelNotFound` as new `ApiProblem`
- Add method `getLabelId` on `RouteParameters`
- Created new `GetLabelRequestHandler` with PSR-7 interfaces
- Created new `SearchLabelsRequestHandler` with PSR-7 interfaces

### Changed
- Changed `PatchLabelRequestHandler` to use `getLabelId`
- Refactored `QueryFactory` to use `ServerRequestInterface`

### Removed
- Deleted obsolete `ReadRestController`

---
Ticket: https://jira.uitdatabank.be/browse/III-4876


JonasVHG (approved):

src/Http/Label/Query/QueryFactoryInterface.php
```diff
@@ -5,9 +5,9 @@
 namespace CultuurNet\UDB3\Http\Label\Query;
 
 use CultuurNet\UDB3\Label\ReadModels\JSON\Repository\Query;
-use Symfony\Component\HttpFoundation\Request;
+use Psr\Http\Message\ServerRequestInterface;
 
 interface QueryFactoryInterface
 {
-    public function createFromRequest(Request $request): Query;
+    public function createFromRequest(ServerRequestInterface $request): Query;
```
bertramakers, line 12: Might be something to simplify later by removing this interface and the implementation and just moving this to `SearchLabelsRequestHandler`, since that seems to be the only place that it is actually used and with the dependency on a `ServerRequestInterface` it's unlikely that it will be used someplace else

bertramakers (approved):

src/Http/Label/Query/QueryFactoryInterface.php
LucWollants: Keeping the `QueryFactory` can help a bit with testing. But you propose to also move the `QueryFactory` implementation to `SearchLabelsRequestHandler`?

bertramakers: Yes, but not in this PR or even in the near future. I just noticed that (in my humble opinion) it's an unnecessary abstraction because there is only 1 implementation and 1 usage, and it's even very specific to this one HTTP request so very little chance that we'll re-use it at one point or swap it with a different implementation. Especially because all it does is parse some query parameters, and because those always have to be the same for this request handler to avoid breaking changes, the parsing of those parameters is very tightly coupled to this request handler anyway.

Not to say that it's terribly bad code, it's very well possible that I wrote it at some point (or something like this but in another place), but over the years I've noticed that UDB3 tends to have a little bit too much abstractions and complexity for what it does sometimes 😄 But it's a learning process and I've definitely been guilty of it myself.

bertramakers: In what way does it help with testing? If we use a mock or another implementation in the tests, aren't we then testing the request handler in a less reliable way? Because then we don't know that it parses those query params correctly. And those are an essential part of the request handling

bertramakers: In any case it's not in scope of this refactor to clean this up. :)

LucWollants: > In what way does it help with testing? If we use a mock or another implementation in the tests, aren't we then testing the request handler in a less reliable way? Because then we don't know that it parses those query params correctly. And those are an essential part of the request handling

When testing the `QueryFactory` directly only the `Request` needs to be mocked. When doing this in the context of the `SearchLabelsRequestHandler` the label `ReadRepositoryInterface` needs to be mocked also.
But that should not be the driving factor for adding abstractions, like you say, we already have enough abstractions 😅

PR #1116, opened 2022-08-26
III-4871 Refactor OfferPermission(s)Controller into 4 handlers

JonasVHG:
### Added
- classes for `UserPermissionChecker` &  `PrivateJsonResponse`
- RequestHandlers & Tests: `CurrentUserHasPermissionRequestHandler`, `GivenUserHasPermissionRequestHandler`, `GetPermissionsForCurrentUserRequestHandler`, `GetPermissionsForGivenUserRequestHandler`

### Changed
- Changed bindings for `permissions` & `permission` endpoints

### Removed
- `OfferPermissionController`, `OfferPermissionsController` & tests

---
Ticket: https://jira.uitdatabank.be/browse/III-4871


app/Offer/OfferControllerProvider.php
```diff
@@ -63,6 +71,14 @@ public function connect(Application $app): ControllerCollection
 
         $controllers->patch('/{offerType}/{offerId}/', PatchOfferRequestHandler::class);
 
+        /**
+         * Legacy routes that we need to keep for backward compatibility.
+         * These routes usually used an incorrect HTTP method.
```
LucWollants, line 76: In this case they just have a singular `permission` path.

bertramakers: Indeed, that comment was from another file where there used to be more legacy routes and most of those had an incorrect HTTP method that was later changed. But in this case it's a different URL + also slightly different behavior

app/Organizer/OrganizerControllerProvider.php
```diff
@@ -154,12 +154,21 @@ public function register(Application $app): void
...
 
-        $app['organizer_permissions_controller'] = $app->share(
+        $app[GetPermissionsForCurrentUserRequestHandler::class] = $app->share(
             function (Application $app) {
-                return new OfferPermissionsController(
+                return new GetPermissionsForCurrentUserRequestHandler(
                     [Permission::organisatiesBewerken()],
                     $app['organizer_permission_voter'],
-                    $app['current_user_id'] ? new StringLiteral($app['current_user_id']) : null
+                    $app['current_user_id'] ?? null
```
bertramakers, line 162: Note: This service does not exist anymore since I merged #1106 . You can get the current user id by getting the `CurrentUser::class` service and then calling the `getId()` method. You can either inject the `CurrentUser` object or a `string` id in the request handler.

app/Offer/OfferControllerProvider.php
```diff
@@ -92,6 +108,44 @@ public function register(Application $app): void
...
 
+        $app[GetPermissionsForCurrentUserRequestHandler::class] = $app->share(
+            fn (Application $app) => new GetPermissionsForCurrentUserRequestHandler(
+                [
+                    Permission::aanbodBewerken(),
+                    Permission::aanbodModereren(),
+                    Permission::aanbodVerwijderen(),
+                ],
+                $app['offer_permission_voter'],
+                $app['current_user_id'] ?? null
```
bertramakers, line 119: This will always be `null` since the merge of #1106 (see other comment)

app/Offer/OfferControllerProvider.php
```diff
@@ -92,6 +108,44 @@ public function register(Application $app): void
...
+                ],
+                $app['offer_permission_voter'],
+            )
+        );
+
+        $app[CurrentUserHasPermissionRequestHandler::class] = $app->share(
+            fn (Application $app) => new CurrentUserHasPermissionRequestHandler(
+                Permission::aanbodBewerken(),
+                $app['offer_permission_voter'],
+                $app['current_user_id'] ?? null
```
bertramakers, line 138: This will always be `null` since the merge of #1106 (see other comment)

LucWollants:
We went through the PR together and also noticed 2 things:
- There are no acceptance tests for these endpoints
- The abstraction layer is a bit confusing and a new class `UserPermissionChecker` with a method `getPermissions` makes more sense

src/Http/Offer/CurrentUserHasPermissionRequestHandler.php
```diff
@@ -33,6 +34,6 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
         $hasPermission = $this->hasPermission($offerId, $this->currentUserId);
 
-        return new JsonResponse($hasPermission, StatusCodeInterface::STATUS_OK, $this->getPrivateHeaders());
+        return new UncacheableJsonResponse($hasPermission, StatusCodeInterface::STATUS_OK);
```
bertramakers: Couple of concerns with this class:

- `Uncacheable` is incorrect, because you set `Cache-Control: private` which means that it should be cached but only in private caches, for example browser caches. It must not be stored in shared caches, for example the caches of proxies.
- This response is for a request that uses an `Authorization` header. But responses to requests with an Authorization header cannot be stored in shared caches by design, since they are unique for a given user. So the `Cache-Control: private` is not necessary / useful I think.

(See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)

src/Http/Offer/CurrentUserHasPermissionRequestHandler.php
```diff
@@ -36,6 +36,6 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
         $hasPermission = $this->userPermissionChecker->hasPermission($offerId, $this->currentUserId);
 
-        return new UncacheableJsonResponse(['hasPermission' => $hasPermission], StatusCodeInterface::STATUS_OK);
+        return new PrivateJsonResponse(['hasPermission' => $hasPermission], StatusCodeInterface::STATUS_OK);
```
bertramakers, line 39: @JonasVHG to clarify my previous comment, you can just use `JsonResponse` and not add the `cache-control` header in my opinion

src/Http/Offer/CurrentUserHasPermissionRequestHandler.php
JonasVHG: The name has been updated to `PrivateJsonResponse` to better describe to actual function.
I've kept it the same as the old code where the header is also set. Since the uses `Authorization` it might indeed be a bit too secure or unnecessary

src/Http/Offer/CurrentUserHasPermissionRequestHandler.php
```diff
@@ -0,0 +1,41 @@
...
+final class CurrentUserHasPermissionRequestHandler implements RequestHandlerInterface
+{
+    private UserPermissionChecker $userPermissionChecker;
+
+    private ?string $currentUserId;
+
+    public function __construct(
+        Permission $permission,
+        PermissionVoter $permissionVoter,
+        ?string $currentUserId = null
```
LucWollants, line 26: I think this doesn't need to be nullable after the change to `CurrentUser`

src/Security/Permission/UserPermissionChecker.php
```diff
@@ -0,0 +1,51 @@
...
+    private array $permissions;
+    private PermissionVoter $permissionVoter;
+
+    public function __construct(array $permissions, PermissionVoter $permissionVoter)
+    {
+        $this->permissions = $permissions;
+        $this->permissionVoter = $permissionVoter;
+    }
+
+    public function getOwnedPermissions(string $resourceId, ?string $userId = null): array
```
LucWollants, line 20: I think this always requires a `userId`, same for the `hasPermission`

tests/Http/Offer/CurrentUserHasPermissionRequestHandlerTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+    private Psr7RequestBuilder $psr7RequestBuilder;
+
+    private Headers $headers;
+
+    public function setUp(): void
+    {
+        $this->voter = $this->createMock(PermissionVoter::class);
+
+        $currentUserId = 'cd8d2005-e978-4f4c-9eb6-a0c0104fd8d0';
```
LucWollants, line 38: No need for an extra local variable.

tests/Http/Offer/CurrentUserHasPermissionRequestHandlerTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * @test
+     * @dataProvider hasPermissionDataProvider
+     */
+    public function it_returns_if_the_current_user_has_permission(bool $hasPermission): void
+    {
+        $this->voter->method('isAllowed')->willReturn($hasPermission);
+
+        $currentUserHasPermissionRequestHandler = $this->psr7RequestBuilder
```
LucWollants, line 60: This is a `request` and not a handler

src/Http/Organizer/GetPermissionsForCurrentUserRequestHandler.php
```diff
@@ -0,0 +1,47 @@
...
+use CultuurNet\UDB3\Http\Request\RouteParameters;
+use CultuurNet\UDB3\Http\Response\PrivateJsonResponse;
+use CultuurNet\UDB3\Role\ValueObjects\Permission;
+use CultuurNet\UDB3\Security\Permission\PermissionVoter;
+use CultuurNet\UDB3\Security\Permission\UserPermissionChecker;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+final class GetPermissionsForCurrentUserRequestHandler implements RequestHandlerInterface
```
LucWollants, line 16: I think there is no need to duplicate this code, you can add a new instance of `GetPermissionsForCurrentUserRequestHandler` to the container with another name and set the organizer permission.

tests/Http/Offer/CurrentUserHasPermissionRequestHandlerTest.php
```diff
@@ -0,0 +1,81 @@
...
+use CultuurNet\UDB3\Role\ValueObjects\Permission;
+use CultuurNet\UDB3\Security\Permission\PermissionVoter;
+use Fig\Http\Message\StatusCodeInterface;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Slim\Psr7\Headers;
+
+final class CurrentUserHasPermissionRequestHandlerTest extends TestCase
+{
+    use AssertJsonResponseTrait;
```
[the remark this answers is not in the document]
JonasVHG: I think it is on line 59.

src/Http/Organizer/GetPermissionsForCurrentUserRequestHandler.php
bertramakers: But that class is in the `Offer` namespace. Wouldn't it be better to keep them separated, even if that means duplicating a little bit of code? We also do that for other features that exist both on organizers and offers like for example labels.

src/Http/Offer/CurrentUserHasPermissionRequestHandler.php
bertramakers: See https://github.com/cultuurnet/udb3-silex/pull/1116#discussion_r959298493

src/Http/Permissions/GetPermissionsForCurrentUserRequestHandler.php
```diff
@@ -0,0 +1,43 @@
...
+        string $currentUserId
+    ) {
+        $this->userPermissionChecker = new UserPermissionChecker($permissions, $permissionVoter);
+        $this->currentUserId = $currentUserId;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $offerId = $routeParameters->getOfferId();
```
LucWollants, line 37: Now both offers and organizers work with an `offerId` route parameter. As discussed on Slack this would be better to be `organizerId` for the organizer endpoint. This could be done with an abstract method that calls `getOfferId` for offers and `getOrganizerId` for organizers.

bertramakers (approved):
Nice work, it's a lot easier now to understand the difference between the legacy `/permission` endpoint and newer `/permissions` endpoint with the very clear request handler names!

LucWollants:
@JonasVHG thanks for the hard work on this pull request! It was a bumpy road but the final result is very comprehendible and maintainable 👌

PR #1122, opened 2022-09-01
III-4888 Refactor event card systems controller

LucWollants:
### Added
- Added new request handler `GetCardSystemsFromEventRequestHandler`
- Added new request handler `SetCardSystemsOnEventRequestHandler`
- Added new request handler `AddCardSystemToEventRequestHandler`
- Added new request handler `DeleteCardSystemFromEventRequestHandler`

### Changed
- Refactor `CardSystemsJsonResponse` to PSR-7 interface
- Use `CallableIriGenerator` instead of Symfony `UrlGeneratorInterface`

### Removed
- Removed the now obsolete `EventCardSystemsController`

---
Note: this also requires PR https://github.com/cultuurnet/udb3-silex/pull/1123 because of the refactored `CardSystemsJsonResponse`

---
Ticket: https://jira.uitdatabank.be/browse/III-4888


JonasVHG (approved):

app/UiTPASService/UiTPASServiceEventControllerProvider.php
```diff
@@ -26,44 +29,49 @@ public function connect(Application $app): ControllerCollection
...
 
-        $app['uitpas.event_card_systems_controller'] = $app->share(
-            function (Application $app) {
-                return new EventCardSystemsController(
-                    $app['uitpas']
-                );
-            }
+        $app[GetCardSystemsFromEventRequestHandler::class] = $app->share(
+            fn (Application $app) => new GetCardSystemsFromEventRequestHandler(
+                $app['uitpas']
```
bertramakers, line 34: Nit: The constructors of these request handlers could be on one line since it's only argument

src/UiTPASService/Controller/SetCardSystemsOnEventRequestHandler.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->uitpas = $uitpas;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $content = $request->getBody()->getContents();
+        $cardSystemIds = empty($content) ? [] : Json::decodeAssociatively($content);
+
+        if (!is_array($cardSystemIds) || empty($cardSystemIds)) {
+            throw ApiProblem::bodyInvalidDataWithDetail('Payload should be an array of card system ids');
```
bertramakers, line 31: Thanks for already converting these exceptions to API problems! 👍

bertramakers (approved):
Only a very small nit but can definitely also be merged as-is!

app/UiTPASService/UiTPASServiceEventControllerProvider.php
```diff
@@ -26,44 +29,49 @@ public function connect(Application $app): ControllerCollection
...
 
         /** @var ControllerCollection $controllers */
         $controllers = $app['controllers_factory'];
 
-        $controllers->get(
-            '/{eventId}/',
-            GetUiTPASDetailRequestHandler::class
-        )->bind(self::EVENT_DETAIL);
+        $controllers->get('/{eventId}/', GetUiTPASDetailRequestHandler::class)
+            ->bind(self::EVENT_DETAIL);
```
bertramakers, line 60: As we just discussed on Slack, reminder that we should remove these `bind()` calls and not inject `Symfony\Component\Routing\Generator\UrlGeneratorInterface` in `GetUiTPASDetailRequestHandler`, but instead inject a `CallableIriGenerator` to generate those URLs. Otherwise the URL generation will break when we move these routes with `bind()` calls to the new router, since the Symfony `UrlGeneratorInterface` won't be able to find them in the old router anymore

bertramakers: There are also 2 other `bind()` calls in other controller providers. I think those can just be removed, because we don't use `UrlGeneratorInterface` in any other place. I think they are left over from long ago.

PR #1126, opened 2022-09-02
III-4906 Lazy load request handlers

bertramakers:
### Added

- Added a decorator to make any RequestHandlerInterface callable
- Added a class to implement the PSR ContainerInterface using the existing Pimple container (from Silex)
- Added a decorator for the PSR ContainerInterface to decorate any non-callable RequestHandlerInterface with the decorator to make it callable

### Changed

- Injected a container into the League Router's strategy, so it can load the request handlers based on their class name after the middlewares (including auth middleware later) are executed
- Updated the route for event/place details to no longer fetch the handler from `$app`, which caused it to be instantiated immediately. (If we did this for every route, we'd be instantiating every request handler, and all their dependencies, to handle 1 request)

---

Because the request handler for a route is now only instantiated after the route is matched and the middlewares have been executed (tested this manually with a temporary PSR-15 middleware not included in this PR), their constructor is only called when the `CurrentUser` has already been set. So we can safely inject the user id into the constructor. The same goes for other services that are required by the request handlers, for example the `AuthorizableCommandBus`.

Ticket: https://jira.uitdatabank.be/browse/III-4906


app/PimplePSRContainerBridge.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Silex;
+
+use InvalidArgumentException;
+use Pimple;
+use Psr\Container\ContainerInterface;
+
+final class PimplePSRContainerBridge implements ContainerInterface
```
bertramakers, line 11: I think we might also be able to use this as a temporary solution while migrating to an actual PSR solution. But to be investigated further when we get to that point.

LucWollants (approved):
Looks good!

PR #1127, opened 2022-09-05
III-4868 refactor EditOfferRestController

JonasVHG:
### Added

- new RequestHandlers: `AddLabelFromJsonBodyRequestHandler`, `AddLabelRequestHandler`, `RemoveLabelRequestHandler` & `UpdateDescriptionRequestHandler` and tests 

### Changed

- Bindings for `AddLabelFromJsonBody`, `UpdateDescription`, `RemoveLabel` & `AddLabel`

### Removed

- `EditOfferRestController `
- `DeprecatedOfferControllerProvider`

---
Ticket: https://jira.uitdatabank.be/browse/III-4868


src/Http/Offer/AddLabelFromJsonBodyRequestHandler.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $offerId = $routeParameters->getOfferId();
+        $bodyContent = $request->getBody()->getContents();
+        try {
+            $label = $this->labelJsonDeserializer->deserialize(new StringLiteral($bodyContent));
+        } catch (\InvalidArgumentException $exception) {
+            throw ApiProblem::urlNotFound('The label should match pattern: ^[^;]{2,255}$');
```
LucWollants, line 40: This is not an `urlNotFound` exception, but rather an invalid data exception.

src/Http/Offer/RemoveLabelRequestHandler.php
```diff
@@ -0,0 +1,33 @@
...
+    public function __construct(CommandBus $commandBus)
+    {
+        $this->commandBus = $commandBus;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $offerId = $routeParameters->getOfferId();
+        $labelName = $routeParameters->getLabelNameAsString();
```
LucWollants, line 28: This is only used here, so I would use the `get` method with a parameter. This also prevents exposing this method instead of `getLabel` which is still the preferred way of getting a label.

tests/Http/Offer/AddLabelFromJsonBodyRequestHandlerTest.php
```diff
@@ -0,0 +1,133 @@
...
+        string $offerType,
+        string $type,
+        AbstractLabelCommand $addLabel
+    ): void {
+        $addLabelFromJsonBodyRequest = $this->psr7RequestBuilder
+            ->withRouteParameter('offerType', $offerType)
+            ->withRouteParameter('offerId', self::OFFER_ID)
+            ->withJsonBodyFromArray([
+                'label' => self::LABEL_NAME,
+                'offers' => [
```
LucWollants, line 62: I don't think this is property `offers` is needed.

tests/Http/Offer/AddLabelFromJsonBodyRequestHandlerTest.php
```diff
@@ -0,0 +1,133 @@
...
+            ])
+            ->build('POST');
+
+        $this->assertCallableThrowsApiProblem(
+            ApiProblem::urlNotFound('The label should match pattern: ^[^;]{2,255}$'),
+            fn () => $this->addLabelFromJsonBodyRequestHandler->handle($addLabelFromJsonBodyRequest)
+        );
+    }
+
+    public function offerTypeDataProvider(): array
```
LucWollants, line 112: This provides more than just an offer type, maybe just call it `labelProvider` because that is the most important resource to provide.

tests/Http/Offer/AddLabelFromJsonBodyRequestHandlerTest.php
JonasVHG: When possible I've looked at the old tests => https://github.com/cultuurnet/udb3-silex/blob/11df7ea144d44070f568bc7f429e3630e5775df7/tests/Http/Offer/EditOfferRestControllerTest.php#L114 
It seemed prudent, to test the same body.
But looking at `LabelJSONDeserializer` it indeed seems useless.

LucWollants (approved):

PR #1130, opened 2022-09-06
Docker for local dev on m1

steftrenson:
### Added
- Docker support to run the project locally.


docker/docker-compose.yml.dist
```diff
@@ -0,0 +1,73 @@
...
+    volumes:
+      - 'udb3-mysql:/var/lib/mysql'
+    networks:
+      uitdatabank:
+        aliases:
+          - mysql.uitdatabank.dev
+    ports:
+      - "3306:3306"
+
+  rabbitmq:
```
LucWollants, line 36: Maybe we need to remove `rabbitmq` because it's not used on M1, or should we try it on another Mac?

docker/.env.dist
```diff
@@ -0,0 +1,7 @@
+MYSQL_ROOT_PASSWORD=vagrant
+MYSQL_DATABASE=udb3
+MYSQL_USER=vagrant
+MYSQL_PASSWORD=vagrant
+RABBITMQ_VHOST=udb3-vagrant
```
LucWollants, line 5: If we remove `RabbitMQ` this can also move.

docker/docker.md
```diff
@@ -0,0 +1,8 @@
+In your config.yml file, you have to change some of the hosts to work with Docker instead of Vagrant.
+You'll need to change the following lines to work with docker hosts:
+- url
```
LucWollants, line 3: Maybe we need to summarize the values that should be used?

LucWollants (changes requested):
I added a few smaller suggestions

LucWollants (approved):
👌

PR #1132, opened 2022-09-06
III-4917 Remove OfferEditingServices

JonasVHG:
### Changed

- Make `EventOrganizerRelationService` & `PlaceOrganizerRelationService` use a `CommandBus` instead of an `OfferEditingServiceInterface`

### Removed

- `EventEditingServiceInterface`, `EventOrganizerRelationService`, `DefaultOfferEditingService`, `OfferEditingServiceInterface` & tests

---
Ticket: https://jira.uitdatabank.be/browse/III-4917


LucWollants (approved):
Nice cleanup!

I wonder now if we have acceptance tests that verify the removal of an organizer and the effect on the related places and events? If not I think it should be added as acceptance tests.

PR #1135, opened 2022-09-08
Improvements to work with Docker container

steftrenson:
### Added
- Makefile to more easily work with Docker

### Changed
- Working dir of PHP Docker container is now the root of the project

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,0 +1,4 @@
+.PHONY: ci
+
+ci:
+	docker exec -it php.uitdatabank composer ci
--- a/docker/docker-compose.yml.dist
+++ b/docker/docker-compose.yml.dist
@@ -7,7 +7,7 @@ services:
     container_name: php.uitdatabank
     volumes:
       - .:/var/www/html:cached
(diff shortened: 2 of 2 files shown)


docker/Makefile
```diff
@@ -0,0 +1,4 @@
+.PHONY: ci
```
LucWollants, line 1: FYI There is also a make file inside the `udb3-vagrant` root folder. But it seems good to add the docker commands to this makefile.

steftrenson: Yeah, I think it makes more sense to have a separate Makefile for Docker related commands.

docker/Makefile
```diff
@@ -0,0 +1,16 @@
+.PHONY: ci stan cs cs-fix test
+
+ci:
+	docker exec -it php.uitdatabank composer ci
```
LucWollants, line 4: What is the preferred way? Using `docker exec` or `docker compose exec`?

LucWollants: We use `docker exec`

PR #1142, opened 2022-09-14
III-4654/III-4913 export JsonLd with calsum

JonasVHG:
### Added

- `CalendarSummaryWithFormatterRepository` & tests

### Changed

- make `$include` & `CalendarSummaryRepositoryInterface` parameters non nullable in Exports.
- Added `CalendarSummaryRepositoryInterface` to `JSONLDEventFormatter`, `JSONLDFileWriter` & `JSONLDFileFormat`
- Improved type hinting

### Fixed

- Exports via JsonLd again have a `calendarSummary`

### Removed

- `HttpCalendarSummaryRepository` & tests

---
Ticket: https://jira.uitdatabank.be/browse/III-4654 / https://jira.uitdatabank.be/browse/III-4913


src/EventExport/Format/JSONLD/JSONLDEventFormatter.php
```diff
@@ -104,17 +115,32 @@ function ($term) use ($includedTerms) {
...
             }
 
-
             $event = json_encode($eventObject);
         }
 
         return $event;
     }
+
+    private function addCalendarSummary(string $eventId): string
```
LucWollants, line 137: This does a `get`

src/EventExport/Format/JSONLD/JSONLDEventFormatter.php
```diff
@@ -104,17 +115,32 @@ function ($term) use ($includedTerms) {
...
+    {
+        try {
+            $calendarSummary = $this->calendarSummaryRepository->get($eventId, ContentType::plain(), Format::sm());
+        } catch (SummaryUnavailableException $exception) {
+            $calendarSummary = $exception->getMessage();
```
LucWollants, line 142: I would leave the summary empty instead of putting the exception message inside the exported JSON.

LucWollants (changes requested):
I have several general remarks:
- When changing legacy code I would take the effort to keep doing boy scouting work
- Many of the constructor parameters are nullable, but the provider does inject all the parameters, so I would make the parameters required
- There is also this ticket https://jira.uitdatabank.be/browse/III-4913 which advises to use another approach

bertramakers:
> There is also this ticket https://jira.uitdatabank.be/browse/III-4913 which advises to use another approach

Maybe we can keep using the `CalendarSummaryRepository` in this PR for now, and refactor that to just use the calendar summary library instead of fetching it via HTTP in both export formatters (JSON & Tabular) in III-4913?

LucWollants:
> > There is also this ticket https://jira.uitdatabank.be/browse/III-4913 which advises to use another approach
> 
> Maybe we can keep using the `CalendarSummaryRepository` in this PR for now, and refactor that to just use the calendar summary library instead of fetching it via HTTP in both export formatters (JSON & Tabular) in III-4913?

Sounds like a good plan. I will do a quick check to see how much work it will be to already replace all the nullable arguments which make the code less readable and maintainable.

src/EventExport/CalendarSummary/CalendarSummaryWithFormatterRepository.php
```diff
@@ -0,0 +1,31 @@
...
+namespace CultuurNet\UDB3\EventExport\CalendarSummary;
+
+use CultuurNet\CalendarSummaryV3\CalendarHTMLFormatter;
+use CultuurNet\CalendarSummaryV3\CalendarPlainTextFormatter;
+use CultuurNet\CalendarSummaryV3\Offer\Offer;
+use CultuurNet\UDB3\ReadModel\DocumentRepository;
+
+final class CalendarSummaryWithFormatterRepository implements CalendarSummaryRepositoryInterface
+{
+    private DocumentRepository $repository;
```
LucWollants, line 14: I think the name can be more explicit, this is an `eventRepository` or `eventJsonRepository`

src/EventExport/CalendarSummary/CalendarSummaryWithFormatterRepository.php
```diff
@@ -0,0 +1,31 @@
...
+
+    public function __construct(DocumentRepository $repository)
+    {
+        $this->repository = $repository;
+    }
+
+    public function get(string $offerId, ContentType $type, Format $format): string
```
LucWollants, line 21: I think in this context it is always an `eventId`

LucWollants (changes requested):
- Do we still use `HttpCalendarSummaryRepository`? If not it can be removed.
- I think we don't test the calendar summary export, would that be possible to test?

LucWollants (approved):

PR #1143, opened 2022-09-14
Use getCardSystemsForOrganizer

LucWollants:
### Fixed
- Use getCardSystemsForOrganizer

---
Ticket: https://jira.uitdatabank.be/browse/III-4940

### DIFF
--- a/src/UiTPASService/Controller/GetCardSystemsFromOrganizerRequestHandler.php
+++ b/src/UiTPASService/Controller/GetCardSystemsFromOrganizerRequestHandler.php
@@ -22,9 +22,9 @@ final class GetCardSystemsFromOrganizerRequestHandler implements RequestHandlerI
 
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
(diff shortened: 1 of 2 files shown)


src/UiTPASService/Controller/GetCardSystemsFromOrganizerRequestHandler.php
```diff
@@ -24,7 +24,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
     {
         $eventId = (new RouteParameters($request))->getOrganizerId();
```
JonasVHG, line 25: Maybe it's best to rename the variable to $organizerId

JonasVHG (approved):
I would also rename the variable from $eventId to $organzerId for clarity but It looks good otherwise.

src/UiTPASService/Controller/GetCardSystemsFromOrganizerRequestHandler.php
LucWollants: Good remark!

PR #1144, opened 2022-09-15
III-4890 remove CdbXmlProxy

JonasVHG:
### Removed

- `CdbXmlProxy`

---
Ticket: https://jira.uitdatabank.be/browse/III-4890


LucWollants (approved):
Don't forget to remove the config from `udb3-vagrant` and `appconfig`. I think we can even merge those config changes because of this line `isset($app['config']['cdbxml_proxy']`

Can you verify this locally on master (remove the config and make sure booting and accessing the URL still works)?

JonasVHG:
> Don't forget to remove the config from `udb3-vagrant` and `appconfig`. I think we can even merge those config changes because of this line `isset($app['config']['cdbxml_proxy']`
> 
> Can you verify this locally on master (remove the config and make sure booting and accessing the URL still works)?

The Url cdbxml.uitdatabank.dev still works. but you get a 404 event not found(which is the same as on production cause for new events cdbxml is no longer created: eg. https://cdbxml.uitdatabank.be/event/e7ae3e96-6ca8-4106-8faa-d552d0ff2688 )

PR #1145, opened 2022-09-16
III-4941 Prevent updating related events

LucWollants:
### Changed
- Prevent updating related events when updating the popularity score of a place.

Note: maybe there are more general solutions for this but I took a straightforward approach to fix the queue issue.

---
Ticket: https://jira.uitdatabank.be/browse/III-4941

### DIFF
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -8,6 +8,7 @@ use Broadway\Domain\DomainEventStream;
 use Broadway\EventHandling\EventBus;
 use CultuurNet\UDB3\EventSourcing\DomainMessageBuilder;
 use CultuurNet\UDB3\Offer\OfferType;
+use CultuurNet\UDB3\Place\Events\PlaceProjectedToJSONLD;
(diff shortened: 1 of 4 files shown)


bertramakers:
As discussed on Slack this can be fixed with an existing approach used in the `ReplayCommand`: https://github.com/cultuurnet/udb3-silex/blob/master/app/Console/ReplayCommand.php#L112-L116

In this case we can start intercepting "EventProjectedToJSONLD" messages right before the PlaceProjectedToJSONLD message is published on the event bus, and stop intercepting right after. (So that other "EventProjectedToJSONLD" messages that _do_ need to be published later are effectively published)

For example:

```
if ($projectedEvent instanceof PlaceProjectedToJSONLD) {
    InterceptingMiddleware::startIntercepting(
        fn (DomainMessage $message) => $message->getPayload() instanceof EventProjectedToJSONLD
    );
}

$this->eventBus->publish(
    new DomainEventStream(
        [(new DomainMessageBuilder())->create($projectedEvent)]
    )
);

InterceptingMiddleware::stopIntercepting();
```

LucWollants:
Replaced with PR https://github.com/cultuurnet/udb3-silex/pull/1149

PR #1146, opened 2022-09-16
III-4882 refactor edit role rest controller

steftrenson:
### Added
- Add new request handler `AddConstraintRequestHandler`
- Add new request handler `CreateRoleRequestHandler`
- Add new request handler `UpdateRoleRequestHandler`
- Added `QueryValidatingRequestBodyParser`
- Added new request handler `UpdateConstraintRequestHandler`
- Added new request handler `DeleteConstraintRequestHandler`

### Changed
- `CreateRole` expects string instead of `StringLiteral`
- `RenameRole` expects string instead of `StringLiteral`

### Removed
- Unused function rename on `RoleEditingServiceInterface`

---
Ticket: https://jira.uitdatabank.be/browse/III-4882


src/Http/Role/DeleteConstraintRequestHandler.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function __construct(CommandBus $commandBus)
+    {
+        $this->commandBus = $commandBus;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $roleId = new UUID($routeParameters->get('id'));
```
LucWollants, line 29: In your previous pull request for the reading part, you introduced `getRoleId`. Maybe reuse that?

src/Http/Role/UpdateRoleRequestHandler.php
```diff
@@ -0,0 +1,54 @@
...
+        $roleId = new UUID($routeParameters->get('id'));
+
+        $body = Json::decodeAssociatively($request->getBody()->getContents());
+
+        $this->commandBus->dispatch(new RenameRole($roleId, (string) $body['name']));
+
+        return new Response(StatusCodeInterface::STATUS_NO_CONTENT);
+    }
+
+    private function ensureContentTypeIsProvided(ServerRequestInterface $request): void
```
LucWollants, line 43: I wonder if we would need a 415 ApiProblem, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415

src/Role/Commands/CreateRole.php
```diff
@@ -25,6 +22,6 @@ public function __construct(
 
     public function getName(): StringLiteral
     {
-        return $this->name;
+        return new StringLiteral($this->name);
```
LucWollants, line 25: This method is only used twice, once in real code and once in a test. So it's a opportunity to remove `StringLiteral` from this file

src/Role/Commands/RenameRole.php
```diff
@@ -9,25 +9,19 @@
...
     }
 
-    /**
-     * @return StringLiteral
-     */
-    public function getName()
+    public function getName(): StringLiteral
     {
-        return $this->name;
+        return new StringLiteral($this->name);
```
LucWollants, line 25: Also only used once in actual code.

src/Role/Services/DefaultRoleEditingService.php
```diff
@@ -47,24 +46,14 @@ public function create(StringLiteral $name): UUID
...
-            $name
+            $name->toNative()
         );
 
         $this->commandBus->dispatch($command);
 
         return $uuid;
     }
 
-    public function rename(UUID $uuid, StringLiteral $name): void
```
LucWollants, line 58: I think more methods can be removed? Or will you remove the complete file in the final pull request?

src/Http/Role/UpdateRoleRequestHandler.php
bertramakers: It already exists in our list of error types: https://docs.publiq.be/docs/errors/baae4bf3cece5-header#unsupported-media-type

So should just be a matter of adding it to the `ApiProblem` class in this app

tests/Http/Role/AddConstraintRequestHandlerTest.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public function it_throws_when_query_is_not_given(): void
+    {
+        $roleId = '1ed1588b-a771-44ce-bac0-8f19f09a7d0f';
+        $request = (new Psr7RequestBuilder())
+            ->withRouteParameter('id', $roleId)
+            ->withJsonBodyFromArray([])
+            ->build('POST');
+
+        $this->commandBus->record();
```
LucWollants, line 46: The record is not needed here.

tests/Http/Role/CreateRoleRequestHandlerTest.php
```diff
@@ -0,0 +1,75 @@
...
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_throws_invalid_argument_exception_when_an_empty_name_is_given(): void
+    {
+        $request = (new Psr7RequestBuilder())
+            ->withJsonBodyFromArray(['name' => ''])
```
LucWollants, line 44: Makes me wonder what happens is `name` is completely missing?

src/Role/Services/DefaultRoleEditingService.php
steftrenson: They have to stay there because the current controller still needs them. They will indeed be removed in the final PR when the switch is made to use the new handlers iso the controller. This function specifically was never used, even in the current controller which dispatch directly on the commandbus

LucWollants (approved):
Only a few minor remarks. Also nice work on removing `StringLiteral`

src/Role/Services/DefaultRoleEditingService.php
LucWollants: True! 😅

src/Http/Role/DeleteConstraintRequestHandler.php
steftrenson: That will require me to change the routes. I was planning on doing that when I made the switch to requestHandlers in the routing provider.

steftrenson: Come to think of it, I can already implement it here, as it's not being used anyway. How do we feel about that getRoleId returning a UUID VO and throwing an API problem when it's invalid. In other places it already throws an invalidArgumentException because of this. Seems like it conceptually belongs on the routeParameters class.

cc @bertramakers

tests/Http/Role/CreateRoleRequestHandlerTest.php
steftrenson: I tested it before but removed the test, added it again.

src/Http/Role/DeleteConstraintRequestHandler.php
LucWollants: I think this code path is safe to use `UUID` as the typed return type because roles are "new" in UDB3 and will always match the `UUID` pattern. On other paths for example labels and events can have route parameters that don't match their VO.

tests/Http/Role/AddConstraintRequestHandlerTest.php
steftrenson: I'm testing if the command bus didn't receive commands, so it should record or it would always return empty, even if it dispatched commands. I've moved the record step to the setup function, as I think it shouldn't be part of an individual test case but just a given.

src/Role/Commands/CreateRole.php
steftrenson: Done, I could take it further but it cascades.

steftrenson:
@LucWollants Handled the feedback and changed a couple of other things as well. The handlers don't throw exceptions anymore but apiProblems, since I think that's the desired outcome, right?

Changes are gest reviewed commit by commit.

src/Http/Role/DeleteConstraintRequestHandler.php
steftrenson: All right, I'll make that change in the last PR

src/Http/ApiProblem/ApiProblem.php
```diff
@@ -303,6 +304,15 @@ public static function bodyMissing(): self
         );
     }
 
+    public static function unsupportedMediaType(): self
+    {
+        return self::create(
+            'https://docs.publiq.be/docs/errors/baae4bf3cece5-header#unsupported-media-type',
```
LucWollants, line 310: You should be able to use `https://api.publiq.be/probs/header/unsupported-media-type` here

src/Http/Role/DeleteConstraintRequestHandler.php
```diff
@@ -26,7 +26,7 @@ public function __construct(CommandBus $commandBus)
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
         $routeParameters = new RouteParameters($request);
-        $roleId = new UUID($routeParameters->get('id'));
+        $roleId = new UUID($routeParameters->getRoleId());
```
LucWollants, line 29: I think there is one other place where `getRoleId` can be used, inside `UpdateRoleRequestHandler`

LucWollants (approved):
Very readable code! Only 2 small remarks left.

PR #1147, opened 2022-09-19
III-4882 refactor edit role rest controller (2/3)

steftrenson:
### Added
- Added new request handler `AddLabelToRoleRequestHandler`
- Added new request handler `AddPermissionToRoleRequestHandler`
- Added new request handler `AddRoleToUserRequestHandler`
- Added new request handler `DeleteRoleRequestHandler`
- Added new request handler `RemoveLabelFromRoleRequestHandler`
- Added new request handler `RemovePermissionFromRoleRequestHandler`
- Added new request handler `RemoveRoleFromUserRequestHandler`

### Changed
- `AddUser` and `RemoveUser` command now expect string iso `StringLiteral` when constructing

---
Ticket: https://jira.uitdatabank.be/browse/III-4882

### DIFF
--- a/src/Http/Role/AddLabelToRoleRequestHandler.php
+++ b/src/Http/Role/AddLabelToRoleRequestHandler.php
@@ -19,6 +19,8 @@ use Slim\Psr7\Response;
 
 final class AddLabelToRoleRequestHandler implements RequestHandlerInterface
 {
+    use GetLabelIdFromRouteParameters;
+
     private CommandBus $commandBus;
 
     private ReadRepositoryInterface $labelRepository;
@@ -40,18 +42,7 @@ final class AddLabelToRoleRequestHandler implements RequestHandlerInterface
             throw ApiProblem::roleNotFound($roleId);
         }
 
-        $labelIdentifier = $routeParameters->get('labelIdentifier');
-        try {
-            $labelId = new UUID($labelIdentifier);
-        } catch (InvalidArgumentException $exception) {
-            $entity = $this->labelRepository->getByName($labelIdentifier);
-
-            if ($entity === null) {
-                throw ApiProblem::blank('There is no label with identifier: ' . $labelIdentifier, 404);
-            }
-
-            $labelId = new UUID($entity->getUuid()->toString());
-        }
+        $labelId = $this->getLabelId($routeParameters);
 
         $this->commandBus->dispatch(new AddLabel($roleId, $labelId));
 
--- a/src/Http/Role/AddLabelToRoleRequestHandler.php
+++ b/src/Http/Role/AddLabelToRoleRequestHandler.php
@@ -6,7 +6,6 @@ namespace CultuurNet\UDB3\Http\Role;
 
 use Broadway\CommandHandling\CommandBus;
 use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
-use CultuurNet\UDB3\Http\ApiProblem\SchemaError;
 use CultuurNet\UDB3\Http\Request\RouteParameters;
 use CultuurNet\UDB3\Label\ReadModels\JSON\Repository\ReadRepositoryInterface;
 use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
@@ -38,7 +37,7 @@ class AddLabelToRoleRequestHandler implements RequestHandlerInterface
         try {
             $roleId = new UUID($roleId);
         } catch (InvalidArgumentException $e) {
-            throw ApiProblem::bodyInvalidData(new SchemaError('/', 'The provided role id is not a uuid'));
+            throw ApiProblem::invalidUUID('roleId');
         }
 
         $labelIdentifier = $routeParameters->get('labelIdentifier');
--- a/src/Http/Role/AddLabelToRoleRequestHandler.php
+++ b/src/Http/Role/AddLabelToRoleRequestHandler.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Role;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Http\ApiProblem\SchemaError;
+use CultuurNet\UDB3\Http\Request\RouteParameters;
+use CultuurNet\UDB3\Label\ReadModels\JSON\Repository\ReadRepositoryInterface;
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\Role\Commands\AddLabel;
+use Fig\Http\Message\StatusCodeInterface;
+use InvalidArgumentException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Slim\Psr7\Response;
+
+class AddLabelToRoleRequestHandler implements RequestHandlerInterface
+{
+    private CommandBus $commandBus;
+
+    private ReadRepositoryInterface $labelRepository;
+
+    public function __construct(CommandBus $commandBus, ReadRepositoryInterface $labelRepository)
+    {
+        $this->commandBus = $commandBus;
+        $this->labelRepository = $labelRepository;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $roleId = $routeParameters->getRoleId();
+
+        try {
+            $roleId = new UUID($roleId);
+        } catch (InvalidArgumentException $e) {
+            throw ApiProblem::bodyInvalidData(new SchemaError('/', 'The provided role id is not a uuid'));
+        }
+
+        $labelIdentifier = $routeParameters->get('labelIdentifier');
+        try {
+            $labelId = new UUID($labelIdentifier);
+        } catch (InvalidArgumentException $exception) {
+            $entity = $this->labelRepository->getByName($labelIdentifier);
+
+            if ($entity === null) {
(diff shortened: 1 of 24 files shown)


src/Http/Role/AddLabelToRoleRequestHandler.php
```diff
@@ -0,0 +1,60 @@
...
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\Role\Commands\AddLabel;
+use Fig\Http\Message\StatusCodeInterface;
+use InvalidArgumentException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Slim\Psr7\Response;
+
+class AddLabelToRoleRequestHandler implements RequestHandlerInterface
```
bertramakers, line 20: Nit: We try to make every class either `final` or `abstract`. We don't have a rule for that in the coding standards yet because we still have a lot of classes that are extended but cannot be made abstract without a refactor.

You can add it to your PHP templates in PHPStorm so it gets added automatically if you want.

src/Http/Role/AddLabelToRoleRequestHandler.php
```diff
@@ -0,0 +1,60 @@
...
+        $routeParameters = new RouteParameters($request);
+        $roleId = $routeParameters->getRoleId();
+
+        try {
+            $roleId = new UUID($roleId);
+        } catch (InvalidArgumentException $e) {
+            throw ApiProblem::roleNotFound($roleId);
+        }
+
+        $labelIdentifier = $routeParameters->get('labelIdentifier');
```
LucWollants, line 43: Maybe `labelId` is a better parameter name as it matches the naming convention of other parameters better.

src/Http/Role/AddLabelToRoleRequestHandler.php
```diff
@@ -0,0 +1,60 @@
...
+
+    private ReadRepositoryInterface $labelRepository;
+
+    public function __construct(CommandBus $commandBus, ReadRepositoryInterface $labelRepository)
+    {
+        $this->commandBus = $commandBus;
+        $this->labelRepository = $labelRepository;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
```
LucWollants, line 32: A lot of the `handle` is duplicated inside `RemoveLabelFromRoleRequestHandler`, maybe a trait can help?

src/Http/Role/AddLabelToRoleRequestHandler.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        $labelIdentifier = $routeParameters->get('labelIdentifier');
+        try {
+            $labelId = new UUID($labelIdentifier);
+        } catch (InvalidArgumentException $exception) {
+            $entity = $this->labelRepository->getByName($labelIdentifier);
+
+            if ($entity === null) {
+                throw ApiProblem::blank('There is no label with identifier: ' . $labelIdentifier, 404);
```
LucWollants, line 50: The `blank` method is marked as `deprecated`

bertramakers, line 50: Is this `::blank()` type recently added? The `about:blank` API problem type may only be used in some specific cases where the error type can be deduced from the status code (for example 502 = always bad gateway, no further explanation needed). But even then it is allowed to use a more specific type than `about:blank`, and definitely in this case it should be more specific.

Given that this error is caused by a URL parameter, the type should probably be `::urlNotFound()`. I know it's a bit weird in this case because the route to handle this request exists, and the role also exists. But it's because the label name is sent as part of the URL instead of being a part of the body that makes this a weird situation.

bertramakers: cc @LucWollants @JonasVHG to avoid any confusion/assumptions about the `about:blank` API problem type

src/Http/Role/AddPermissionToRoleRequestHandler.php
```diff
@@ -0,0 +1,51 @@
...
+use CultuurNet\UDB3\Role\Commands\AddPermission;
+use CultuurNet\UDB3\Role\ValueObjects\Permission;
+use Fig\Http\Message\StatusCodeInterface;
+use InvalidArgumentException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Slim\Psr7\Response;
+
+class AddPermissionToRoleRequestHandler implements RequestHandlerInterface
```
LucWollants, line 20: Here is also some code duplication with `RemovePermissionFromRoleRequestHandler`

src/Http/Role/AddLabelToRoleRequestHandler.php
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            $labelId = new UUID($entity->getUuid()->toString());
+        }
+
+        $this->commandBus->dispatch(new AddLabel($roleId, $labelId));
+
+        return new Response(StatusCodeInterface::STATUS_NO_CONTENT);
```
LucWollants, line 58: We have a dedicated class `NoContentResponse` for this

src/Http/Role/RemoveLabelFromRoleRequestHandler.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        $labelIdentifier = $routeParameters->get('labelIdentifier');
+        try {
+            $labelId = new UUID($labelIdentifier);
+        } catch (InvalidArgumentException $exception) {
+            $entity = $this->labelRepository->getByName($labelIdentifier);
+
+            if ($entity === null) {
+                throw ApiProblem::blank('There is no label with identifier: ' . $labelIdentifier, 404);
```
bertramakers, line 50: (Same comment as https://github.com/cultuurnet/udb3-silex/pull/1147/files#r978490281)

src/Http/Role/AddLabelToRoleRequestHandler.php
bertramakers, line 58: Nit: Could be replaced with `CultuurNet\UDB3\Http\Response\NoContentResponse`

bertramakers (commented):
Just some small remarks, the overall changes look good!

src/Http/Role/AddLabelToRoleRequestHandler.php
bertramakers: Haha oops sorry for the double comments @steftrenson , looks like Luc and I reviewed at exactly the same time :)

src/Role/Commands/AddUser.php
```diff
@@ -4,6 +4,13 @@
...
 namespace CultuurNet\UDB3\Role\Commands;
 
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\StringLiteral;
+
 class AddUser extends AbstractUserCommand
 {
+    public function __construct(UUID $uuid, string $userId)
+    {
+        parent::__construct($uuid, new StringLiteral($userId));
```
LucWollants, line 14: A smart approach to make sure the new code does not use `StringLiteral`!

tests/Http/Role/AddLabelToRoleRequestHandlerTest.php
```diff
@@ -0,0 +1,169 @@
...
+    private AddLabelToRoleRequestHandler $handler;
+
+    private TraceableCommandBus $commandBus;
+
+    /**
+     * @var ReadRepositoryInterface | MockObject
+     */
+    private $labelRepository;
+
+    protected function setUp()
```
LucWollants, line 38: This method can have a `void` return type. PHPStorm helps me to spot these things ☺️

LucWollants (changes requested):
Just a few smaller remarks. I like the clean code and the extra effort to remove `StringLiteral`

src/Http/Role/AddLabelToRoleRequestHandler.php
steftrenson: Done.

src/Http/Role/AddLabelToRoleRequestHandler.php
steftrenson: I'll change that in part 3, as it changes the routing there I can do it in one go.

bertramakers: Is not either a label id of a label name? I think that's why it was called "identifier" (although it remains vague)

bertramakers: Note how it does `$entity = $this->labelRepository->getByName($labelIdentifier);` in the catch if it's not a valid UUID

bertramakers: `labelIdOrName` is maybe the clearest?

src/Http/Role/AddLabelToRoleRequestHandler.php
steftrenson: Extracted getLabelId logic in a trait. The roleId is refactored in part 3

src/Http/Role/AddLabelToRoleRequestHandler.php
steftrenson: The `::urlNotFound()` makes sense imo. I'll update it.

src/Http/Role/AddPermissionToRoleRequestHandler.php
steftrenson: Extracted it away to the RouteParameters

steftrenson:
@LucWollants @bertramakers : Thanks for the review. I've implemented the feedback.

src/Http/Role/AddLabelToRoleRequestHandler.php
bertramakers: Cleverly solved @steftrenson !

bertramakers (approved):
Nice work! ✅

LucWollants (approved):

PR #1149, opened 2022-09-19
III-4941 Intercept event projected events when place is projected

LucWollants:
### Changed
- Intercept event projected events when the popularity score of a place is updated. This should prevent updates to the related events.

---
Ticket: https://jira.uitdatabank.be/browse/III-4941

### DIFF
--- a/app/Console/ReindexOffersWithPopularityScore.php
+++ b/app/Console/ReindexOffersWithPopularityScore.php
@@ -25,18 +25,18 @@ class ReindexOffersWithPopularityScore extends Command
 
     private EventBus $eventBus;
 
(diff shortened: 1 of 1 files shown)


bertramakers (approved):
Looks good to me, just a small possible improvement I noticed: 
This command class is instantiated twice, once for events and once for places. And we only need to intercept the "related" `EventProjectedToJSONLD` messages in the case of places. So instead of starting/stopping the intercepting every time in `handleEvent()` by checking if the message is `PlaceProjectedToJSONLD`, I think we can start/stop it once in `execute()` before/after the `foreach()` when the injected `OfferType` in the constructor is `OfferType::place()`.

JonasVHG (approved):

PR #1150, opened 2022-09-20
III-4943 Bind routes in PsrRouterServiceProvider

LucWollants:
### Changed
- Bound routes with the new League Router inside `PsrRouterServiceProvider`
- Decode the params from the new League router which are URL encoded, make sure to use `rawurldecode` to prevent decoding the `+` sign to a ` ` space.

### Removed
- Removed various `ControllerProviderInterface` implementations

---

### Notes
- The new League Router does encode URL parameters
- The order of routes is now important, first dynamic routes then hardcoded routes
- Dynamic routes can not be used as a RouteGroup prefix

---
Ticket: https://jira.uitdatabank.be/browse/III-4943


bertramakers (approved):
Looks good to me! But we'll hold off with merging this until the refactored role endpoints are deployed, so we can deploy this separately.

JonasVHG (approved):

bertramakers:
Merging this now that the other commits on `master` have been deployed.

PR #1151, opened 2022-09-20
III-4890 SAPI3 proxy request handler

bertramakers:
### Added

- Added `ProxyRequestHandler` that can proxy a request to a different host
- Added routes for `GET /events`, `GET /places`, `GET /offers`, and `GET /organizers` to be proxied to SAPI3

### Removed

- Removed `Proxy` class and middleware and everything related to it

---
Ticket: https://jira.uitdatabank.be/browse/III-4890

Config PRs:

- https://github.com/cultuurnet/udb3-vagrant/pull/128
- https://github.com/cultuurnet/appconfig/pull/233

Note: I deliberately made this new Proxy RequestHandler much simpler than the old Proxy class/middleware, because the use case is much more limited now that we removed a lot of the old proxy endpoints recently.


bertramakers:
Thanks for helping fix the bug in [0952ecb](https://github.com/cultuurnet/udb3-silex/pull/1151/commits/0952ecbd451fb1acb0b0d5faa4a77c5b104d6fd3) @LucWollants 👌💯 I'm going to add an extra commit to add a comment to explain why the `Transfer-Encoding` should be removed, so it's still clear for people looking at this in the future

LucWollants (approved):
Very straight forward code. I also did a manual test which worked after removing `Transfer-Encoding` header

LucWollants (approved):

JonasVHG (approved):

PR #1158, opened 2022-09-23
III-4913 Remove calendarSummary property fallback in Excel exports

bertramakers:
### Fixed

- Tabular data exports no longer look for a non-existent `calendarSummary` property on the event JSON-LD if the calendar summary cannot be generated
- The related unit test now actually tests that the calendar summary is generated, instead of testing the fallback

---
Ticket: https://jira.uitdatabank.be/browse/III-4913


JonasVHG (approved):
Good catch.

PR #1161, opened 2022-09-27
III-4948 By-pass Silex router

bertramakers:
### Changed

- The PSR router is now called directly instead of running `$app->run()` in Silex
- We no longer log IP addresses in the metadata in the event store. See commit description [here](https://github.com/cultuurnet/udb3-silex/commit/c7ada07bff4f0daa0f8e4d35edd6cc26aab868de) for more info why.

### Removed

- `$app->error()` callback, which won't be called anymore now that we by-pass the Silex router
- Unused dependencies and custom classes related to routing

---
Ticket: https://jira.uitdatabank.be/browse/III-4948


composer.json
```diff
@@ -64,7 +64,6 @@
     "publiq/udb3-json-schemas": "dev-main",
     "ramsey/uuid": "^3.2.0",
     "rase/socket.io-emitter": "0.6.1",
-    "respect/validation": "~1.1",
```
bertramakers, line 67: Not really in scope but I noticed it was practically unused

JonasVHG (approved):

PR #1163, opened 2022-09-29
III-3888 Add a PSR-11 container bridge

bertramakers:
### Added

- `league/container` as a PSR-11 container implementation
- Classes to bridge the Silex `Application` class and the League `Container` class so they can look up services in each other if they are not found in their own container
- Auto-wiring on the League `Container`, so if a service is not found in either container it will be auto-wired if possible (for example classes without constructor arguments, or classes with constructor arguments that are found in either container)

### Changed

- Refactored the three service providers that were registered in `web/index.php` to League service providers on the new container as proof-of-concept that we can refactor service providers one by one
- Changed the namespace of `app/` to `CultuurNet\UDB3`, and moved all classes in the `CultuurNet\UDB3\Silex` namespace to `app/Silex` so we can move them to the `CultuurNet\UDB3` namespace gradually

---
Ticket: https://jira.uitdatabank.be/browse/III-3888


JonasVHG (approved):
Looks great👏

JonasVHG (approved):

PR #1164, opened 2022-09-29
III-4863 psr 11 amqp

JonasVHG:
### Changed

- Refactored `AMQPPublisherServiceProvider` &  `AMQPConnectionServiceProvider` service providers to League service providers 

---
Ticket: https://jira.uitdatabank.be/browse/III-4863


bertramakers (changes requested):
Looks good so far but a couple of small things are still missing:

- You'll need to replace the `$app->register()` call with these service providers with `$container->register()`, as the new service providers only work on the new container (and we want to get rid of `$app` later)
- Some config is currently passed to the service provider as the second argument in the `$app->register()` call. This won't work anymore in the new service container. So instead of doing e.g. `$container->get('amqp.connection.host')`, you'll need to do `$container->get('config')['amqp']['host']`

See the `register()` call here: https://github.com/cultuurnet/udb3-silex/blob/ecad603f54dfc407248a6c56ddd81b4158f2522e/bootstrap.php#L824-L833 (and note how the second argument creates a list of config parameters and makes them accessible to the service provider under different names, which you won't be able to use anymore)

LucWollants (approved):
Looks good!

Maybe a local variable `$config` which holds the return from `$container->get('config')` can improve readability and also avoids some method calls.

bertramakers (approved):
LGTM! @LucWollants maybe we should look for an improvement for the 'config' service later when all service providers are moved. Then we avoid multiple approaches in the different PRs now, and we can easily find all usages of `$container->get('config')` later and replace them with something that's easier to work with. In SAPI3 we added an extra method to the base service provider to easily get config parameters. We could do something similar here.

PR #1168, opened 2022-09-30
III-4963 make Auth0ServiceProvider PSR11

JonasVHG:
### Changed

- Refactored `Auth0ServiceProvider` service providers to League service providers 

---
Ticket: https://jira.uitdatabank.be/browse/III-4963


app/Auth0/Auth0ServiceProvider.php
```diff
@@ -0,0 +1,58 @@
...
+                    new CacheRepository(
+                        $container->get('config')('auth0-management-token')
+                    )
+                );
+                return $provider->token();
+            }
+        );
+
+        $container->addShared(
+            'auth0.management-token',
```
LucWollants, line 49: This seems to have the same name as the previous container entry.

bertramakers: That should be `Auth0UserIdentityResolver::class` indeed

bertramakers (approved):

PR #1170, opened 2022-09-30
III-4965 psr11 CommandHandling

JonasVHG:
### Changed

- Converted `CommandBusServiceProvider` From a Silex to a PSR-11 Container
- `ContextFactory::createFromGlobals()` now expects a PSR-11 `ContainerInterface` instead of `Silex\Application`
- Move `CommandBusServiceProvider`, `ContextDecoratedCommandBus`, `ContextFactory`,  `LazyLoadingCommandBus`, `RetriedCommandFailed` & `RetryingCommandBus` outside of the Silex Folder
- Move test for `RetryingCommandBus` outside of `tests/Silex` folder

### Removed

- `tests/Silex` folder cause it contained no more tests
- `resque_command_bus_factory`, by defining the resque command buses explicitly so we could more easily remove the `$app->extend()` calls on them

---
Ticket: https://jira.uitdatabank.be/browse/III-4965


bertramakers:
@steftrenson @LucWollants can one of you review this? Since I helped get this to the finish line, so my review would be a bit biased

app/Silex/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -124,41 +123,27 @@ function (): CompositeCommandValidator {
...
-                                $app[$queueName . '_command_bus_factory'],
-                                $app
-                            ),
-                            $app[$queueName . '_command_validator']
-                        );
+                        return $commandBus;
                     }
                 );
-
-                $app[$queueName . '_command_bus_out'] = $app->share(
```
steftrenson, line 135: Why is the `_command_bus_out` part removed?

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -253,6 +253,13 @@ function () use ($queueName, $container) {
                             );
                         }
                     );
+
+                    $container->addShared(
```
steftrenson: Fixes my previous comment 👍

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -36,6 +35,10 @@ protected function getProvidedServiceNames(): array
             'event_command_bus',
             'event_command_validator',
             'resque_command_bus_factory',
```
steftrenson: This can be removed now, right?

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -216,31 +219,44 @@ function (): CompositeCommandValidator {
...
-                            return new ValidatingCommandBusDecorator(
-                                new ContextDecoratedCommandBus(
-                                    self::createResqueCommandBus($queueName, $container),
-                                    $container
-                                ),
-                                new CompositeCommandValidator()
-                            );
-                        }
-                    );
+        $container->addShared(
```
steftrenson, line 214: Like this approach a lot more than the looping mechanism. It's a lot more explicit and thus readable imo.

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -31,7 +31,6 @@ protected function getProvidedServiceNames(): array
             'authorized_command_bus',
             'event_command_bus',
             'event_command_validator',
-            'resque_command_bus_factory',
```
steftrenson: Again fixes a previous comment

steftrenson (commented):
LGTM

app/CommandHandling/ContextFactory.php
```diff
@@ -80,4 +69,15 @@ public static function prepareForLogging(Metadata $metadata): Metadata
 
         return new Metadata($metadata);
     }
+
+    public static function createFromGlobals(ContainerInterface $container): Metadata
```
steftrenson, line 73: My 2 cents: I wouldn't pass the container to classes but inject the properties directly. Unless there's a reason that can't be done?

bertramakers: In general I agree, but in some cases we do this because some things like `CurrentUser` are not always set immediately but only when a Request middleware is executed. So we sometimes need to delay the fetching until the last moment. I'm not 100% sure if it's needed here, but I figured it would be best to leave it like it is (but refactor it to use `ContainerInterface` instead of `Application`), and then revisit it later to figure out the details

bertramakers: (Not sure if it was clear but the `createFromGlobals()` already existed, Jonas removed it and injected the properties directly into the constructor to get rid of the `Application` dependency, but I re-added it to be safe and refactored it to use `ContainerInterface` instead)

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -0,0 +1,269 @@
...
+                        $commandBus->subscribe($container->get(LabelServiceProvider::COMMAND_HANDLER));
+                    }
+                );
+
+                return $commandBus;
+            }
+        );
+
+        $container->addShared(
+            'event_command_validator',
```
LucWollants, line 208: Should this `event_command_validator` be exposed for reuse?

LucWollants (approved):
I already approved, I only had two minor remarks.

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -0,0 +1,269 @@
...
+            'bulk_label_offer_command_bus_out',
+            function () use ($container) {
+                $commandBus = self::createResqueCommandBus('bulk_label_offer', $container);
+                $commandBus->subscribe($container->get('bulk_label_offer_command_handler'));
+                return $commandBus;
+            }
+        );
+    }
+
+    public static function createResqueCommandBus(string $queueName, ContainerInterface $container): ResqueCommandBus
```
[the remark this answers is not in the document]
bertramakers: Ah no. I first intended to use it outside of this class but I used another approach in the end, so I'll change it

app/CommandHandling/CommandBusServiceProvider.php
bertramakers: No it can probably be moved inline indeed. But I discussed with @JonasVHG and @steftrenson to not focus on that too much now, since it's often easier to review these big refactors if the list of services stays the same. But in this PR we already did some bigger changes anyway so I'll move it inline in this case

app/CommandHandling/CommandBusServiceProvider.php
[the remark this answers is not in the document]
bertramakers: See [609c69a](https://github.com/cultuurnet/udb3-silex/pull/1170/commits/609c69af9460b0c425d5a6411378e363d47d8410) and [b2f5a29](https://github.com/cultuurnet/udb3-silex/pull/1170/commits/b2f5a29caa578df96c1a4cd35bfc100fb8604e50)

bertramakers:
Merging this since I think I addressed all comments

PR #1171, opened 2022-10-03
III-4315 Use PSR-11 container in ./bin/udb3.php console app

bertramakers:
### Changed

- Refactored `ConsumeCommand` to no longer depend on the Silex Application object
- Moved commands to PSR-11 container ServiceProvider
- The Console Application now uses a `ContainerCommandLoader` so that only the command that is executed gets fetched from the container. This way only that command and its dependencies are instantiated, compared to earlier when all commands and all their dependencies were always instantiated just to run one command.

### Removed

- Unused `dbal_connection:keepalive` service (after refactoring of `ConsumeCommand`)

---
Ticket: https://jira.uitdatabank.be/browse/III-4315


bin/udb3.php
```diff
@@ -52,106 +20,66 @@
...
-$consoleApp->add(new UpdateUniqueLabels($app['dbal_connection']));
-$consoleApp->add(new UpdateUniqueOrganizers($app['dbal_connection'], new WebsiteNormalizer()));
-$consoleApp->add(new RemoveFacilitiesFromPlace($app['event_command_bus'], $app[Sapi3SearchServiceProvider::SEARCH_SERVICE_PLACES]));
-$consoleApp->add(new RemoveLabelOffer($app['dbal_connection'], $app['event_command_bus']));
-$consoleApp->add(new RemoveLabelOrganizer($app['dbal_connection'], $app['event_command_bus']));
-
-$consoleApp->add(new ImportOfferAutoClassificationLabels($app['dbal_connection'], $app['event_command_bus']));
-
-$consoleApp->add(new ReplaceNewsArticlePublisher($app['dbal_connection']));
+$commands = [
```
LucWollants, line 39: Would it be an option to expose the service list from `ConsoleServiceProvider` and use that to generate the command map?

LucWollants (approved):

bin/udb3.php
bertramakers: Yes, good idea!

bertramakers: See [75de095](https://github.com/cultuurnet/udb3-silex/pull/1171/commits/75de095f7b2f49b1f57c305be6845f187d3292a4)

LucWollants: Nice approach!

PR #1177, opened 2022-10-05
Iii 4966 culture feed provider

steftrenson:
### Fixed

---


app/Silex/CultureFeed/CultureFeedServiceProvider.php
```diff
@@ -4,28 +4,33 @@
...
+            function () use ($container): CultureFeed_DefaultOAuthClient {
+                $oauthClient = new CultureFeed_DefaultOAuthClient(
+                    $container->get('config')['uitid']['consumer']['key'],
+                    $container->get('config')['uitid']['consumer']['secret'],
                 );
-                $oauthClient->setEndpoint($app['culturefeed.endpoint']);
+                $oauthClient->setEndpoint($container->get('config')['uitid']['base_url']);
 
-                return new \CultureFeed($oauthClient);
+                return $oauthClient;
```
bertramakers: This still needs to be wrapped in `new \CultureFeed()`.

(Some context: https://github.com/cultuurnet/culturefeed-php/ is an old PHP library that existed before UiTdatabank v3, which we used as an SDK to communicate with UiTdatabank v2, and also UiTPAS and UiTID. These days it's only still used in a handful of places in this app)

bertramakers (changes requested):
Everything looks great, except for 1 small (but important) bug

JonasVHG (approved):

app/Silex/CultureFeed/CultureFeedServiceProvider.php
steftrenson: Damn, can't believe I missed that. Nice catch!

bertramakers (approved):
Perfect now! 👌

app/Silex/CultureFeed/CultureFeedServiceProvider.php
bertramakers: It's understandable, it should become more obvious once we replace those string literals with FQNs of the classes that we register :)

bertramakers:
@steftrenson I'm going to go ahead and merge this so we avoid potential merge conflicts when refactoring other service providers (this PR already had one after merging #1178)

PR #1185, opened 2022-10-06
III-4977 psr11 metadata

JonasVHG:
### Changed

- `MetadataServiceProvider` is now shared via the Container instead of the Silex App.
- `MetadataServiceProvider` is moved outside the Silex folder.

---
Ticket: https://jira.uitdatabank.be/browse/III-4977


app/Silex/Metadata/MetadataServiceProvider.php
```diff
@@ -8,29 +8,38 @@
...
+            'event_stream_metadata_enricher',
+            'command_bus_event_dispatcher',
+        ];
+    }
+
+    public function register(): void
+    {
+        $container = $this->getContainer();
+
+        $container->addShared('context', '');
```
JonasVHG, line 30: I've changed this from null to empty string. Because adding a null causes different behaviour in the PSR-Container vs the Silex Application.

app/Metadata/MetadataServiceProvider.php
```diff
@@ -2,35 +2,44 @@
...
-                    function () use ($app) {
+                    function () use ($container) {
                         // Create a default context from application globals.
-                        $container = $app->getLeagueContainer();
                         $context = ContextFactory::createFromGlobals($container);
 
                         // Allow some processes to overwrite the context, like resque workers.
-                        if ($app['context'] instanceof Metadata) {
-                            $context = $app['context'];
+                        if ($container->get('context') instanceof Metadata) {
```
bertramakers, line 41: Is there any code that still writes to `$app['context']` (or now `$container->addShared('context')`? I cannot find any. So I think this `if()` and the `$container->addShared('context', '');` can just be removed.

bertramakers: + If there was still a place that overwrote it using `$app['context'] = ...` that would not work as intended anymore at this point because it would write the `context` service into the Silex Application object, but the `context` in the PSR container would remain an empty string.

bertramakers (changes requested):
Most looks good but I think the `context` service can be removed OR needs a solution to make it work like before if still used (see other comment)

PR #1186, opened 2022-10-06
III-4717 Handle (some) `Culturefeed_Exception` throwables as 404

bertramakers:
### Changed

- Some `CultureFeed_Exception` throwables are now converted to 404 ApiProblems as described in the ticket
- Creation of `ApiProblem` from a `Throwable` was moved to a separate factory class
- The `ErrorLogger` now always converts a `Throwable` that it is asked to log to check its status code to determine if it should log it or not (as we don't want to log any Throwable that results in a 4XX error, as those are not bugs). This was the easiest way to make sure that the `CultureFeed_Exception` throwables that are now converted to 404 responses would not be logged in Sentry.

---
Ticket: https://jira.uitdatabank.be/browse/III-4717

Note: I will open another PR with some acceptance tests to check these 404 errors on the UiTPAS endpoints. While the UiTPAS integration is hard to test locally, 404 errors should be easy to test.


src/Http/ApiProblem/ApiProblemFactory.php
```diff
@@ -77,23 +79,22 @@ public static function createFromThrowable(
             case $e instanceof DocumentDoesNotExist:
             case $e instanceof EntityNotFoundException:
             case $e instanceof MediaObjectNotFoundException:
-                $routeParameters = new RouteParameters($request);
-                if ($routeParameters->hasEventId()) {
+                if ($routeParameters && $routeParameters->hasEventId()) {
```
steftrenson, line 82: I think by extracting this case statement into a separate method, like you did with the cultureFeedException, you can just check once if you have route parameters and make the code even more readable.

steftrenson (approved):

src/Http/ApiProblem/ApiProblemFactory.php
LucWollants: Or use an early return with `return ApiProblem::urlNotFound();` when `$routeParameters` are `null`?

LucWollants (approved):

JonasVHG (approved):

src/Http/ApiProblem/ApiProblemFactory.php
bertramakers: I added the early return in [a85df3c](https://github.com/cultuurnet/udb3-silex/pull/1186/commits/a85df3ce937620bf63d829f16536cfcff0728818). If I'd extract it to a separate method I'd probably need an early return as well to avoid the duplicate checks. So I didn't extract it for now but we can definitely do that when this logic becomes too big.

PR #1188, opened 2022-10-06
III-4723 Return 502 for CultureFeed timeouts

bertramakers:
### Changed

- When `culturefeed-php` throws an `\Exception` with a specific curl timeout error message, we treat it as a 502 Bad Gateway since it indicates that either UiTPAS or UiTID v1 is down. Otherwise we treat it the same as before

---
Ticket: https://jira.uitdatabank.be/browse/III-4723


steftrenson (approved):

src/Http/ApiProblem/ApiProblemFactory.php
```diff
@@ -116,11 +117,47 @@ public static function createFromThrowable(
...
+        // Note: When we have improved the validation of most endpoints, we should make this return a plain
+        // "Internal Server Error" instead.
+        return ApiProblem::blank($e->getMessage(), $e->getCode() ?: 500);
+    }
+
+    /**
+     * Usually we will convert \Exception to a generic ApiProblem. However in some cases we want to convert generic
+     * exceptions thrown by external libraries to a more specific ApiProblem.
+     */
+    private static function convertGenericExceptionToApiProblem(
```
LucWollants, line 141: Should the name of the method be  `convertExceptionToGenericApiProblem` to closer match `convertToGenericApiProblem` ?

JonasVHG (approved):

src/Http/ApiProblem/ApiProblemFactory.php
bertramakers: No because the "generic" is different in those two cases. In this case you convert a generic exception (`\Exception`) to an `ApiProblem`, which could be a specific one like the Bad Gateway or just a blank one for exceptions it doesn't know how to handle.

In `convertToGenericApiProblem` you convert the throwable to the blank ApiProblem for exceptions it doesn't kow how to handle.

But since I have to explain it the wording is not great so I'll see if I can improve it in some way :)

bertramakers: I tried to improve it, let me know if this is clearer. Only thing I don't know how to improve/clarify is "generic exception" (to indicate `\Exception` specifically). Other things like "base exception", "literal exception", ... also feel vague.

LucWollants: Perfect now, using `default` did the trick.

PR #1197, opened 2022-10-07
III-5005 Improve file upload errors

bertramakers:
### Fixed

- If there is one uploaded file but its form data name is not `file`, we no longer get a type error but return an appropriate ApiProblem
- If the size of the uploaded file is `0` we return an ApiProblem instead of throwing a RuntimeException. If the size is `null` we still throw a `RuntimeException` because it could be a server issue (we'll need to see if it happens again or not after this is deployed)

---
Ticket: https://jira.uitdatabank.be/browse/III-5005

### DIFF
--- a/src/Media/ImageUploaderService.php
+++ b/src/Media/ImageUploaderService.php
@@ -107,7 +107,7 @@ class ImageUploaderService implements ImageUploaderInterface
         $fileSize = $file->getSize();
 
         if ($this->maxFileSize && !$fileSize) {
-            throw new \RuntimeException('The size of the uploaded image could not be determined.');
+            throw new InvalidFileSize('The size of the uploaded image could not be determined.');
         }
 
         if ($this->maxFileSize && $fileSize > $this->maxFileSize) {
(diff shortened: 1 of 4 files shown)


JonasVHG (approved):
Looks great!

tests/Http/Media/UploadMediaRequestHandlerTest.php
```diff
@@ -142,6 +142,29 @@ public function it_throws_if_description_is_missing(): void
         $this->uploadMediaRequestHandler->handle($request);
     }
 
+    /**
+     * @test
+     * @bugfix https://jira.uitdatabank.be/browse/III-5005
+     */
+    public function it_throws_if_no_a_file_was_uploaded_with_the_wrong_form_data_name(): void
```
LucWollants, line 149: I think there is a small typo here and should be `it_throws_if_a_file_was_uploaded_with_the_wrong_form_data_name`

src/Media/ImageUploaderService.php
```diff
@@ -47,7 +48,7 @@ public function __construct(
         int $maxFileSize = null
     ) {
         if ($maxFileSize < 0) {
-            throw new \RuntimeException('Max file size should be 0 or bigger inside config.yml.');
+            throw new RuntimeException('Max file size should be 0 or bigger inside config.yml.');
```
LucWollants, line 51: The config filename was changed 😉

PR #1198, opened 2022-10-07
III-5010 Improve file field errors

bertramakers:
### Fixed

- When the `copyrightHolder` field is invalid according to the `CopyrightHolder` VO a `body/invalid-data` ApiProblem is returned
- When the `language` field is invalid according to the `Language` VO a `body/invalid-data` ApiProblem is returned
- When a field is missing a `body/invalid-data` ApiProblem is returned

---
Ticket: https://jira.uitdatabank.be/browse/III-5010

### DIFF
--- a/src/Media/ImageUploaderService.php
+++ b/src/Media/ImageUploaderService.php
@@ -107,7 +107,7 @@ class ImageUploaderService implements ImageUploaderInterface
         $fileSize = $file->getSize();
 
         if ($this->maxFileSize && !$fileSize) {
-            throw new \RuntimeException('The size of the uploaded image could not be determined.');
+            throw new InvalidFileSize('The size of the uploaded image could not be determined.');
         }
 
         if ($this->maxFileSize && $fileSize > $this->maxFileSize) {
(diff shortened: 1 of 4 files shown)


src/Media/ImageUploaderService.php
```diff
@@ -47,7 +48,7 @@ public function __construct(
         int $maxFileSize = null
     ) {
         if ($maxFileSize < 0) {
-            throw new \RuntimeException('Max file size should be 0 or bigger inside config.yml.');
+            throw new RuntimeException('Max file size should be 0 or bigger inside config.yml.');
```
JonasVHG, line 51:
```suggestion
            throw new RuntimeException('Max file size should be 0 or bigger inside config.php.');
```

JonasVHG: Best to use `config.php` for clarity

JonasVHG (changes requested):

src/Media/ImageUploaderService.php
bertramakers: I've updated it so it's agnostic of the config file name, since anything in `src` should not be concerned with how the app is bootstrapped.

JonasVHG (approved):

PR #1204, opened 2022-10-10
III-4986 PSR-11 UiTPAS

JonasVHG:
### Changed

- `UiTPASIncomingEventServicesProvider` is now shared via the Container instead of the Silex App.
- `UiTPASIncomingEventServicesProvider` was moved outside the Silex folder.
- merge `api_name` in `config` array instead of service container.

---
Ticket: https://jira.uitdatabank.be/browse/III-4986


app/Silex/UiTPAS/UiTPASIncomingEventServicesProvider.php
```diff
@@ -5,30 +5,39 @@
...
+    {
+        $container = $this->getContainer();
+
+        $container->addShared(
+            'amqp.uitpas_event_bus_forwarding_consumer',
+            function () use ($container) {
                 // If this service gets instantiated, it's because we're running the AMQP listener for UiTPAS messages
                 // so we should set the API name to UiTPAS listener.
-                $app['api_name'] = ApiName::UITPAS_LISTENER;
+                $container->addShared('api_name', ApiName::UITPAS_LISTENER);
```
steftrenson, line 40: Q: I wonder if this will overwrite a previously set api_name. With the previous implementation, it will overwrite it because of the array access. A quick glance at the current implementation of the container makes me think this will just add a new definition to the container, without erasing the previous one.

steftrenson: Screenshot 2022-10-10 at 18 55 45
Screenshot 2022-10-10 at 18 55 31

Tested it out, it keeps the original value, so this change will likely break something.

bertramakers: Some possible solutions I could think of off the top of my head (better ones may also be possible):

- Using global variables as a temporary fix
- Using environment variables (using `putenv()` so we can set them in the code)
- Adding these kinds of variables to the `config` array? (Not sure if possible, maybe by reference?)

bertramakers: \+ Even if the `addShared()` could overwrite another service in the League container, it would not work in this case because the `api_name` is set on `$app` here, which means it will exist in two service containers (in this PR) with a different value in each: https://github.com/cultuurnet/udb3-silex/blob/7eeb838d1a925c1bd72a585764edcd8091fc4af3/bootstrap.php#L98

LucWollants (approved):

PR #1209, opened 2022-10-11
III-4990 Bootstrap dependencies

steftrenson:
Best reviewed commit-by-commit.

### Changed
- All directly shared dependencies in `bootstrap.php` are now shared on the PSR container through separate service providers.

### Removed
- Silex framework

---
Ticket: https://jira.uitdatabank.be/browse/III-4990


bootstrap.php
```diff
@@ -428,12 +428,6 @@ function ($app) {
     }
 );
 
-$app['role_permissions_cache'] = $app->share(
-    function ($app) {
-        return $app['cache']('role_permissions');
-    }
-);
```
bertramakers, line 587: Nice, thanks for checking that it's actually (un)used and removing it!  🙌

bootstrap.php
```diff
@@ -179,20 +179,16 @@ function ($cdbid) use ($app) {
...
-                $app['config']['cache']['redis'],
-                [
-                    'prefix' => $cacheType . '_',
-                ]
-            ));
-        };
-    }
+$container->addShared(
+    'cache',
+    fn ($cacheType) => new Doctrine\Common\Cache\PredisCache(
```
bertramakers: You will need to wrap this in `new League\Container\Argument\Literal\CallableArgument()`, otherwise the container will try to call this function to instantiate the `cache` service instead of just returning the callable as-is. See https://container.thephpleague.com/4.x/argument-types/

bertramakers (commented):
Nice work, this was a breeze to review so far with the super small incremental steps and very clear commit messages. I only have some minor remarks and found one small issue.

bootstrap.php
steftrenson: Good catch, I hope this is the first one I encountered.

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -246,6 +253,36 @@ function () use ($container) {
...
+                        $redis = new Redis();
+                        $redis->connect(
+                            $redisConfig['host'],
+                            $redisConfig['port']
+                        );
+                    } else {
+                        $redis = new Client(
+                            [
+                                'host' => $redisConfig['host'],
+                                'port' => $redisConfig['port']
```
JonasVHG, line 275: Still a small linting error because of no trailing comma 😉

bootstrap.php
```diff
@@ -154,539 +101,35 @@ function (AggregateType $aggregateType) use ($app) {
...
-            );
-            $redis->connect();
-        }
-        $socketIOHandler = new SocketIOEmitterHandler(new Emitter($redis), Logger::INFO);
-
-        return LoggerFactory::create($app->getLeagueContainer(), LoggerName::forResqueWorker($queueName), [$socketIOHandler]);
-    }
-);
-
 /** Production */
```
JonasVHG, line 118: This comment is no longer relevant and might best be removed.

JonasVHG (changes requested):
Looks very clean & good. Great to no longer have a massive bootstrap file.
I've only added some minor remarks.

app/Clock/ClockServiceProvider.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public function register(): void
+    {
+        $container = $this->getContainer();
+
+        $container->addShared(
+            'timezone',
+            fn () => new DateTimeZone(
+                empty($container->get('config')['timezone']) ? 'Europe/Brussels' : $container->get('config')['timezone']
```
bertramakers, line 27: At the top of `bootstrap.php` we also set the default timezone to a hardcoded value:
https://github.com/cultuurnet/udb3-silex/blob/621378c54ba6af1aed2a9593ef747c10e4ced630/bootstrap.php#L75

Maybe we should also use this `timezone` container value for that? (We'd need to move the `date_default_timezone_set` down a little bit but that should be okay)

Of course not very urgent/important because no one else is running this app in another timezone anyway at the moment.

bertramakers (approved):
Everything looks super good! Thanks for the very clear steps to follow, it was a breeze to review! Only the linting error to fix and the small remark from @JonasVHG about the obsolete comment, but otherwise this looks good to merge!

bootstrap.php
steftrenson: I re-ordered the whole file so it makes a bit more sense, this comment was indeed not necessary 👍

JonasVHG (approved):

PR #1213, opened 2022-10-11
III-5020 Internalize SwiftmailerServiceProvider

bertramakers:
### Changed

- The `mailer` service definition (a `Swift_Mailer` class instance) is now registered using a custom service provider on the new League container instead of a ServiceProvider provided by `silex/silex`

---
Ticket: https://jira.uitdatabank.be/browse/III-5020

Since this is hard to test with an acceptance test or unit test, I tested it manually by reloading the supervisor commands (so the `udb3-event-export-worker` was restarted and loaded the new code), and then doing an export of some events via the UI. I then verified that there was a new email in the local mailcatcher:

Screen Shot 2022-10-11 at 18 50 27

### DIFF
--- a/app/SwiftMailer/SwiftMailerServiceProvider.php
+++ b/app/SwiftMailer/SwiftMailerServiceProvider.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\SwiftMailer;
+
+use CultuurNet\UDB3\Container\AbstractServiceProvider;
+use Swift_Events_SimpleEventDispatcher;
+use Swift_Mailer;
+use Swift_StreamFilters_StringReplacementFilterFactory;
+use Swift_Transport_Esmtp_Auth_CramMd5Authenticator;
+use Swift_Transport_Esmtp_Auth_LoginAuthenticator;
+use Swift_Transport_Esmtp_Auth_PlainAuthenticator;
+use Swift_Transport_Esmtp_AuthHandler;
+use Swift_Transport_EsmtpTransport;
+use Swift_Transport_StreamBuffer;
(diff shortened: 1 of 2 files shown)


app/SwiftMailer/SwiftMailerServiceProvider.php
```diff
@@ -0,0 +1,74 @@
...
+                        'auth_mode' => null,
+                    ],
+                    $container->get('config')['swiftmailer.options']
+                );
+
+                $transport->setHost($options['host']);
+                $transport->setPort($options['port']);
+                $transport->setEncryption($options['encryption']);
+
+                // The following methods are "magic" methods which PHPStan does not recognize on this class for some
```
LucWollants, line 64: I think PHPStan is correct because they are only part of the docblock from `Swift_SmtpTransport` and not from `Swift_Transport_EsmtpTransport`. But it is perfect to make PHPStan ignore these.

LucWollants (approved):

steftrenson (approved):

app/SwiftMailer/SwiftMailerServiceProvider.php
bertramakers: Aha you're right it's a mistake in the docblock(s) of the `swiftmailer/swiftmailer` package. The `Swift_Transport_EsmtpTransport` class defines the magic `_call()` method and supports these methods, but instead the `Swift_SmtpTransport` class that extends it has the docblock (and we don't use that class). But yeah there is no real solution on our end except making PHPStan ignore the errors I think. Or we need to extend `Swift_Transport_EsmtpTransport` with a custom class and define the docblock with magic methods on that one, but that feels very overkill 😄

PR #1218, opened 2022-10-12
III-5031 Enable PHP v8.x tests

bertramakers:
### Fixed

---


bertramakers:
It looks like this will require a lot of package updates, and we'll need to update our own `cultuurnet/calendar-summary-v3` as well to add compatibility with 8.x in its composer.json (and check that it works of course). So I'm going to close this for now until we actually pick up PHP 8 as a project. Don't worry it's planned in the near future :) But I was hoping to already get an idea of what will need to happen and now we have some more info. We can still merge #1216 and #1217 of course

PR #1224, opened 2022-10-19
Add start cluster id option to the process cluster command

LucWollants:
### Added
- Added start cluster id option to the process cluster command

### DIFF
--- a/app/Console/Command/ProcessDuplicatePlaces.php
+++ b/app/Console/Command/ProcessDuplicatePlaces.php
@@ -67,7 +67,7 @@ final class ProcessDuplicatePlaces extends AbstractCommand
             'Execute a dry-run of the process-duplicates script.'
         );
         $this->addArgument(
-            'starting_cluster_id',
+            'start-cluster-id',
             InputArgument::OPTIONAL,
             'Enter starting cluster id?'
         );
@@ -76,7 +76,7 @@ final class ProcessDuplicatePlaces extends AbstractCommand
     public function execute(InputInterface $input, OutputInterface $output): int
     {
         $dryRun = (bool)$input->getOption('dry-run');
-        $startingClusterId = (int)$input->getArgument('starting_cluster_id');
+        $startingClusterId = (int)$input->getArgument('start-cluster-id');
 
         $clusterIds = $this->duplicatePlaceRepository->getClusterIds();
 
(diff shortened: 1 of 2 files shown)


app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -65,11 +66,17 @@ public function configure(): void
             InputOption::VALUE_NONE,
             'Execute a dry-run of the process-duplicates script.'
         );
+        $this->addArgument(
+            'starting_cluster_id',
+            InputArgument::OPTIONAL,
+            'Enter starting cluster id?'
```
bertramakers, line 72: This will not be prompted so I would not make it a question but a description. Example how it will be printed:

```
vagrant@udb3:/var/www/udb-silex$ ./bin/udb3.php place:process-duplicates --help
Description:
  Process duplicate places (determine canonical, update event locations and reindex)

Usage:
  place:process-duplicates [options] [--] [<starting_cluster_id>]

Arguments:
  starting_cluster_id   Enter starting cluster id? // <-- here

Options:
  -d, --dry-run         Execute a dry-run of the process-duplicates script.
  -h, --help            Display this help message
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi            Force ANSI output
      --no-ansi         Disable ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
```

Also it currently does not give any more information than the name of the argument, so you could also just omit it.

app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -65,11 +66,17 @@ public function configure(): void
             InputOption::VALUE_NONE,
             'Execute a dry-run of the process-duplicates script.'
         );
+        $this->addArgument(
+            'starting_cluster_id',
```
bertramakers, line 70: Looking at other usages of `addArgument`, we always give them a `kebab-case` or `camelCase` name. While that's not consistent, I'd suggest to pick on of those to avoid adding a third convention

LucWollants:
@bertramakers thanks for the review! I was a bit too hasty to use the code we created while fixing this command on production.

bertramakers (approved):

PR #1226, opened 2022-10-26
Cleanup offer command factory

steftrenson:
Found these classes when doing some research on a previous PR. They seem to be unused, but a double-check wouldn't hurt.

### Removed
- `OfferCommandFactoryInterface` and it's two instantiated classes.


bertramakers (approved):
Thanks for the cleanup! 👌

JonasVHG (approved):

PR #1228, opened 2022-10-27
III-4256 offer image post

steftrenson:
### Changed
- use JSON schema validation in `Offer\AddImageRequestHandler`.
- `AddImageRequestHandler` will throw an `ApiProblem` when the image could not be found.

### Fixed
- `AddImageRequestHandler` would crash when the `mediaObjectId` wasn't a UUID.

---
Ticket: https://jira.uitdatabank.be/browse/III-4256

### DIFF
--- a/src/Http/Offer/AddImageRequestHandler.php
+++ b/src/Http/Offer/AddImageRequestHandler.php
@@ -30,6 +30,8 @@ final class AddImageRequestHandler implements RequestHandlerInterface
     {
         $routeParameters = new RouteParameters($request);
         $offerId = $routeParameters->getOfferId();
+        $offerType = $routeParameters->getOfferType();
+
         $bodyContent = Json::decode($request->getBody()->getContents());
 
         if (empty($bodyContent->mediaObjectId)) {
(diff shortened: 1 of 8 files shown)


src/Http/Offer/AddImageRequestHandler.php
```diff
@@ -5,14 +5,16 @@
 namespace CultuurNet\UDB3\Http\Offer;
 
 use Broadway\CommandHandling\CommandBus;
-use CultuurNet\UDB3\Event\Commands\AddImage as EventAddImage;
-use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use Broadway\Repository\Repository;
+use CultuurNet\UDB3\Http\Organizer\AddMediaObjectPropertiesRequestBodyParser;
```
bertramakers, line 9: This is a class in the `Organizer` namespace, so it should not be used in the `Offer` namespace. The request bodies are also different, on offers we only accept `mediaObjectId` at the moment it seems, and on organizers we accept `id`, `language`, `description`, `copyrightHolder` (on offers these can only be set afterwards with an update it seems). It would definitely be good to align those better in the future, but for now I would keep these separated in the code as two separate classes for the different endpoints / request bodies

bertramakers (approved):
Looks perfect now! 👌

PR #1231, opened 2022-11-02
III-4258 update offer image

steftrenson:
### Changed
- A request to `UpdateImageRequestHandler` is now validated with JSON schema.
- `UpdateImage` command expects string iso `StringLiteral` for a description in its constructor. It returns a string now as well.

### Fixed
- An `ApiProblem` is thrown when mediaId is not a valid UUID iso throwing an exception.

---
Ticket: https://jira.uitdatabank.be/browse/III-4258

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
     "predis/predis": "~1.0",
     "psr/http-server-middleware": "^1.0",
     "psr/log": "^1.0",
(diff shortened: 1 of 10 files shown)


bertramakers (approved):
Looks good, but if I understand correctly based on our conversation on Slack this will also need to throw the new API problem if the image is not linked to the event/place right?

steftrenson:
@bertramakers I was going to implement that change in a separate PR see #1233

PR #1233, opened 2022-11-03
III-4260 Image must be linked to resource exception

steftrenson:
### Added
- `ThrowingTraceableCommandBus` which is a wrapper around `TraceableCommandBus` to be able to test throwing domain exceptions.

### Fixed
- `SelectMainImageRequestHandler` will now throw an `ApiProblem` when trying to set a main image that's not linked to the offer.
- `UpdateMainImageRequestHandler` will now throw an `ApiProblem` when trying to set a main image that's not linked to the organizer.

---
Ticket: https://jira.uitdatabank.be/browse/III-4260


tests/Http/Offer/SelectMainImageRequestHandlerTest.php
```diff
@@ -152,6 +149,35 @@ public function it_throws_when_image_does_not_exist(): void
...
     }
 
+    /**
+     * @test
+     */
+    public function it_throws_when_image_is_not_linked_to_offer(): void
+    {
+        $this->givenThereIsAnImage();
+
+        $this->commandBus->throwsOnDispatch(new ImageMustBeLinkedException());
```
bertramakers, line 159: Nit: As an alternative you could just create another instance of `SelectMainImageRequestHandler` in this test method instead of using `$this->selectMainImageRequestHandler`, and inject a mock `CommandBus` that throws an exception when `dispatch()` is called. Then you don't need to create a new `ThrowingTraceableCommandBus` class specifically for this one test case.

bertramakers (approved):
Looks good! Do we also still need this on organizers or was that already implemented?

steftrenson:
@bertramakers I was implementing it when you asked the question :)

bertramakers:
Alright! Also see my small suggestion :)

bertramakers (approved):

PR #1237, opened 2022-11-21
III-4269 offer typical age range

steftrenson:
### Changed
- `UpdateTypicalAgeRangeRequestHandler` now uses JSON schema validation.
- `AbstractUpdateTypicalAgeRange` improved typehinting

---
Ticket: https://jira.uitdatabank.be/browse/III-4269

### DIFF
--- a/tests/Http/Offer/UpdateTypicalAgeRangeRequestHandlerTest.php
+++ b/tests/Http/Offer/UpdateTypicalAgeRangeRequestHandlerTest.php
@@ -11,6 +11,7 @@ use CultuurNet\UDB3\Http\Response\AssertJsonResponseTrait;
 use CultuurNet\UDB3\Http\Response\NoContentResponse;
 use CultuurNet\UDB3\Offer\Commands\AbstractUpdateTypicalAgeRange;
 use CultuurNet\UDB3\Place\Commands\UpdateTypicalAgeRange as PlaceUpdateTypicalAgeRange;
+use Iterator;
 use PHPUnit\Framework\TestCase;
 
 final class UpdateTypicalAgeRangeRequestHandlerTest extends TestCase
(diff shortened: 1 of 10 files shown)


tests/Http/Offer/UpdateTypicalAgeRangeRequestHandlerTest.php
```diff
@@ -167,7 +167,9 @@ public function provideInvalidRequestBodies(): Iterator
             yield 'minimum age is bigger than maximum age ' . $offerType => [
                 'offerType' => $offerType,
                 'request' => '{ "typicalAgeRange": "12-6"}',
-                'expectedProblem' => ApiProblem::bodyInvalidData(),
+                'expectedProblem' => ApiProblem::bodyInvalidData(
+                    new SchemaError('/typicalAgeRange', '"from" age should not exceed "to" age')
```
steftrenson, line 171: @bertramakers Don't know if it makes sense to make this a documented exception like we did with the main image selection?

bertramakers: In this case I would go with `https://api.publiq.be/probs/body/invalid-data` because of the following reasons:

- Ranges in general need to start with a lower bound and end with a higher bound, that is not specific to our domain
- The data that causes the error is all included in the body (both the "from" & "to" ages), and editing the body is the only solution (= it's definitely invalid body data, there is no other fix)

Sometimes it's a grey zone though and only of those two reasons apply. In that case I tend to give the first reason (is it specific to us or also the case in general software) more weight when trying to decide about it.

JonasVHG (approved):

tests/Http/Offer/UpdateTypicalAgeRangeRequestHandlerTest.php
steftrenson: Makes sense, thanks for the explanation.

LucWollants (approved):
Looks good. I did a scan through the code but did not run any acceptance tests.

bertramakers (approved):

PR #1238, opened 2022-11-21
III-4272 Endpoint to update `workflowStatus`

bertramakers:
### Added

- Added a new route & request handler to update the `workflowStatus` of an event or place, to replace the quirky PATCH endpoint that uses custom content-types to determine the intended state transition.

---
Ticket: https://jira.uitdatabank.be/browse/III-4272

I will add some acceptance tests as well, but in the meantime this can already be reviewed if someone has the time (as well as the schema in https://github.com/cultuurnet/apidocs/pull/115)
Edit: Done, see https://github.com/cultuurnet/udb3-acceptance-tests/pull/169

I'll also fix the merge conflict in `composer.lock` once the schema is merged, because I'll have to update the `publiq/udb3-json-schemas` dependency then again anyway
Edit: Done

### DIFF
--- a/tests/Http/Offer/UpdateWorkflowStatusRequestHandlerTest.php
+++ b/tests/Http/Offer/UpdateWorkflowStatusRequestHandlerTest.php
@@ -60,8 +60,7 @@ final class UpdateWorkflowStatusRequestHandlerTest extends TestCase
     public function it_dispatches_a_publish_command_for_workflow_status_ready_for_validation(
         string $offerType,
         string $publishCommandClassName
-    ): void
-    {
+    ): void {
         $request = (new Psr7RequestBuilder())
             ->withRouteParameter('offerType', $offerType)
             ->withRouteParameter('offerId', 'd1422721-f226-48fd-a26d-cb21599ee533')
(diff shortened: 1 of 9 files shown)


tests/Http/Offer/UpdateWorkflowStatusRequestHandlerTest.php
```diff
@@ -0,0 +1,255 @@
...
+        $this->updateWorkflowStatusRequestHandler = new UpdateWorkflowStatusRequestHandler($this->commandBus);
+    }
+
+    /**
+     * @test
+     */
+    public function it_dispatches_no_command_for_workflow_status_draft(): void
+    {
+        $request = (new Psr7RequestBuilder())
+            ->withRouteParameter('offerType', 'events')
```
LucWollants, line 41: Would it make sense to also test the `place` endpoint to make sure that the if-statement to create the correct command works properly?

LucWollants (approved):

bertramakers:
Acceptance tests can now be found at https://github.com/cultuurnet/udb3-acceptance-tests/pull/169

tests/Http/Offer/UpdateWorkflowStatusRequestHandlerTest.php
bertramakers: It kind of hurts the readability of that data provider because I need to prefix the keys of the `yield` statement so they remain unique then. But I've implemented it as well as data providers to run the other tests for both events & places in https://github.com/cultuurnet/udb3-backend/pull/1238/commits/7d9c2f598909010790ac1ac8f355e077d37b1107

LucWollants (approved):

PR #1247, opened 2023-01-02
III-5159 Rounding Error € 2.30

JonasVHG:
### Changed
- use `bcmath` for more precision when working with float prices

### Fixed
- € 2.3 is no longer changed to € 2.29

---
Ticket: https://jira.uitdatabank.be/browse/III-5159


composer.json
```diff
@@ -11,6 +11,7 @@
   "minimum-stability": "dev",
   "require": {
     "php": ">=7.4",
+    "ext-bcmath": "*",
```
JonasVHG, line 14: I've checked with `php -m` and it should already be installed on the server.

JonasVHG:
This will be fixed in a different PR: https://github.com/cultuurnet/udb3-backend/pull/1248

PR #1248, opened 2023-01-02
III-5159 Rounding Error eg. € 2.30

JonasVHG:
### Added
- `MoneyFactory` to better handle the conversions from `int`, `float` & `string` to `Money` 

### Fixed
- € 2.3 is no longer changed to € 2.29
- Temp Fix for `setTestNow()`

---
Ticket: https://jira.uitdatabank.be/browse/III-5159

### DIFF
--- a/src/MoneyFactory.php
+++ b/src/MoneyFactory.php
@@ -16,8 +16,11 @@ final class MoneyFactory
         return new Money((int) round(($price*100)), $currency);
     }
 
-    public static function createFromStringInCents(
-        string $price,
+    /**
+     * @param string|int $price
+     */
+    public static function createFromCentsValue(
+        $price,
         Currency $currency
     ): Money {
         return new Money((int) $price, $currency);
--- a/src/MoneyFactory.php
+++ b/src/MoneyFactory.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3;
+
+use Money\Currency;
+use Money\Money;
+
+final class MoneyFactory
+{
+    public static function createFromFloat(
+        float $price,
+        Currency $currency
+    ): Money {
+        return new Money((int) round(($price*100)), $currency);
+    }
+}
--- a/src/MoneyFactory.php
+++ b/src/MoneyFactory.php
@@ -9,8 +9,11 @@ use Money\Money;
 
 final class MoneyFactory
 {
-    public static function createFromFloat(
-        float $price,
+    /**
+     * @param string|int|float $price
+     */
+    public static function create(
+        $price,
         Currency $currency
     ): Money {
         return new Money((int) round(($price*100)), $currency);
--- a/src/MoneyFactory.php
+++ b/src/MoneyFactory.php
@@ -15,4 +15,11 @@ final class MoneyFactory
     ): Money {
         return new Money((int) round(($price*100)), $currency);
     }
+
+    public static function createFromStringInCents(
+        string $price,
+        Currency $currency
+    ): Money {
+        return new Money((int) $price, $currency);
+    }
 }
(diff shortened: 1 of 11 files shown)


src/MoneyFactory.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * @param string|int|float $price
+     */
+    public static function create(
+        $price,
+        Currency $currency
+    ): Money {
+        self::guard($price);
+        return new Money((int) round(($price*100)), $currency);
```
bertramakers, line 20: This `round()` introduces an interesting question. If an integrator sends a price like `2.99` it will work correctly. But if they send a price like `2.9999999` it will get rounded to `3.00`. A price like `2.9999999` is weird of course, but is rounding to `3.00` in that case the intended behavior? 🤔

JonasVHG: As far as I looked, that would require string manipulation, because you cannot `floor()` to 2 decimals. So we get it as a `float` or `integer` from the json, convert it to `string` and then back again to `float` which could cause new rounding errors. While preventing bad user input is a good thing, it can never be foolproof.

LucWollants: In euros something smaller than a cent does not exist, so I think it's fine to round this up.

LucWollants: I also did a quick test with `number_format`, `sprintf`, and `intval` but they all suffer from the rounding error @bertramakers pointed out.

bertramakers: I tested the current behavior on acc by sending a `PUT /events/{eventId}/price-info` request with this body:
```
[{"category":"base","name":{"nl":"Basistarief","fr":"Tarif de base","en":"Base tariff","de":"Basisrate"},"price":10,"priceCurrency":"EUR"},{"category":"tariff","name":{"nl":"Jongeren","en":"Youth"},"price":12.99999,"priceCurrency":"EUR"}]
```

Result: https://io-acc.uitdatabank.be/events/5330a84f-7496-46a1-b60d-fa7d62ec5fb8
```
"price": 12.99
```

So in the current implementation it just drops everything after the 2nd decimal.

I'm just pointing it out that it's different. If there is no other solution this bugfix is more important than keeping the same logic for such an edge case.

bertramakers: Playing devil's advocate for a second, you could argue that the current behavior (of dropping everything after the 2nd decimal) is also a bug and that it should in fact round it.

LucWollants: >  If there is no other solution this bugfix is more important than keeping the same logic for such an edge case.

I think we could handle it ourselves with string manipulation and use `substr` to convert to a precision of 2 digits.

LucWollants: Another option is maybe to prevent using a value like `2.999` by updating the JSON schema rules on the common price info?

bertramakers: Sadly JSON schema has no way to define the expected/allowed number of decimals for a number. You can only differentiate between integer and number (float). See https://json-schema.org/understanding-json-schema/reference/numeric.html

JonasVHG: That would be the cleanest solution, though if somebody is already posting prices like `2.999` it might be considered a breaking change. Another issue is if such a thing is possible, I've found ` "multipleOf": 0.01` might/might not work.

bertramakers: Ah that's clever with the `multipleOf`, that could work in this case. For example https://www.jsonschemavalidator.net/s/63REPppO (you can play around with the price value to test it)

LucWollants: Then I guess we can keep this PR as is and create a follow-up ticket to only allow a precision of 2 digits.
I would not worry about it being a breaking change.

bertramakers (approved):

src/MoneyFactory.php
JonasVHG: III-5064

bertramakers: @JonasVHG that looks like a typo? It's an unrelated issue that's already closed

PR #1253, opened 2023-02-06
III-5214 Add Contributor Logic

JonasVHG:
### Added
- Migration for `contributor_relations`
- new ServiceProvider for `ContributorRepository`
- `ContributorRepository` for `CRUD`-operations on table `contributor_relations`
- `RequestHandlers` for `UpdateContributors` on `Offers` & `Organizers`
- "Non-event sourced" Commands `UpdateContributors` for `Offers` & `Organizers`
- `CommandHandlers` for `UpdateContributors` on `Offers` & `Organizers`
- Registered route `contributor` on `Offer` & `Organizer`
- `ContributorDenormalizer` to parse Contributor e-mails

### Changed
- updated `publiq/udb3-json-schemas` in `composer`
- Added `*_CONTRIBUTORS_PUT` locations to `JsonSchemaLocator`

### TODO in following PR
- `GET` method so front-end can use it
- Find out how to map `email` to `userId`
- Allow editing of items based on the `userId`

---


src/Contributor/ContributorRepositoryInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Contributor;
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddress;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddresses;
+
+interface ContributorRepositoryInterface
```
bertramakers, line 11: We used this convention of suffixing interface with `Interface` in the past, but have since switched to omitting the suffix and giving the implementations more concrete names that indicate how they are implemented. So in this case the interface should be named `ContributorRepository`, and the implementation for example `DbalContributorRepository`.

Here's a great article about it: https://verraes.net/2013/09/sensible-interfaces/

app/Offer/OfferServiceProvider.php
```diff
@@ -562,6 +566,14 @@ public function register(): void
             fn () => new UpdateWorkflowStatusRequestHandler($container->get('event_command_bus'))
         );
 
+        $container->addShared(
+            ManageContributorsRequestHandler::class,
+            fn () => new ManageContributorsRequestHandler(
+                $container->get(OfferRepository::class),
+                $container->get(ContributorRepository::class)
```
bertramakers, line 573: If you use the `ContributorRepository` interface name instead in the service container (with suffix at the moment, or without if you remove it like I suggested in the other comment) the service container should be able to auto-wire `ManageContributorsRequestHandler` because it knows both the required `OfferRepository` and the required `ContributorRepository(Interface)`. So you should be able to leave out this explicit `$container->addShared(ManageContributorsRequestHandler::classaddShared, ...)` then.

src/Contributor/ContributorRelationsConfigurator.php
```diff
@@ -0,0 +1,22 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Contributor;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Type;
+
+final class ContributorRelationsConfigurator
```
bertramakers, line 11: Nit: This seems to be only used in unit tests, so maybe move it to the `tests` folder

src/Http/Offer/ManageContributorsRequestHandler.php
```diff
@@ -0,0 +1,78 @@
...
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $offerType = $routeParameters->getOfferType();
+        $offerId = $routeParameters->getOfferId();
+
+        try {
+            $this->offerRepository->load($offerId);
+        } catch (AggregateNotFoundException $exception) {
+            throw ApiProblem::offerNotFound($offerType, $offerId);
```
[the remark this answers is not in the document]
bertramakers: (See also https://github.com/cultuurnet/udb3-backend/blob/master/src/Http/ApiProblem/ApiProblemFactory.php#L73-L104)

src/Http/Offer/ManageContributorsRequestHandler.php
```diff
@@ -0,0 +1,78 @@
...
+
+        /** @var EmailAddresses $emails */
+        $emails = $parser->parse($request)->getParsedBody();
+
+        $this->contributorRepository->deleteContributors(new UUID($offerId));
+
+        $emailsAsArray = $emails->toArray();
+        /** @var EmailAddress $email */
+        foreach ($emailsAsArray as $email) {
+            $this->contributorRepository->addContributor(
```
bertramakers, line 71: Just a minor suggestion, but maybe it could be useful to move this "delete all, then add every email address" to the repository itself as a kind of "overwriteContributors(UUID $id, EmailAddresses $contributors)" method?

src/Http/Offer/ManageContributorsRequestHandler.php
```diff
@@ -0,0 +1,78 @@
...
+use CultuurNet\UDB3\Http\Response\NoContentResponse;
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddress;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddresses;
+use CultuurNet\UDB3\Offer\OfferRepository;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+final class ManageContributorsRequestHandler implements RequestHandlerInterface
```
bertramakers, line 25:
```suggestion
final class UpdateContributorsRequestHandler implements RequestHandlerInterface
```

To be consistent with other request handlers for requests with PUT methods.

src/Role/ValueObjects/Permission.php
```diff
@@ -21,6 +21,7 @@ public static function getAllowedValues(): array
             'Voorzieningen bewerken',
             'Producties aanmaken',
             'Films aanmaken',
+            'Bijdragers beheren',
```
bertramakers, line 24: I was the one who suggested this extra permission, but we could also use the existing "Aanbod beheren" / "Organisaties beheren" permissions for this 🤔 The difference would of course be that with a new permission we could have more fine-grained roles, but with the latter we have less configuration work to enable this permission on existing roles. Ideally we should check this with Corneel. Maybe use the existing one for now, and add this one later in a follow-up PR if actually needed?

src/Contributor/ContributorRepository.php
```diff
@@ -55,24 +55,27 @@ public function isContributor(UUID $id, EmailAddress $emailAddress): bool
...
-    }
-
-    public function deleteContributors(UUID $id): void
-    {
-        $this->connection
-            ->delete(
-                self::TABLE,
-                ['uuid' => $id->toString()]
-            );
+        $this->connection->transactional(
```
bertramakers: Oh good idea, the transaction is a nice added benefit of doing it in the repository. I hadn't even considered that 👍

src/Offer/CommandHandlers/UpdateContributorsHandler.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->offerRepository = $offerRepository;
+        $this->contributorRepository = $contributorRepository;
+    }
+    public function handle($command): void
+    {
+        if (!($command instanceof UpdateContributors)) {
+            return;
+        }
+
+        $this->offerRepository->load($command->getItemId());
```
bertramakers, line 30: Nit: Maybe add a comment here that you load the offer to check that it actually exists

src/Organizer/CommandHandler/UpdateContributorsHandler.php
```diff
@@ -0,0 +1,37 @@
...
+    private OrganizerRepository $organizerRepository;
+
+    private ContributorRepository $contributorRepository;
+
+    public function __construct(OrganizerRepository $organizerRepository, ContributorRepository $contributorRepository)
+    {
+        $this->organizerRepository = $organizerRepository;
+        $this->contributorRepository = $contributorRepository;
+    }
+    public function handle($command): void
```
bertramakers, line 24: Nit: Usually we put an empty line before each method, but it looks like the linter does not check that

bertramakers (approved):
Nice work 👌

src/Organizer/CommandHandler/UpdateContributorsHandler.php
JonasVHG: I have too much faith in `composer cs-fix` 🫣, I'll fix it.

bertramakers: That's okay, that's what is there for! You shouldn't have to check every file manually yourself.

app/Offer/OfferServiceProvider.php
```diff
@@ -352,6 +357,14 @@ public function register(): void
             )
         );
 
+        $container->addShared(
+            UpdateContributorsHandler::class,
```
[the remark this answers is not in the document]
JonasVHG: You're correct, I've made the changes.

PR #1254, opened 2023-02-08
III-5214 GET endpoint contributor

JonasVHG:
### Added

- `GetContributorsRequestHandler` for `Offers` & `Organizers`
- registered `GET` endpoint for `contributors` on `Offers` & `Organizers`
- Tests for `GetContributorsRequestHandler`

### TODO in following PR
- Find out how to map `email` to `userId`
- Allow editing of items based on the `userId`

---


app/Offer/OfferServiceProvider.php
```diff
@@ -566,6 +568,14 @@ public function register(): void
             fn () => new UpdateWorkflowStatusRequestHandler($container->get('event_command_bus'))
         );
 
+        $container->addShared(
+            GetContributorsRequestHandler::class,
```
bertramakers, line 572: As in https://github.com/cultuurnet/udb3-backend/pull/1253/files#r1099961153, I think this class can be auto-wired if you define the contributor repository in the service container using its interface name instead of its implementation name.

bertramakers (changes requested):
Looks good so far, but like #1253 this is missing the (important) permission check so that these email addresses are not shared publicly - which is the reason that we didn't want to put them in the event/place/organizer JSON instead.

In this case you won't be able to work with a command to let the command bus do the permission check for you, since you're reading instead of writing and the command bus cannot return data.

I'll look into the permission system and send you a DM on Slack with some more info how you can implement it here.

app/Offer/OfferServiceProvider.php
JonasVHG: This is unfortunatly no longer relevant since I have to use non FQN definitions.

bertramakers: Indeed, now that you have to inject `offer_permission_voter` you cannot use auto-wiring.

bertramakers (approved):

PR #1255, opened 2023-02-09
III-5214 Give contributors edit rights

JonasVHG:
### Added
- `ContributorVoter` to  give Contributors `edit` rights.
- `UserEmailAddressRepository` to retrieve `email` from `userId`
- `InMemoryUserEmailAddressRepository` as implementation of `UserEmailAddressRepository`

### Changed
- Added `ContributorVoter` to `OfferSecurityServiceProvider` & `OrganizerSecurityServiceProvider`
- `getEmailAddress()` on `JsonWebToken`

---

### DIFF
--- a/src/Http/Auth/RequestAuthenticatorMiddleware.php
+++ b/src/Http/Auth/RequestAuthenticatorMiddleware.php
@@ -17,6 +17,7 @@ use CultuurNet\UDB3\Http\Auth\Jwt\JwtValidator;
 use CultuurNet\UDB3\Http\Auth\Jwt\JsonWebToken;
 use CultuurNet\UDB3\Role\ReadModel\Permissions\Doctrine\UserPermissionsReadRepository;
 use CultuurNet\UDB3\Role\ValueObjects\Permission;
+use CultuurNet\UDB3\Security\UserEmailAddressRepository;
 use CultuurNet\UDB3\User\CurrentUser;
 use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
@@ -43,6 +44,7 @@ final class RequestAuthenticatorMiddleware implements MiddlewareInterface
     private ApiKeyConsumerReadRepository $apiKeyConsumerReadRepository;
     private ApiKeyConsumerSpecification $apiKeyConsumerPermissionCheck;
     private UserPermissionsReadRepository $userPermissionReadRepository;
+    private UserEmailAddressRepository $userEmailAddressRepository;
 
     public function __construct(
         JwtValidator $uitIdV1JwtValidator,
@@ -50,7 +52,8 @@ final class RequestAuthenticatorMiddleware implements MiddlewareInterface
         ApiKeyAuthenticator $apiKeyAuthenticator,
         ApiKeyConsumerReadRepository $apiKeyConsumerReadRepository,
         ApiKeyConsumerSpecification $apiKeyConsumerPermissionCheck,
-        UserPermissionsReadRepository $userPermissionsReadRepository
+        UserPermissionsReadRepository $userPermissionsReadRepository,
+        UserEmailAddressRepository $userEmailAddressRepository
     ) {
         $this->uitIdV1JwtValidator = $uitIdV1JwtValidator;
         $this->uitIdV2JwtValidator = $uitIdV2JwtValidator;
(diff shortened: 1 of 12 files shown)


src/Security/InMemoryUserEmailAddressRepository.php
```diff
@@ -0,0 +1,27 @@
...
+
+    public static function addUserEmail(string $userId, EmailAddress $emailAddress)
+    {
+        self::$mappedUserIds[$userId] = $emailAddress;
+    }
+
+    public function getEmailForUserId(string $userId): ?EmailAddress
+    {
+        if (array_key_exists($userId, self::$mappedUserIds)) {
+            /** @var EmailAddress $emailAddress */
```
bertramakers, line 21: You can add the following docblock on the property instead and then you don't need to add the `@var` annotation here I think, and then you can just do `return self::$mappedUserIds[$userId];`

```
/**
 * @var EmailAddress[] $mappedUserIds
 */
private static array $mappedUserIds = [];
```

src/Security/Permission/ContributorVoter.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->contributorRepository = $contributorRepository;
+    }
+
+    public function isAllowed(Permission $permission, StringLiteral $itemId, StringLiteral $userId): bool
+    {
+        return $this->repository->getEmailForUserId($userId->toNative()) &&
+            in_array(
+                $this->repository->getEmailForUserId($userId->toNative()),
+                $this->getEmailAddressesForItem($itemId->toNative())->toArray()
+            );
```
bertramakers, line 34: * I'd suggest to put the return value of `$this->repository->getEmailForUserId($userId->toNative())` in a variable so you don't need to call it twice
* I'd also suggest to not fetch all the email addresses but add some kind of `hasEmailAddress` on the contributor repository that does a query to check that a given item id has the given email address

src/Http/Auth/RequestAuthenticatorMiddleware.php
```diff
@@ -141,6 +145,12 @@ private function authenticateToken(ServerRequestInterface $request): void
 
         $validator->verifySignature($this->token);
         $validator->validateClaims($this->token);
+        if ($this->token->getEmailAddress()) {
+            $this->userEmailAddressRepository::addUserEmail(
+                $this->token->getUserId(),
+                $this->token->getEmailAddress()
+            );
+        }
```
bertramakers, line 153: I'm wondering if you really need to set the email address on the repository here using a static method. I think you could just get the `JsonWebToken::class` "service" from the container in the `UserEmailAddressRepository` service definition (in `UserServiceProvider`) and inject it in the repository that way (using the constructor or another method like this one).

I wasn't sure before because the `JsonWebToken::class` gets its value from the `RequestAuthenticatorMiddleware::getToken()` method, which will return `null` until the middleware is called by the router with a `Request`: https://github.com/cultuurnet/udb3-backend/blob/4728e8a920e4aadc3cbdcb7aebcf10ba5839ee48/app/Authentication/AuthServiceProvider.php#L131

However, this request handler for example also gets `JsonWebToken` injected directly in its constructor upon instantiation: https://github.com/cultuurnet/udb3-backend/blob/bca8bb38ada8aa07806652fc93a77f9873e1119d/app/User/UserServiceProvider.php#L52

The reason why it works is that we register our routes with just the class name of the request handler, for example: https://github.com/cultuurnet/udb3-backend/blob/64770269f8545d4e37ce541e0ad3bcb6c1ef94be/app/Http/PsrRouterServiceProvider.php#L465

Later, after the router has called all the middlewares (and thus there is a `JsonWebToken` in most cases), the router uses our custom `InvokableRequestHandlerContainer` which looks up the given class name in the service container and returns it. It's only at that point that the request handler *and all its dependencies* are instantiated.

So that means that the command bus, the permission voters, and their dependency on your new `UserEmailAddressRepository` are only instantiated after there is already a `JsonWebToken` "service" in the container (if a token was used, remember that there are some routes that do not need a token so it can always be `null` after all)

src/Security/InMemoryUserEmailAddressRepository.php
```diff
@@ -4,25 +4,20 @@
...
+        $this->token = $token;
     }
 
     public function getEmailForUserId(string $userId): ?EmailAddress
     {
-        if (array_key_exists($userId, self::$mappedUserIds)) {
-            return self::$mappedUserIds[$userId];
-        }
-        return null;
+        return $this->token->getEmailAddress();
```
bertramakers, line 21: We should probably double-check here that the given `string $userId` is the same as the user id inside the token, to avoid bugs down the line when/if someone uses this class in another place

JonasVHG: Good find, I've made the changes.

bertramakers (approved):
Nice work! 👌

PR #1259, opened 2023-02-16
III-5334 add itemtype contributors

JonasVHG:
### Added
- Migration to add `type` to `contributor_repository`
- Added `ItemType` to `ContributorRepository`

### Changed

- Add `OfferType` to constructor of `Offer\UpdateContributors`
- HardCode `ItemType::organizer()`  to `Organizer\UpdateContributorsHandler`
- Check for commands in `TraceableCommandBus` in `UpdateContributorsRequestHandlers`

---


tests/Contributor/ContributorRepositoryTest.php
```diff
@@ -128,7 +134,8 @@ public function it_can_overwrite_contributor(): void
                     new EmailAddress('pol@gent.be'),
                     new EmailAddress('mieke@gent.be'),
                 ]
-            )
+            ),
+            ItemType::event()
```
LucWollants, line 138: I think it's never tested that the type is actually saved inside the database.

LucWollants (changes requested):
Only one small remark to check the saved item type

tests/Contributor/ContributorRepositoryTest.php
```diff
@@ -146,4 +147,50 @@ public function it_can_overwrite_contributor(): void
             $this->contributorRepository->getContributors($this->ghentEvent)
         );
     }
+
+    /**
+     * @test
+     * @dataProvider itemTypeDataProvider
+     */
+    public function it_saves_the_correct_itemType(ItemType $itemType): void
```
LucWollants, line 155: No need to do this in a separate unit test, you can do it inside the existing `it_can_overwrite_contributor`.
There is basically no difference between `it_can_overwrite_contributor` and `it_saves_the_correct_itemType`

JonasVHG: It thought it would be better to have 1 test that focuses on the updating of an item that has contributors, and a separate one that focuses on seeing that it correctly saves the type. But I can indeed put those 2 in 1 test.

LucWollants: @JonasVHG with this change you only test the item type of the first contributor, but all 3 rows should have item type event. Maybe you can use a query builder and create a query with the UUID and the item type. When doing that you only need to check the returned number of rows and that should be 3.

LucWollants (approved):

PR #1261, opened 2023-02-16
III-5334 messaging contributors

JonasVHG:
### Added

- `BroadcastingContributorRepository` as `Decorator` to publish messages
- Abstract `ContributorsUpdated` And childClasses `EventContributorsUpdated`, `PlaceContributorsUpdated` & `OrganizerContributorsUpdated` for Messages
- `ContributorsUpdatedFactory` to create message depending on `ItemType`

### Changed

- Use `BroadcastingContributorRepository` instead of `DbalContributorRepository` in `ServiceProvider`

---

### DIFF
--- a/src/Contributor/BroadcastingContributorRepository.php
+++ b/src/Contributor/BroadcastingContributorRepository.php
@@ -40,9 +40,9 @@ final class BroadcastingContributorRepository implements ContributorRepository
         return $this->repository->isContributor($id, $emailAddress);
     }
 
-    public function overwriteContributors(UUID $id, EmailAddresses $emailAddresses, ItemType $itemType): void
+    public function updateContributors(UUID $id, EmailAddresses $emailAddresses, ItemType $itemType): void
     {
-        $this->repository->overwriteContributors($id, $emailAddresses, $itemType);
+        $this->repository->updateContributors($id, $emailAddresses, $itemType);
 
         if ($itemType->sameAs(ItemType::event())) {
             $contributorsUpdated = $this->contributorOverwrittenFactory->createEventContributorsUpdated($id->toString());
--- a/src/Contributor/BroadcastingContributorRepository.php
+++ b/src/Contributor/BroadcastingContributorRepository.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Contributor;
+
+use Broadway\Domain\DomainEventStream;
+use Broadway\EventHandling\EventBus;
+use CultuurNet\UDB3\EventSourcing\DomainMessageBuilder;
+use CultuurNet\UDB3\Model\ValueObject\Identity\ItemType;
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddress;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddresses;
+
+final class BroadcastingContributorRepository implements ContributorRepository
+{
+    private ContributorRepository $repository;
+
+    private EventBus $eventBus;
+
+    private ContributorsUpdatedFactory $contributorOverwrittenFactory;
+
+    public function __construct(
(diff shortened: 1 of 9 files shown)


src/Contributor/BroadcastingContributorRepository.php
```diff
@@ -0,0 +1,57 @@
...
+    public function isContributor(UUID $id, EmailAddress $emailAddress): bool
+    {
+        return $this->repository->isContributor($id, $emailAddress);
+    }
+
+    public function updateContributors(UUID $id, EmailAddresses $emailAddresses, ItemType $itemType): void
+    {
+        $this->repository->updateContributors($id, $emailAddresses, $itemType);
+
+        if ($itemType->sameAs(ItemType::event())) {
```
LucWollants, line 47: Deciding which event needs to be created should be done inside the factory based on the `ItemType`. So the `ContributorsUpdatedFactory` should have one method `createForItemType(ItemType $itemType)` which contains early returns for every event type.

src/Contributor/ContributorsUpdatedFactory.php
```diff
@@ -0,0 +1,54 @@
...
+namespace CultuurNet\UDB3\Contributor;
+
+use CultuurNet\UDB3\Event\EventContributorsUpdated;
+use CultuurNet\UDB3\Iri\IriGeneratorInterface;
+use CultuurNet\UDB3\Organizer\OrganizerContributorsUpdated;
+use CultuurNet\UDB3\Place\PlaceContributorsUpdated;
+
+final class ContributorsUpdatedFactory
+{
+    private IriGeneratorInterface $eventContributorsUpdatedIriGenerator;
```
LucWollants, line 14: The name can be tweaked, it is not the IRI generator of the update endpoint but of the get endpoint

LucWollants (approved):

bertramakers (commented):
IMHO it would also be okay to have a single ContributorsUpdated message with just the IRI, that requires a lot less code and will also achieve the goal of making it possible for SAPI3 to index the contributors. If we ever need this event for something else where we cannot use the IRI but need the id/type we can easily add those as well (but still keep it 1 class). This is just a suggestion though, the approach with 3 events also works.

To actually get the messages published on AMQP you will need to map the class name(s) of the event(s) here to a content-type: https://github.com/cultuurnet/udb3-backend/blob/7edebd7770f89983a4579a690afcbc1432edddc6/app/AMQP/AMQPPublisherServiceProvider.php#L47-L51 (You can follow the same structure as the existing content-types to invent new ones). The event(s) also need(s) to implement Serializable, but that is already the case I see 👍

LucWollants:
> single ContributorsUpdated message

I thought we had to make the distinction between event, place and organizer to make sure the correct index gets updated?

A functional test is still needed.

bertramakers:
> > single ContributorsUpdated message
> 
> @bertramakers I thought we had to make the distinction between event, place and organizer to make sure the correct index gets updated?
> 
> A functional test is still needed.

Right, overlooked that. That is easier with 3 messages. It's also possible based on the JSON but less "clean" and a bit hard with the current code in SAPI3 I think.

JonasVHG:
> IMHO it would also be okay to have a single ContributorsUpdated message with just the IRI, that requires a lot less code and will also achieve the goal of making it possible for SAPI3 to index the contributors. If we ever need this event for something else where we cannot use the IRI but need the id/type we can easily add those as well (but still keep it 1 class). This is just a suggestion though, the approach with 3 events also works.
> 
> To actually get the messages published on AMQP you will need to map the class name(s) of the event(s) here to a content-type:
> 
> https://github.com/cultuurnet/udb3-backend/blob/7edebd7770f89983a4579a690afcbc1432edddc6/app/AMQP/AMQPPublisherServiceProvider.php#L47-L51
> 
> (You can follow the same structure as the existing content-types to invent new ones). The event(s) also need(s) to implement Serializable, but that is already the case I see 👍

I've tested this with CloudAmqp And I see messages when I locally update contributors.

LucWollants (approved):

PR #1264, opened 2023-02-20
III-4421 places actortype2eventtype

JonasVHG:
### Added

- Command to update `places` which have a deprecated `actorType` and give them an correct `eventType`

---
Ticket: https://jira.uitdatabank.be/browse/III-4421


Makefile
```diff
@@ -1,11 +1,14 @@
-.PHONY: up down install ci stan cs cs-fix test migrate
+.PHONY: up down install ci stan cs cs-fix test migrate ssh
 
 up:
 	docker-compose up -d
 
 down:
 	docker-compose down
 
+ssh:
```
LucWollants, line 9: I added the same command but with name `bash`

JonasVHG: I just missed it, I've removed it from this PR.

bertramakers:
Code looks good, my only real suggestion would be regarding Corneel's suggestion:

> Option 2 gives a little more flexibility but since this script may only be run once, that flexibility is probably not necessary?

While the flexibility is okay, it's kind of limited now because the person who runs the command can only pick from a list of known actor types and event types. So since we're already restricting that, I think we could just as well provide the mapping itself and not ask for input. That we the command also only needs to run once instead for every actortype.

bertramakers (approved):

PR #1266, opened 2023-02-20
Ask for replay confirmation

LucWollants:
### Added
- Added confirmation question before starting a replay

### DIFF
--- a/app/Console/Command/ReplayCommand.php
+++ b/app/Console/Command/ReplayCommand.php
@@ -68,8 +68,7 @@ final class ReplayCommand extends AbstractCommand
             ->addArgument(
                 'aggregate',
                 InputArgument::OPTIONAL,
-                'Aggregate type to replay events from. One of: ' . $aggregateTypeEnumeration . '.',
-                null
+                'Aggregate type to replay events from. One of: ' . $aggregateTypeEnumeration . '.'
             )
             ->addOption(
                 self::OPTION_START_ID,
--- a/app/Console/Command/ReplayCommand.php
+++ b/app/Console/Command/ReplayCommand.php
@@ -240,14 +240,28 @@ final class ReplayCommand extends AbstractCommand
         $cdbids = $input->getOption(self::OPTION_CDBID);
 
         $message = 'Are you sure you want to replay all events? [y/N] ';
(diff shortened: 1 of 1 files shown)


app/Console/Command/ReplayCommand.php
```diff
@@ -228,4 +233,30 @@ private function getAggregateType(InputInterface $input): ?AggregateType
...
+
+        $message = 'Are you sure you want to replay all events? [y/N] ';
+        if ($aggregateType) {
+            $message = 'Are you sure you want to replay all events for aggregate type ' . $aggregateType->toString() . '? [y/N] ';
+        }
+        if ($startId) {
+            $message = 'Are you sure you want to replay all events starting from id ' . $startId . '? [y/N] ';
+        }
+        if ($cdbids) {
+            $message = 'Are you sure you want to replay all events with giving cdbid? [y/N] ';
```
bertramakers, line 251:
```suggestion
            $message = 'Are you sure you want to replay all events with given cdbid? [y/N] ';
```

bertramakers (commented):
I think you can combine options, for example `--aggregateType=... --startId=...`. In those cases the messages may not be 100% correct and it might seem like you used a wrong combination?

app/Console/Command/ReplayCommand.php
```diff
@@ -240,14 +240,28 @@ private function askConfirmation(InputInterface $input, OutputInterface $output)
...
+            }
+
+            if ($cdbids) {
+                if (!empty($options)) {
+                    $options .=', ';
+                }
+                $options .= 'given cdbids';
+            }
+
+            $message = 'Are you sure you want to replay events ( ' . $options . ' )? [y/N]';
```
bertramakers, line 264: Nit: You can make `$options` an array of strings and do `implode(', ' $options)` here to concatenate them without the repeated `if (!empty($options)) { $options .=', '; }` checks.
(If you do an implode on an array with no values it will return an empty string, with 1 value it will return that value, and with multiple values it will concatenate them with the given "glue" in between every value)

bertramakers (approved):
Nice, a good improvement! Thanks for processing the feedback 👍

PR #1267, opened 2023-02-20
Iii 4422 places with event eventtypes

JonasVHG:
### Fixed

---

### DIFF
--- a/app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
+++ b/app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
@@ -1,0 +1,204 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Model\ValueObject\Identity\ItemIdentifier;
+use CultuurNet\UDB3\Offer\Commands\UpdateType;
+use CultuurNet\UDB3\Place\Commands\Moderation\Reject;
+use CultuurNet\UDB3\Search\ResultsGenerator;
+use CultuurNet\UDB3\Search\SearchServiceInterface;
+use CultuurNet\UDB3\StringLiteral;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Logger\ConsoleLogger;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+final class ChangePlaceTypeOnPlacesWithEventEventType extends AbstractCommand
+{
+    private SearchServiceInterface $searchService;
+
+    public function __construct(
+        CommandBus $commandBus,
+        SearchServiceInterface $searchService
+    ) {
+        $this->searchService = $searchService;
+        parent::__construct($commandBus);
+    }
+
+    public function configure(): void
+    {
+        $this->setName('place:eventtype:update:reject');
+
+        $this->setDescription(
+            'Updates all places with an Event eventType and optionally reject them'
+        );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $output->setVerbosity(OutputInterface::VERBOSITY_VERY_VERBOSE);
+        $logger = new ConsoleLogger($output);
+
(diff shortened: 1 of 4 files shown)


app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
```diff
@@ -0,0 +1,204 @@
...
+            'Updates all places with an Event eventType and optionally reject them'
+        );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $output->setVerbosity(OutputInterface::VERBOSITY_VERY_VERBOSE);
+        $logger = new ConsoleLogger($output);
+
+        $eventEventType = $this->askEventEventType($input, $output);
```
LucWollants, line 46: If I read the ticket it seems there is a mapping between the existing event type and the new event type. But with this script both the existing and new event type are asked?

app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
```diff
@@ -0,0 +1,204 @@
...
+            ['created' => 'asc'],
+            100
+        );
+
+        $query = 'terms.id:' . $eventEventType;
+        if (!$this->askConfirmation($input, $output, $resultsGenerator->count($query))) {
+            return 0;
+        }
+
+        $reject = $this->askToReject($input, $output);
```
LucWollants, line 60: Should this not be an option of the console command instead of a user question?

app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
LucWollants: If only the event type is required it also seems better to make this an argument.

app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
JonasVHG: In think if you make it an option, it could easily be forgotten. And once the type has been changed, it would be a lot harder to find those places which had an `EventType` for `Events`.
So I think it would be wiser to explicitly asked the user.

app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
JonasVHG: Agreed, I had flexibility in mind, but that might be overkill for this case.

bertramakers: I would either go for full flexibility (asking the "from" and "to"), or none at all and then this command could just go over all known incorrect types and convert them to the right ones.
I don't think there's a use case to run the command for one incorrect type on its own.

app/Console/Command/ChangePlaceTypeOnPlacesWithEventEventType.php
bertramakers: I think the rejecting should be a different command

LucWollants: > I think the rejecting should be a different command, according to the ticket:

Will that not make it tricky to keep track of the modified places?

bertramakers: True, if we give them a new eventtype first. But I'm now wondering why that is even necessary in the case of places that should be rejected. Feels like trying to clean up data that does not matter anyway 🤔 (And it's not like a place that's about a concert is suddenly a semantically correct place just by getting the type "Zaal of expohal" - which is also why we reject them)

Anyway, we can keep it 1 command but if we go the route of not making it flexible at all I would also not ask this at all but infer it from the info in the ticket.

bertramakers: Clarified in the ticket: We will change the eventtype but not update the workflowStatus

bertramakers (commented):
Thanks for the changes @JonasVHG ! My only last suggestion would be to use a more specific name for the command, one that indicates that it will fix incorrect eventtypes on place. Since `place:eventtype:update` is probably a bit too generic and maybe confusing in the future.

Edit: If this CLI command is removed again in the near future, after it has run on all environments, the name does not matter _that_ much so it can be merged as-is.

bertramakers (approved):
Already approving because the suggestion about the name is minor

PR #1268, opened 2023-02-20
III-4423 Reject places with deprecated actorTypes: 8.15.0.0.0 8.0.0.0.0

JonasVHG:
### Added

- `Command` to reject `Places` which have a termId of `8.15.0.0.0` or `8.0.0.0.0`

---
Ticket: https://jira.uitdatabank.be/browse/III-4423


app/Console/Command/RejectPlaceDeprecatedType.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $output->setVerbosity(OutputInterface::VERBOSITY_VERY_VERBOSE);
+        $logger = new ConsoleLogger($output);
+
+        $deprecatedTypeId = $this->askDeprecatedType($input, $output);
+
+        $resultsGenerator = new ResultsGenerator(
```
JonasVHG, line 47: There is some duplcated code in my PR's, so I thought a trait might be useful. But since these command should probably be removed after they have served their purpose, it might be overkill?

app/Console/Command/RejectPlaceDeprecatedType.php
```diff
@@ -0,0 +1,122 @@
...
+            'Rejects all places with a given actorType'
+        );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $output->setVerbosity(OutputInterface::VERBOSITY_VERY_VERBOSE);
+        $logger = new ConsoleLogger($output);
+
+        $deprecatedTypeId = $this->askDeprecatedType($input, $output);
```
bertramakers, line 45: As in the other commands, I think we should just run this automatically for all known deprecated types.

bertramakers (approved):

PR #1269, opened 2023-02-20
Add script to find out of sync projections

LucWollants:
### Added
- Added script to find out-of-sync projections based on a first and last row id

### Note
- This script is very basic, it needs to be finetuned but can be used to get a feeling of the amount of data to process and how well it fits the expected format.

### DIFF
--- a/app/Console/Command/FindOutOfSyncProjections.php
+++ b/app/Console/Command/FindOutOfSyncProjections.php
@@ -1,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use CultuurNet\UDB3\AggregateType;
+use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
+use CultuurNet\UDB3\ReadModel\DocumentRepository;
+use DateTime;
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class FindOutOfSyncProjections extends Command
+{
+    private Connection $connection;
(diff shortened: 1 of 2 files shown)


app/Console/Command/FindOutOfSyncProjections.php
```diff
@@ -0,0 +1,123 @@
...
+            $uuid = $result['uuid'];
+            $recorded = $result['recorded'];
+            $modified = $this->getModified($uuid, $aggregateType, $output);
+
+            if ($modified) {
+                $recordedDate = new DateTime($recorded);
+                $modified = new DateTime($modified);
+
+                $difference = $modified->diff($recordedDate);
+                if ($difference->s > 5) {
```
bertramakers, line 93: The `modified` is based on the last `recorded_on`, so it should be exactly the same in theory.

- The code for offers: https://github.com/cultuurnet/udb3-backend/blob/7f01c6d10e34612b08041b5ed13bcaabcc2d41bf/src/Offer/ReadModel/JSONLD/OfferLDProjector.php#L839
- The code for organizers: https://github.com/cultuurnet/udb3-backend/blob/0f5004011dc77b74afa500bf27780b6e170e5566/src/Organizer/OrganizerLDProjector.php#L655
- And for example in a test assertion: https://github.com/cultuurnet/udb3-backend/blob/9b1b6befd85083ea4860ddea9a5b50d0f7e864be/tests/OfferLDProjectorTestBase.php#L128

LucWollants: Maybe we check for a 1-second difference because the `recorded_on` inside the database has milliseconds and the `modified` inside the json projection only seconds?

bertramakers: You can convert the `recorded_on` to the exact value that it should be in `modified` using `RecordedOn::toString()` and compare that they are exactly the same or not? We don't need to know the date/time difference, just that they match or not.

bertramakers (approved):

PR #1274, opened 2023-02-22
III-5334 publish ordinary messages

JonasVHG:
### Added

- `ContributorEnrichedRepository` to show `contributors` in `JSON-LD`

### Changed

- Added `ContributorEnrichedRepository` to `EventJSONLDServiceProvider`, `PlaceJSONLDServiceProvider` & `OrganizerJSONLDServiceProvider`
- Use `authenticateToken()` to parse token on `publicRoutes`
- Dispatch a `*ProjectedToJSONLD` when `contributors` are added instead of a `*ContributorsUpdated`
- Refactor `ContributorsUpdatedFactory` to Generic `ProjectedToJSONLDFactory`

### Removed

- `ContributorsUpdated` and child classes: `EventContributorsUpdated`, `PlaceContributorsUpdated` & `OrganizerContributorsUpdated`

---


bertramakers (commented):
Looks good overall, my only suggestion would be not to add a new `checkToken()` method but instead call `authenticateToken()`, catch any Exception that it throws and ignore that if the route is public or CORS-related, or re-throw it if a token is required.

app/Contributor/ContributorServiceProvider.php
```diff
@@ -27,15 +26,9 @@ public function register(): void
                 new DbalContributorRepository($container->get('dbal_connection')),
                 $container->get(EventBus::class),
                 new ContributorsUpdatedFactory(
```
LucWollants, line 28: The name for this factory is a bit weird now `ContributorsUpdatedFactory` because those messages no longer exist.

bertramakers: Maybe there is already another similar factory for this, now that we return the existing `ProjectedToJSONLD` messages?

JonasVHG: There does not seem to be a Generic ProjectedToJSONLDFactory, maybe now is a good time to create one?

src/Http/Auth/RequestAuthenticatorMiddleware.php
```diff
@@ -83,14 +83,15 @@ public function authenticate(ServerRequestInterface $request): void
...
         // Users. eg. show contributors
-        $this->checkToken($request);
-
-        if ($this->isCorsPreflightRequest($request) || $this->isPublicRoute($request)) {
-            return;
+        try {
+            $this->authenticateToken($request);
+        } catch (\Exception $exception) {
+            if ($this->isCorsPreflightRequest($request) || $this->isPublicRoute($request)) {
+                return;
```
bertramakers, line 90: Probably best to also do `$this->token = null;` first like you did in `checkToken()`, since otherwise some parts of the code may think that the token was successfully authenticated on these public/CORS routes and incorrectly return the `contributors` property for example (depending on where the authentication failed).

app/Contributor/ContributorServiceProvider.php
bertramakers: Sounds good!

src/Contributor/ContributorEnrichedRepository.php
```diff
@@ -0,0 +1,74 @@
...
+
+    public function fetch(string $id, bool $includeMetadata = false): JsonDocument
+    {
+        $jsonDocument = parent::fetch($id, $includeMetadata);
+
+        if ($this->hasPermission($jsonDocument->getId())) {
+            $jsonDocument = $this->enrich($jsonDocument);
+        } else {
+            $jsonDocument = $jsonDocument->applyAssoc(
+                function (array $json) {
```
JonasVHG, line 43: For some reason this is necessary for places. If you are unknown or unauthorized you get an empty array for contributors in the JSON Projection of a `Place`. This does not happen with the Projection of an `Event` or an `Organizer`.

LucWollants: I would search for the root cause of this issue. Do you get it inside a unit test or when running the acceptance tests or when doing a manual test?

JonasVHG: This was the problem: 
`            function (array $body) use ($contributors) {
                    $body['contributors'] = $contributors;
                return $body;
            }`
            Solution was to add a check to see if the `array` is empty or not.

LucWollants: As discussed through Slack the issue is a little bit more subtle than this. When a projection gets updated by an event the existing projection gets `fetched` from Redis, the change related to the event gets applied, and the JSON gets `saved` again.

With this extra enricher to add contributors the `fetch` method will always return JSON with contributors when authenticated and they will get saved in Redis. So the next time the JSON projection is `fetched` the contributors are still on the JSON.

For other enriched data like popularity, this will not happen because the popularity scores are only added when `includeMetadata` is true. That flag is always false inside the projectors and only true inside the HTTP request handlers.

Proposed solution is to only embed the contributors when `includeMetadata` is true.

JonasVHG: Proposed solution is to only embed the contributors when `includeMetadata` is true will not work cause `Organizer` endpoint does not have `metadata`

JonasVHG: solution:`contributors` will be shown in the JsonBody and are removed in `save()`.
Unit & Acceptance tests have been added to verify this behaviour.

bertramakers (approved):
Nice work 👌

PR #1278, opened 2023-02-28
III-5356 Project place TitleUpdated and TitleTranslated to RDF/OSLO

bertramakers:
### Added

- `TitleUpdated` and `TitleTranslated` projections in RDF/OSLO

---
Ticket: https://jira.uitdatabank.be/browse/III-5356

Notes:

- This only projects the "naam" property of "locaties" so far (based on the ticket scope), other properties including the type etc will be added in later tickets. Focus of this ticket is to have a working projector that can project 1 property.
- You will need the changes from https://github.com/cultuurnet/udb3-vagrant/pull/147#pullrequestreview-1322014993 in a local vagrant box so you have a local Fuseki triple store
- You will need to add the following config in `config.php` (I'll open a PR in udb3-vagrant for that as well):
```
    'rdf' => [
        'enabled' => true,
        'resourceBaseUri' => 'https://data.publiq.dev/',
        'placesGraphStoreUrl' => 'http://localhost:3030/locaties/',
    ]
```
- This should not cause any errors on environments where there is no triple store (yet) because all of the config is optional and the RDF-related code is not initiated if the `rdf.enabled` key is missing or not set to true


src/Place/ReadModel/RDF/RdfProjector.php
```diff
@@ -0,0 +1,118 @@
...
+
+    public function handle(DomainMessage $domainMessage): void
+    {
+        $payload = $domainMessage->getPayload();
+        $granularEvents = $payload instanceof ConvertsToGranularEvents ? $payload->toGranularEvents() : [];
+        $events = [$payload, ...$granularEvents];
+
+        $uri = $this->iriGenerator->iri($domainMessage->getId());
+
+        $mapping = [
```
simon-debruijn, line 46:
```suggestion
        $eventClassToHandler = [
```

You could rename it to something like `$eventClassToHandler`. `$mapping` is a bit abstract. But since it's only used here it's clear enough as it is.

src/Place/ReadModel/RDF/RdfProjector.php
```diff
@@ -0,0 +1,118 @@
...
+            if ($literal->getLang() === $language) {
+                $resource->delete($property, $literal);
+            }
+        }
+
+        $resource->add($property, [
+            'type' => 'literal',
+            'value' => $value,
+            'lang' => $language,
+        ]);
```
simon-debruijn, line 116:
```suggestion
        $resource->add(
                $property,
                new Literal($value, $language).toArray()
        );
```

Maybe a literal could be created with its own class?

src/RDF/GraphStoreRepository.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace CultuurNet\UDB3\RDF;
+
+use EasyRdf\Graph;
+use EasyRdf\GraphStore;
+use EasyRdf\Http\Exception;
+
+final class GraphStoreRepository implements GraphRepository
+{
+    private const GRAPH_URI_SUFFIX = '.ttl';
```
simon-debruijn, line 13: 🐢

simon-debruijn (approved):
Looks really good overall 👍🏻

src/Place/ReadModel/RDF/RdfProjector.php
bertramakers: Apparently the `easyrdf/easyrdf` library already has a `Literal` class which is supported in this method, so I've used that instead.

src/RDF/GraphStoreRepository.php
bertramakers: 🐇

simon-debruijn (approved):
Perfect ✅

bertramakers:
@simon-debruijn I also just pushed a small change so that the order of language-tagged values remains the same when a value for a single language is updated, instead of that updated value always being appended to the end of the list.

simon-debruijn:
> 

Nice catch code looks good 👍🏻

PR #1280, opened 2023-03-01
Switch to Apache2 container

LucWollants:
### Changed
- Switched to Apache2 container


JonasVHG (approved):
Just tested it, works great 👍

PR #1281, opened 2023-03-01
III-5389 Add MainLanguageDefined interface

bertramakers:
### Added

- Added `MainLanguageDefined` interface, implemented on events that create new events|places. This was initially added in #1278 but I split it off to keep the PRs focused on a small scope.

---
Ticket: https://jira.uitdatabank.be/browse/III-5389


simon-debruijn (approved):

bertramakers:
Ticket reference was wrong btw, should have been III-5398 instead of III-5389. I've fixed it (except for the branch name, which is deleted now anyway)

PR #1284, opened 2023-03-03
III-5392 Project common place properties

bertramakers:
### Added

- Projection of `rdf:type` property
- Projection of `dcterms:created` property ("aangemaaktOp" in OSLO generiek)
- Projection of `dcterms:modified` property ("laatstAangepast" in OSLO generiek)
- Projection of `adms:identifier` resource with:
  - `skos:notation`
  - `dcterms:creator`
  - `generiek:naamruimte` (from OSLO generiek)
  - `generiek:lokaleIdentificator` (from OSLO generiek)
  - `generiek:versieIdentificator` (from OSLO generiek)

---
Ticket: https://jira.uitdatabank.be/browse/III-5392

Notes:

- The correct URI for the `dcterms:creator` property inside `adms:Identifier` is not decided yet, needs to be replaced later (hence the "fixme.com" URL so it's clear that this is not final)
- We will also need to add a "bron" attribute to the `adms:Identifier`, but the URI for that property is not clear yet + the URI for the value neither, so I've chosen not to include that yet. We can add a ticket for that later.
- We can extract the code for setting the common properties and identificator to a separate class when we also implement the RDF/OSLO projection for events / "activiteiten", but for now I've opted to keep it simple to avoid bad abstractions early on

### DIFF
--- a/src/RDF/RdfNamespaces.php
+++ b/src/RDF/RdfNamespaces.php
@@ -10,6 +10,10 @@ final class RdfNamespaces
 {
     public static function register(): void
     {
+        // Delete the initial 'dc' prefix for dcterms, so the other initial 'dcterms' namespace is used instead.
+        RdfNamespace::delete('dc');
+
+        // Set custom namespaces.
         RdfNamespace::set('locn', 'http://www.w3.org/ns/locn#');
     }
 }
--- a/src/RDF/RdfNamespaces.php
+++ b/src/RDF/RdfNamespaces.php
@@ -14,6 +14,7 @@ final class RdfNamespaces
         RdfNamespace::delete('dc');
 
         // Set custom namespaces.
+        RdfNamespace::set('adms', 'http://www.w3.org/ns/adms#');
         RdfNamespace::set('locn', 'http://www.w3.org/ns/locn#');
     }
 }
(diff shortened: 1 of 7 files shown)


src/RDF/RdfNamespaces.php
```diff
@@ -10,6 +10,12 @@ final class RdfNamespaces
 {
     public static function register(): void
     {
+        // Delete the initial 'dc' prefix for dcterms, so the other initial 'dcterms' namespace is used instead.
+        RdfNamespace::delete('dc');
+
+        // Set custom namespaces.
+        RdfNamespace::set('adms', 'http://www.w3.org/ns/adms#');
         RdfNamespace::set('locn', 'http://www.w3.org/ns/locn#');
+        RdfNamespace::set('generiek', 'https://data.vlaanderen.be/ns/generiek/#');
```
simon-debruijn, line 19: So the namepace `dcterms` is already known without setting it right?

simon-debruijn (approved):

src/RDF/RdfNamespaces.php
bertramakers: Yep, it's one of the defaults provided by EasyRDF itself in the `EasyRdf\RdfNamespace` class in its `$initial_namespaces` property. It also sets the `dc` namespace by default, which I had to remove because they were both the same and it would otherwise always use `dc` instead of `dcterms` (which we usually use in examples of OSLO data etc so that's probably clearer than `dc`).

PR #1286, opened 2023-03-06
III-5405 remove themes on places

JonasVHG:
### Changed

- Polyfill to remove `terms` from Domain `theme` on `Places`.
- UnitTest to check removal of `themes` from `Places`
- UnitTest to check `Event` keeps `themes`

---
Ticket: https://jira.uitdatabank.be/browse/III-5405


bertramakers (approved):
Maybe the `if ($this->offerType->sameAs(OfferType::place()))` could be moved inside `removeThemes()` with an early return there instead if it's not a place, to keep the list of method calls inside `fetch()` as easy to read as possible at a glance, but that's a very minor nit and subjective. Everything else looks great!

PR #1289, opened 2023-03-08
III-5412 Feature Toggle on Public Routes

JonasVHG:
### Fixed

---
Ticket: https://jira.uitdatabank.be/browse/III-5412

### DIFF
--- a/src/Http/Auth/RequestAuthenticatorMiddleware.php
+++ b/src/Http/Auth/RequestAuthenticatorMiddleware.php
@@ -84,17 +84,18 @@ final class RequestAuthenticatorMiddleware implements MiddlewareInterface
      */
     public function authenticate(ServerRequestInterface $request): void
     {
-        // For requests to public routes, that provide extra information to Authenticated
-        // Users. eg. show contributors
-        try {
-            $this->authenticateToken($request);
-        } catch (\Exception $exception) {
-            if ($this->isCorsPreflightRequest($request) || $this->isPublicRoute($request)) {
-                $this->token = null;
-                return;
-            }
-            throw $exception;
+        if ($this->isCorsPreflightRequest($request)) {
+            return;
         }
+
+        if ($this->isPublicRoute($request)) {
+            if ($this->authenticatePublicRoutes) {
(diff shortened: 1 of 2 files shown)


src/Http/Auth/RequestAuthenticatorMiddleware.php
```diff
@@ -81,18 +84,24 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+        }
+
+        if ($this->isPublicRoute($request)) {
+            if ($this->authenticatePublicRoutes) {
+                try {
+                    $this->authenticateToken($request);
+                } catch (\Exception $exception) {
+                    $this->token = null;
+                    return;
+                }
```
simon-debruijn, line 98:
```php
    $isPublic = $this->isPublicRoute($request);

    if ($isPublic && $this->authenticatePublicRoutes) {
      try {
        $this->authenticateToken($request);
      } catch (\Exception $exception) {
        $this->token = null;
      }
    }

    if ($isPublic) {
      return;
    }
```

I think you could write it less nested like this.

Also you could rename `authenticatePublicRoutes` to `shouldAuthenticatePublicRoutes`, down to preferences.

LucWollants: Or even with an extra early return statement:

```
if ($isPublic && !$this->authenticatePublicRoutes) {
  return;
}

if ($isPublic) {
  try {
    $this->authenticateToken($request);
  } catch (\Exception $exception) {
    $this->token = null;
  }
}
```

simon-debruijn: Yeah, but If I'm not mistaken you also have to return in both cases:

When authenticatePublicRoutes is true or false.
So you would have to ass a return statement after the try catch

JonasVHG: Indeed, you have to return in both cases regardless of `authenticatePublicRoutes`.

simon-debruijn (approved):

PR #1291, opened 2023-03-08
III-5412 keep logic the same if toggle is set to true

JonasVHG:
### Fixed

- Refactored `authenticate()` in `RequestAuthenticatorMiddleware` so that exception is thrown on invalid token & unauthorized `apiKey`, if toggle `authenticate_public_routes` is enbled.

---
Ticket: https://jira.uitdatabank.be/browse/III-5412

### DIFF
--- a/src/Http/Auth/RequestAuthenticatorMiddleware.php
+++ b/src/Http/Auth/RequestAuthenticatorMiddleware.php
@@ -92,17 +92,13 @@ final class RequestAuthenticatorMiddleware implements MiddlewareInterface
             return;
         }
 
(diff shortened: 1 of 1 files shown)


bertramakers:
As mentioned on Slack this works as expected for the API key errors, but now there is never an error when you use an invalid token (even when the feature toggle is enabled). Since SAPI3 expects a 401 for an expired token, and it would be the most consistent/transparent, we should also return an error for invalid/expired/... tokens. Only when the token is completely missing should there be no error in the case of public routes.

bertramakers (approved):

PR #1293, opened 2023-03-16
III-5351 place imported from udb2 to smaller events

JonasVHG:
### Added

- `SerializableSimpleXmlElement` to convert complex `SimpleXmlElement` to `array`
- `trait` `PlaceFromUDB2` to Convert Places from UDB2 to granular Events
- test file: `actor_with_translations.xml`
- `PlaceUpdatedFromUDB2Test`: test conversion to granular events

### Changed

- Make `PlaceImportedFromUDB2` & `PlaceUpdatedFromUDB2` implement `ConvertsToGranularEvents`
- `PlaceImportedFromUDB2Test`: tests to check conversion to granular events
- boy scouting

---
Ticket: https://jira.uitdatabank.be/browse/III-5351

### DIFF
--- a/src/Place/Events/PlaceFromUDB2.php
+++ b/src/Place/Events/PlaceFromUDB2.php
@@ -48,7 +48,7 @@ trait PlaceFromUDB2
         return $granularEvents;
     }
 
-    public function getPlaceAsArray(): array
+    private function getPlaceAsArray(): array
     {
         $cdbXml = new SerializableXML(
             $this->cdbXml,
--- a/src/Place/Events/PlaceFromUDB2.php
+++ b/src/Place/Events/PlaceFromUDB2.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\Events;
+
+use CultureFeed_Cdb_Data_ActorDetail;
+use CultureFeed_Cdb_Data_Address;
+use CultureFeed_Cdb_Item_Actor;
+use CultuurNet\UDB3\Address\Address;
+use CultuurNet\UDB3\Address\Locality;
+use CultuurNet\UDB3\Address\PostalCode;
+use CultuurNet\UDB3\Address\Street;
+use CultuurNet\UDB3\Language;
+use CultuurNet\UDB3\Model\ValueObject\Geography\CountryCode;
+use CultuurNet\UDB3\Title;
+
+trait PlaceFromUDB2
+{
+    public function toGranularEvents(): array
+    {
+        $granularEvents = [];
+        $cultureFeedActor = $this->getCultureFeedActor();
+        $details = $cultureFeedActor->getDetails();
+        $firstDetail = $details->getFirst();
+
+        $granularEvents[] = new TitleUpdated($this->actorId, new Title($firstDetail->getTitle()));
+
+        $details->next();
+        while ($details->valid()) {
+            /** @var CultureFeed_Cdb_Data_ActorDetail $detail */
+            $detail = $details->current();
+            $granularEvents[] = new TitleTranslated(
+                $this->actorId,
+                new Language($detail->getLanguage()),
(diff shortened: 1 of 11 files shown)


src/Place/Events/PlaceFromUDB2.php
```diff
@@ -0,0 +1,68 @@
...
+                new PostalCode($physicalAddress->getZip()),
+                new Locality($physicalAddress->getCity()),
+                new CountryCode($physicalAddress->getCountry())
+            )
+        );
+
+        return $granularEvents;
+    }
+
+    private function getCultureFeedActor(): CultureFeed_Cdb_Item_Actor
```
JonasVHG, line 57: Unfortunately, Conversion from a complex `XML` with namespaces & attributes to an array is not as simple as `Json::decodeAssociatively(Json::encode($xml))`

src/Place/Events/PlaceFromUDB2.php
```diff
@@ -0,0 +1,68 @@
...
+                new Language($detail->getLanguage()),
+                new Title($detail->getTitle())
+            );
+            $details->next();
+        }
+        /** @var CultureFeed_Cdb_Data_Address $address */
+        $address = $cultureFeedActor->getContactInfo()->getAddresses()[0];
+        $physicalAddress = $address->getPhysicalAddress();
+
+        $granularEvents[] = new AddressUpdated(
```
JonasVHG, line 44: CdbXML only allowed for Translations of the DetailsNode. So unlike in JSON an address could not be translated. hence an `AddressTranslated` cannot appear here.

src/Place/Events/PlaceFromUDB2.php
```diff
@@ -0,0 +1,68 @@
...
+            $this->actorId,
+            new Address(
+                new Street($physicalAddress->getStreet() . ' ' . $physicalAddress->getHouseNumber()),
+                new PostalCode($physicalAddress->getZip()),
+                new Locality($physicalAddress->getCity()),
+                new CountryCode($physicalAddress->getCountry())
+            )
+        );
+
+        return $granularEvents;
```
JonasVHG, line 54: GeoCodes from an XML are ignored, and recalculated, so an `GeoCoordinatesUpdated` should not be included.

bertramakers: Thanks for checking!

src/Place/Events/PlaceFromUDB2.php
bertramakers: Okay, makes sense!

src/Place/Events/PlaceFromUDB2.php
bertramakers: And there is no feasible way to parse the XML using `SimpleXMLElement` without CultureFeed? Because with this implementation, we won't be able to remove the `cultuurnet/culturefeed-php` dependency in the future.

JonasVHG: I've found several gists to convert XML with namespaces and attributes to JSON, and they also worked.
On the one hand it would mean introducing new extra code to `udb3-backend`, but it would be a lot less than `cultuurnet/culturefeed-php`. So in the long run it might be a good thing?

src/SerializableSimpleXmlElement.php
```diff
@@ -22,25 +22,27 @@ public function serialize(): array
...
                 }
 
-                $array[$tag][] = $temp;
+                // add the node to the array with the tagname as key
+                $array[$tag][] = $serializedChild;
             }
         } else {
             // serialize attributes and text for a leaf-elements
-            $temp = trim((string) $this);
+            $serialize = trim((string) $this);
```
simon-debruijn, line 41:
```suggestion
            $serialized = trim((string) $this);
```

The naming `serialize` suggests that it is a function

simon-debruijn (approved):

PR #1307, opened 2023-03-29
Docker config setup

JonasVHG:
### Added
- Shellscript `config.sh` to automatically retrieve and setup config for docker from `appconfig`.
- `make config` commando to run the shellscript.
- `make init` to setup `composer` packages & database with the same commando.

### Changed
- Added info about the shellscript to `docker.md`
- Added `appconfig` as a `prerequisite`

### Removed
- Manual config setup in `docker.md`


docker.md
```diff
@@ -13,38 +13,7 @@ Copy `env.dist` to the root folder and rename it to `.env`
...
-        'scheme' => 'http',
-        'port' => 9000,
-    ],
-],
-```
-
-### pem files
-
-Copy `public.pem` and `public-auth0.pem` from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/keys to the root
+Copy `docker-config.php` to the root folder and rename it to `config.php`
```
bertramakers, line 16: Maybe the PHP docker image can do that by itself in its Dockerfile? (https://github.com/cultuurnet/udb3-backend/blob/master/docker/php/Dockerfile)

docker-config.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+return array_replace_recursive(
```
bertramakers, line 3: Cool, I didn't know this existed 👍

docker.md
JonasVHG: I am afraid it cannot be done in the `Dockerfile`. That creates the container.
Only in the `docker-compose.yml` (after the dockerfile) the local filesystem is linked to the container.
Alternatively it could be done in the shellscript, so that we automatically create the `config.php` & `.env` with a `make config`

bertramakers: Sounds good to me!

docker.md
```diff
@@ -76,30 +29,34 @@ search_api_base_url: 'http://host.docker.internal:9000'
...
 
 To execute all CI tasks, run the following command:
 ```
 $ make ci
-```
\ No newline at end of file
+```
+
+## Config changes
+To avoid duplicate out-of-sync development config files,
```
LucWollants, line 60: Is this still correct?

docker/setup_config.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# copy env files
+cp .env.dist .env
+
+# setup config & key files
+DIR="../appconfig/files/udb3/docker/udb3-backend/"
+if [ -d "$DIR" ]; then
+  cp "$DIR"/* .
+else
+  echo "Error: see docker.md prerequisites to setup config"
```
LucWollants, line 11: The error can also provide the solution, missing appconfig

Makefile
```diff
@@ -28,4 +28,9 @@ migrate:
 	docker exec -it php.uitdatabank ./vendor/bin/doctrine-dbal migrations:migrate --no-interaction
 
 bash:
-	docker exec -it php.uitdatabank bash
\ No newline at end of file
+	docker exec -it php.uitdatabank bash
+
+config:
+	sh ./docker/setup_config.sh
```
LucWollants, line 34: Maybe for consistency change the filename from `setup_config.sh` to `config.sh`

docker.md
```diff
@@ -76,30 +29,34 @@ search_api_base_url: 'http://host.docker.internal:9000'
...
 ## Start
 
-### Docker
-
-Start the docker containers with the following command. Make sure to execute this inside the root of the project so the `.env` can be used.
+### Configuration setup
+To get or update the configuration files, run the following command in the root of the project
 ```
-$ make up
+$ make config
```
LucWollants, line 35: I think this should be added to the Configure chapter.

docker.md
JonasVHG: Indeed.

LucWollants (approved):
🎉

PR #1311, opened 2023-03-31
III-4684 Offers with MainImage and no MediaObject

JonasVHG:
### Added
-  `OfferLDProjectorTest`: added test to check if `ImagedRemoved` with `http`-link unsets the same MainImage with `https`

### Changed
- `OfferLDProjector`: make comparison with `mainImageUrl` protocol-agnostic, when an `ImageRemoved` is applied.
- `OfferLDProjector`: always unset `mainImageUrl` if no `MainObjects` remain

### Fixed
- `ImageRemoved` with `http` can now unset the `mainImageUrl` if it is the same but url with `https`

---
Ticket: https://jira.uitdatabank.be/browse/III-4684

### DIFF
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -418,7 +418,7 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
         );
 
         // Unset the main image if it matches the removed image
-        // @see https://jira.uitdatabank.be/browse/III-2201
+        // @see https://jira.uitdatabank.be/browse/III-4684
         // some historic ImageRemoved Events have http instead of https in the eventStore
         if (isset($offerLd->image) && stristr($offerLd->{'image'}, ':') === stristr($imageUrl, ':')) {
             unset($offerLd->{'image'});
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -418,7 +418,9 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
         );
 
         // Unset the main image if it matches the removed image
-        if (isset($offerLd->image) && $offerLd->{'image'} === $imageUrl) {
+        // stristr() comparison is done for edge cases where an offer has multiple images
+        // @see III-4684
+        if (isset($offerLd->image) && stristr($offerLd->{'image'}, ':') === stristr($imageUrl, ':')) {
             unset($offerLd->{'image'});
         }
 
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -418,7 +418,9 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
         );
 
         // Unset the main image if it matches the removed image
-        if (isset($offerLd->image) && $offerLd->{'image'} === $imageUrl) {
+        // @see https://jira.uitdatabank.be/browse/III-2201
+        // some historic ImageRemoved Events have http instead of https in the eventStore
+        if (isset($offerLd->image) && stristr($offerLd->{'image'}, ':') === stristr($imageUrl, ':')) {
             unset($offerLd->{'image'});
         }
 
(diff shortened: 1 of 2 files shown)


src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -418,7 +418,9 @@ protected function applyImageRemoved(AbstractImageRemoved $imageRemoved): ?JsonD
         );
 
         // Unset the main image if it matches the removed image
-        if (isset($offerLd->image) && $offerLd->{'image'} === $imageUrl) {
+        // @see https://jira.uitdatabank.be/browse/III-2201
```
LucWollants, line 421: I am confused about this linked ticket

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -418,7 +418,9 @@ protected function applyImageRemoved(AbstractImageRemoved $imageRemoved): ?JsonD
...
+        // some historic ImageRemoved Events have http instead of https in the eventStore
+        if (isset($offerLd->image) && stristr($offerLd->{'image'}, ':') === stristr($imageUrl, ':')) {
```
LucWollants, line 423: I had to read the documentation of `stristr` 🙂

If I understand correctly the `$offerLd->{'image'}` and `$imageUrl` get converted to string URLs without the protocol?

LucWollants: Maybe a better command is 
```
Unset the main image if it matches the removed image. The comparison is done after stripping the protocol because historical data can have http instead of https
```

LucWollants: Indeed comment 🙈

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -418,19 +418,19 @@ protected function applyImageRemoved(AbstractImageRemoved $imageRemoved): ?JsonD
...
             unset($offerLd->{'image'});
         }
 
         if (!isset($offerLd->image) && count($filteredMediaObjects) > 0) {
             $offerLd->image = array_values($filteredMediaObjects)[0]->contentUrl;
         }
 
-        // If no media objects are left remove the attribute.
+        // If no media objects are left remove the attribute and the imageUrl.
+        // @see III-4684
```
bertramakers, line 430: I think this is a sensible think to do, if there is no `mediaObject` property it never hurts to also unset the `image` just in case it's still present 👍 

However, I'd also keep the initial check that you added because the PR now doesn't cover this scenario for example:

- Event has 2 images, at least one imported from UDB2 using `http://`
- The image with `http://` is the main image
- The image with `http://` is removed

➡️ The removed image with `http://` will still be set as the main image if I'm not mistaken.

But this is also a very niche edge case given that XML imports are no longer possible. For existing places that were imported from UDB2 it might be relevant, but still a minor issue. I'd personally re-add the `stristr` check (and keep this extra `unset()`), but all in all your `unset()` when `mediaObject` is empty will also cover _most_ cases.

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -418,7 +418,9 @@ protected function applyImageRemoved(AbstractImageRemoved $imageRemoved): ?JsonD
         );
 
         // Unset the main image if it matches the removed image
-        if (isset($offerLd->image) && $offerLd->{'image'} === $imageUrl) {
+        // stristr() comparison is done for edge cases where an offer has multiple images
+        // @see III-4684
```
LucWollants, line 422: Maybe make this a real link?

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
LucWollants: @JonasVHG maybe best to also add unit test for this edge case?

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
JonasVHG: I've thought about this, but I had my doubts because It could crash the projector If there still some faulty/LegacyUrls in the eventStore. And it would still require some Url-manipulation. But maybe it would be easier to create a `sameAsProtocolAgnostic()` in `Url`?

LucWollants: I was talking about the ticket number, but I admit that was not clear 😉

LucWollants (approved):

PR #1313, opened 2023-04-04
III-5281 Add retry when saving to Redis results in different data

LucWollants:
### Added
- Added retry when saving to Redis results in different data

### Notes
- Config updated:
  - For UDB3 Vagrant: https://github.com/cultuurnet/udb3-vagrant/pull/156
  - For appconfig: https://github.com/cultuurnet/appconfig/pull/324

---
Ticket: https://jira.uitdatabank.be/browse/III-5281

### DIFF
--- a/app/Event/EventJSONLDServiceProvider.php
+++ b/app/Event/EventJSONLDServiceProvider.php
@@ -70,6 +70,7 @@ final class EventJSONLDServiceProvider extends AbstractServiceProvider
             function () use ($container): BroadcastingDocumentRepositoryDecorator {
                 $repository = new CacheDocumentRepository(
                     $container->get('event_jsonld_cache'),
+                    $container->get('config')['cache']['allowed_retries'] ?? 3,
                     $container->get('config')['cache']['milliseconds_between_retry'] ?? 0
                 );
                 $repository->setLogger(LoggerFactory::create($container, LoggerName::forWeb()));
--- a/app/Event/EventJSONLDServiceProvider.php
+++ b/app/Event/EventJSONLDServiceProvider.php
@@ -68,7 +68,10 @@ final class EventJSONLDServiceProvider extends AbstractServiceProvider
         $container->addShared(
             'event_jsonld_repository',
             function () use ($container): BroadcastingDocumentRepositoryDecorator {
-                $repository = new CacheDocumentRepository($container->get('event_jsonld_cache'));
+                $repository = new CacheDocumentRepository(
+                    $container->get('event_jsonld_cache'),
+                    $container->get('config')['cache']['milliseconds_between_retry'] ?? 0
+                );
                 $repository->setLogger(LoggerFactory::create($container, LoggerName::forWeb()));
 
                 $repository = EmbeddingRelatedResourcesOfferRepository::createForEventRepository(
--- a/src/Doctrine/ReadModel/CacheDocumentRepository.php
+++ b/src/Doctrine/ReadModel/CacheDocumentRepository.php
@@ -18,11 +18,13 @@ final class CacheDocumentRepository implements DocumentRepository
 
     private Cache $cache;
 
+    private int $allowedRetries;
     private int $millisecondsBetweenRetry;
 
-    public function __construct(Cache $cache, int $millisecondsBetweenRetry = 0)
+    public function __construct(Cache $cache, int $allowedRetries = 3, int $millisecondsBetweenRetry = 0)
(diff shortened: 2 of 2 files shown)


app/Event/EventJSONLDServiceProvider.php
```diff
@@ -68,7 +68,10 @@ public function register(): void
         $container->addShared(
             'event_jsonld_repository',
             function () use ($container): BroadcastingDocumentRepositoryDecorator {
-                $repository = new CacheDocumentRepository($container->get('event_jsonld_cache'));
+                $repository = new CacheDocumentRepository(
+                    $container->get('event_jsonld_cache'),
+                    $container->get('config')['cache']['milliseconds_between_retry'] ?? 0
```
JonasVHG, line 73: Maybe best to also make a PR with this key in `appconfig` & `udb3-vagrant`?

JonasVHG (approved):
Just a minor remark that doesn't break anything, for the rest it looks good, so I'll approve.

simon-debruijn (approved):

src/Doctrine/ReadModel/CacheDocumentRepository.php
```diff
@@ -35,17 +38,24 @@ public function fetch(string $id, bool $includeMetadata = false): JsonDocument
         return new JsonDocument($id, $value);
     }
 
-    public function save(JsonDocument $document): void
+    public function save(JsonDocument $document, int $attempts = 3): void
```
bertramakers, line 41: Maybe also make this config? You can still default it to 3 like you defaulted the milliseconds to 0. That way we can easily experiment with it. And there is no reason to have a different number of attempts depending on the caller of this function I think.

src/Doctrine/ReadModel/CacheDocumentRepository.php
```diff
@@ -35,17 +38,24 @@ public function fetch(string $id, bool $includeMetadata = false): JsonDocument
...
-            throw new RuntimeException('Could not save document ' . $document->getId() . 'to cache.');
+            throw new RuntimeException('Could not save document ' . $document->getId() . ' to cache.');
         }
 
         $savedDocument = $this->fetch($document->getId());
         if ($savedDocument->getRawBody() !== $document->getRawBody()) {
-            $this->logger->error('Saved document in cache does not match provided document ' . $document->getId());
+            $this->logger->error(
+                'Saved document in cache does not match provided document ' . $document->getId() . '. Retry attempts left: ' . $attempts
+            );
```
bertramakers, line 53: `error()` logs are also sent to Sentry. Maybe make this a `debug()` instead (which is only saved to a file), and add a new `error()` for when it is not saved even after the max numer of retries? Only in the latter case do we need to be alerted in Slack I think, so we can replay the event/place/organizer later to fix it.

LucWollants: I used log level `warning` when there are attempts left, because the save should never fail.

src/Doctrine/ReadModel/CacheDocumentRepository.php
LucWollants: The allowed retries are no injected in the constructor.

app/Event/EventJSONLDServiceProvider.php
LucWollants: For UDB3 Vagrant: https://github.com/cultuurnet/udb3-vagrant/pull/156
For appconfig: https://github.com/cultuurnet/appconfig/pull/324

simon-debruijn (approved):

src/Doctrine/ReadModel/CacheDocumentRepository.php
bertramakers: Good choice!

src/Doctrine/ReadModel/CacheDocumentRepository.php
```diff
@@ -48,7 +48,8 @@ public function save(JsonDocument $document, int $attempts = 3): void
 
         $savedDocument = $this->fetch($document->getId());
         if ($savedDocument->getRawBody() !== $document->getRawBody()) {
-            $this->logger->error(
+            $this->logger->log(
+                $attempts > 1 ? 'warning' : 'error',
```
bertramakers, line 58: Shouldn't it also be `warning` if there is still 1 attempt left? I'd expect an error if the last attempt fails

PR #1321, opened 2023-04-07
III-5529 Add `wfstatus` to granular events

JonasVHG:
### Added
- Added `getModeration()` to calculate `WorkflowStatus` from `CdbXml` in trait `EventFromUDB2`
- Added tests to check getModeration() in `EventFromUDB2`
- test `CdbXml`-files with various `wfstatus`

---
Ticket: https://jira.uitdatabank.be/browse/III-5529


src/Event/Events/EventFromUDB2.php
```diff
@@ -201,4 +215,28 @@ public function getExternalId(): ?string
...
+
+        if ($wfstatus === 'rejected') {
+            return new Rejected($this->eventId, new StringLiteral('Reason unknown (imported from UiTdatabank v2)'));
+        }
+
+        if ($wfstatus === 'deleted') {
+            return new EventDeleted($this->eventId);
+        }
+
+        return null;
```
LucWollants, line 240: Which `wfstatus` in XML can not be converted to a moderation event?

src/Event/Events/EventFromUDB2.php
```diff
@@ -201,4 +215,28 @@ public function getExternalId(): ?string
         $eventAsArray = $this->getEventAsArray();
         return $eventAsArray['location'][0]['label'][0]['@attributes']['externalid'] ?? null;
     }
+
+    private function getModeration(?string $wfstatus, ?string $lastUpdated): ?AbstractEvent
```
LucWollants, line 219: Can these parameters be nullable if an `isset` is used before calling it?

src/Event/Events/EventFromUDB2.php
JonasVHG: The `wfstatus` could still be `DRAFT` in `CdbXml`, which is not a moderation event in UDB3.

src/Event/Events/EventFromUDB2.php
JonasVHG: Indeed, fixed it.

src/Event/Events/EventFromUDB2.php
LucWollants: Thanks for clarifying, maybe add a small comment for this in the code?

src/Event/Events/EventFromUDB2.php
```diff
@@ -201,4 +215,28 @@ public function getExternalId(): ?string
         $eventAsArray = $this->getEventAsArray();
         return $eventAsArray['location'][0]['label'][0]['@attributes']['externalid'] ?? null;
     }
+
+    private function getModeration(string $wfstatus, string $lastUpdated): ?AbstractEvent
+    {
+        if ($wfstatus === 'readyforvalidation') {
+            return new Published(
+                $this->eventId,
+                \DateTimeImmutable::createFromFormat('Y-m-d\TH:i:s', $lastUpdated, new DateTimeZone('Europe/Brussels')),
```
LucWollants, line 224: Do we need this explicit hardcoded timezone?

src/Event/Events/EventFromUDB2.php
JonasVHG: Done.

LucWollants (approved):

PR #1322, opened 2023-04-07
III-5385 - As an admin I can give a client permissions on the events/places/organizers created by an user

simon-debruijn:
### Added

- Added support for configuring permissions/sapi3 constraint/labels on clients like on users, but via app config.

---
Ticket: https://jira.uitdatabank.be/browse/III-5385


tests/Role/ReadModel/Permissions/AppConfigUserPermissionsReadRepositoryTest.php
```diff
@@ -0,0 +1,76 @@
...
+            'jkfhsjkfsdhjk@clients' => [Permission::aanbodBewerken(), Permission::productiesAanmaken()],
+        ];
+
+        $this->repository = new AppConfigUserPermissionsReadRepository($userPermissionsReadRepository, $config);
+    }
+
+    /**
+     * @test
+     */
+    public function it_returns_all_enabled_permissions_for_client_id_when_executing_get_permissions()
```
JonasVHG, line 28: I would still add the return type, even if it is `void` (same for the rest of the file).

app/Configuration/ConfigurationServiceProvider.php
```diff
@@ -49,6 +50,26 @@ private function getConfiguration($container): array
     {
         $config = file_exists(__DIR__ . '/../../config.php') ? require __DIR__ . '/../../config.php' : [];
 
+
```
LucWollants, line 53: Nit: too many new lines

app/Labels/LabelServiceProvider.php
```diff
@@ -157,15 +158,18 @@ private function setUpReadModels(Container $container): void
...
                 $labels = file_exists(__DIR__ . '/../../config.excluded_labels.php') ? require __DIR__ . '/../../config.excluded_labels.php' : [];
 
-                return new GodUserReadRepositoryDecorator(
-                    new JsonReadRepository(
-                        $container->get('dbal_connection'),
-                        new StringLiteral(self::JSON_TABLE),
-                        new StringLiteral(self::LABEL_ROLES_TABLE),
-                        new StringLiteral(UserPermissionsServiceProvider::USER_ROLES_TABLE),
-                        new InMemoryExcludedLabelsRepository($labels ?? [])
+                return new AppConfigReadRepositoryDecorator(
```
LucWollants, line 161: Maybe make this name also more explicit as is done for the other repositories and make it clear this is specific for labels?

bertramakers: Simon and I just followed the naming pattern of every specific repository interface and its existing implementations. The one for Labels never contain the word "Labels", just like almost every other class/interface in that namespace. I feel like it would be better to keep it consistent for now, and add the name "Label" at a later point to all classes/interfaces inside that namespace?

app/Configuration/ConfigurationServiceProvider.php
bertramakers: https://github.com/cultuurnet/udb3-backend/pull/1322/commits/8d650629adb71452205eb1f0b71f2f71ad3a3d2f

src/Role/ReadModel/Permissions/AppConfigUserPermissionsReadRepository.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function getPermissions(string $userId): array
+    {
+        return $this->clientIdToPermissions[$userId] ??
+            $this->userPermissionsReadRepository->getPermissions($userId);
+    }
+
+    public function hasPermission(string $userId, Permission $permission): bool
+    {
+        return in_array($permission, $this->getPermissions($userId));
```
LucWollants, line 34: Should this also not delegate to `userPermissionsReadRepository` to check if it has the correct permission?

LucWollants: Ah, not needed because it uses `getPermissions`

LucWollants (approved):

JonasVHG (approved):

PR #1323, opened 2023-04-11
III-5544 polyfill mainimage

JonasVHG:
### Added

- Polyfill to remove `image` from `json` if no `mediaObjects` are present
- tests for new Polyfill

### Note
- This does not handle the edge case where an a legacy `mainImage` is removed but the `json` still contains other images.

---
Ticket: https://jira.uitdatabank.be/browse/III-5544


simon-debruijn (approved):

src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
```diff
@@ -264,6 +265,20 @@ function (array $json) {
         );
     }
 
+    private function fixMainImage(JsonDocument $jsonDocument): JsonDocument
```
LucWollants, line 268: I would make the name of the method a bit more explicit. I think the names of the test methods describe very well what happens. For example `removeMainImageWhenMediaObjectIsEmpty`

LucWollants (approved):

PR #1324, opened 2023-04-11
III-5512 polyfill ActorType

JonasVHG:
### Added

- Polyfill to remove `terms` with deprecated domain `actortype`.
- tests for new Polyfill

### Note
Because [III-5410](https://jira.uitdatabank.be/browse/III-5410) has to have run on production first, some changes were made:
-  `removeActorType()` is no longer called in `fetch()`
- now unused `removeActorType()` is ignored in `phpstan`
-  test `it_should_remove_actor_types()` is skipped.

---
Ticket: https://jira.uitdatabank.be/browse/III-5512


bertramakers (approved):
Looks good! You'll also need to add it in SAPI3 to remove it from the search results on-the-fly. Unless we reindex all these documents after this polyfill is merged & deployed, but that seems unlikely.

See https://github.com/cultuurnet/udb3-search-service/blob/4c58ebfc9335d7044136f34988e4a7d62550814f/src/ElasticSearch/JsonDocument/JsonLdPolyfillJsonTransformer.php

LucWollants (approved):

LucWollants (approved):

JonasVHG:
> Looks good! You'll also need to add it in SAPI3 to remove it from the search results on-the-fly. Unless we reindex all these documents after this polyfill is merged & deployed, but that seems unlikely.
> 
> See https://github.com/cultuurnet/udb3-search-service/blob/4c58ebfc9335d7044136f34988e4a7d62550814f/src/ElasticSearch/JsonDocument/JsonLdPolyfillJsonTransformer.php

Made a ticket for it: https://jira.uitdatabank.be/browse/III-5547

PR #1328, opened 2023-04-12
III-5359 Group duplicate code inside `GraphEditor`

LucWollants:
### Added
- Added `GraphEditor` with common code to edit graph:
  - `replaceLanguageValue`
  - `deleteLanguageValue`
  - `setGeneralProperties`

---
Ticket: https://jira.uitdatabank.be/browse/III-5359


app/RDF/GraphEditor.php
```diff
@@ -0,0 +1,141 @@
...
+        $this->graph = $graph;
+    }
+
+    public static function for(Graph $graph): self
+    {
+        return new self($graph);
+    }
+
+    public function setGeneralProperties(
+        string $uri,
```
bertramakers, line 43: Nit: Maybe using `$subject` or `$resource` instead of `$uri` might be a little bit clearer here, because the `$type` for example will also be a URI. And same in the other methods where for example `$property` will also contain a URI.

bertramakers (approved):

PR #1329, opened 2023-04-13
III-5537 Project workflow status on RDF projector of event

LucWollants:
### Added
- Implemented handlers inside the RDF projector of an event to set the workflow status
- Added common `WorkflowEditor` to handle the various workflow status

### Changed
- Moved `GraphEditor` to the `src` folder

---
Ticket: https://jira.uitdatabank.be/browse/III-5537


src/RDF/Editor/WorkflowEditor.php
```diff
@@ -0,0 +1,89 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\RDF\Editor;
+
+use EasyRdf\Graph;
+use EasyRdf\Literal;
+use EasyRdf\Resource;
+
+final class WorkflowEditor
```
bertramakers, line 11:
```suggestion
final class WorkflowStatusEditor
```
Small nit but this may be clearer, since "Workflow" could also be interpreted as something specific to RDF/Graphs maybe, while "WorkflowStatus" is a known concept

bertramakers (approved):
Nice cleanup of the duplication! 👌

bertramakers:
I'm going to merge this already so that we can get the changes from #1330 onto acc. My nit can optionally be implemented in a follow-up PR

src/RDF/Editor/WorkflowEditor.php
LucWollants: Done in this commit https://github.com/cultuurnet/udb3-backend/pull/1331/commits/8dee494e3be781f53aeb1c4b9b78c21a9700924d

PR #1334, opened 2023-04-14
III-5363 Handle `calendarUpdated` on event (single/multiple/periodic)

LucWollants:
### Added
- Added handler for `CalendarUpdated` on event for single calendar type.
- Added repo to store the `LocationId` when `LocationUpdated` is handled so it can be used when `CalendarUpdated` is handled.

### Note
- For the `periodic` calendar the opening hours are ignored.

---
Ticket: https://jira.uitdatabank.be/browse/III-5363

### DIFF
--- a/app/Event/EventRdfServiceProvider.php
+++ b/app/Event/EventRdfServiceProvider.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace CultuurNet\UDB3\Event;
 
 use CultuurNet\UDB3\Container\AbstractServiceProvider;
+use CultuurNet\UDB3\Event\ReadModel\RDF\CacheLocationIdRepository;
 use CultuurNet\UDB3\Event\ReadModel\RDF\RdfProjector;
 use CultuurNet\UDB3\RDF\MainLanguageRepository;
 use CultuurNet\UDB3\RDF\RdfServiceProvider;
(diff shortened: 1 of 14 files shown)


JonasVHG (commented):
This PR looks fine to me, but I think I would be best if someone with more knowledge of RDF approves it.

bertramakers (approved):
Looks good, just some points that can be done in follow-up PRs as well:

- If the calendar type is `single`, `multiple`, or `periodic`, there should be no `cpa:location` on the event (it should only be in `cp:ruimteTijd` as `cidoc:P161`)
- If the calendar type is `permanent`, there should be no `cp:ruimteTijd` (so it should be removed when switching to calendar type permanent), and the location should be set as `cpa:location` on the event
- Stan and I just discovered that the predicates `cidoc:E92`, `cidoc:P161` and `cidoc:P160` are not completely correct. They need some suffixes. Stan will update the example and the sheet and share the details with you

LucWollants:
Thanks for the extra feedback. I created two tickets for this:
- https://jira.uitdatabank.be/browse/III-5566: to make sure the `cpa:location` is added or removed
- https://jira.uitdatabank.be/browse/III-5563: to make sure the `cidoc:E92`, `cidoc:P161` and `cidoc:P160` are correct

JonasVHG (approved):

PR #1338, opened 2023-04-20
III-5566 Fix prov:atLocation

LucWollants:
### Changed
- Make sure `permanent` calendar has `prov:atLocation` but no `cp:ruimtetijd`
- Make sure `single`, `multiple` and `periodic` have `cp:ruimtetijd` but no `prov:atLocation`

---
Ticket: https://jira.uitdatabank.be/browse/III-5566

### DIFF
--- a/src/Event/ReadModel/RDF/RdfProjector.php
+++ b/src/Event/ReadModel/RDF/RdfProjector.php
@@ -205,11 +205,18 @@ final class RdfProjector implements EventListener
     private function handleLocationUpdated(LocationUpdated $event, string $uri, Graph $graph): void
     {
         $this->locationIdRepository->save($uri, $event->getLocationId());
+        $locationUri = $this->placesIriGenerator->iri($event->getLocationId()->toString());
 
         $resource = $graph->resource($uri);
 
-        $locationUri = $this->placesIriGenerator->iri($event->getLocationId()->toString());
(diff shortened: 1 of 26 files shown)


src/Event/ReadModel/RDF/RdfProjector.php
```diff
@@ -251,6 +255,27 @@ private function handleTypeUpdated(TypeUpdated $event, string $uri, Graph $graph
...
+        $resource = $graph->resource($uri);
+
+        $resource->set(self::PROPERTY_ACTVITEIT_LOCATIE, null);
+    }
+
+    private function addLocation(string $uri, Graph $graph): void
+    {
+        $locationId = $this->locationIdRepository->get($uri);
+
+        if ($locationId === null) {
```
JonasVHG, line 269: in `addLocationOnSpaceTimeResource()` there is no check on whether the `locationIdRepository` returns `null`.  Is this correct?

LucWollants: That is indeed not needed because `CalendarUpdated` happens after `LocationUpdated`.

But while thinking about this question I noticed that a scenario is missing. When an extra `LocationUpdated` would happen after the initial `CalendarUpdated` the location would be added to the root of the RDF projection because it does not take into account the calendar type.

I will add extra unit tests for this.

JonasVHG (approved):

PR #1345, opened 2023-05-02
III-5548 geopunt zipcode null

JonasVHG:
### Added

- `GeopuntAddressParser`: test for empty `Zipcode`
- `GeopuntAddressParser`: test for empty `Thoroughfarename`

### Changed

- `GeopuntAddressParser`: changed `jsonSchema`, make `Zipcode` & `Thoroughfarename`  nullable
- `GeopuntAddressParser`: expected result in test for incorrect property values
- `ParsedAddress`: make `thoroughfare` & `postalCode` nullable
- `RdfProjector`: add check if `thoroughfare` is not null.

---
Ticket: https://jira.uitdatabank.be/browse/III-5548

### DIFF
--- a/src/Address/GeopuntAddressParser.php
+++ b/src/Address/GeopuntAddressParser.php
@@ -69,7 +69,12 @@ final class GeopuntAddressParser implements AddressParser, LoggerAwareInterface
                         'type' => 'object',
                         'properties' => [
                             'Municipality' => ['type' => 'string'],
-                            'Zipcode' => ['type' => 'string'],
+                            'Zipcode' => [
+                                'anyOf' => [
+                                    ['type' => 'string'],
+                                    ['type' => 'null'],
+                                ],
+                            ],
                             'Thoroughfarename' => ['type' => 'string'],
                             'Housenumber' => [
                                 'anyOf' => [
--- a/src/Address/GeopuntAddressParser.php
+++ b/src/Address/GeopuntAddressParser.php
@@ -112,6 +112,11 @@ final class GeopuntAddressParser implements AddressParser, LoggerAwareInterface
         }
         $result = $data->LocationResult[0];
 
+        if ($result->Thoroughfarename === null) {
+            $this->logger->info('Response body did not contain a thoroughfarename inside "LocationResult" property. The address is not precise enough.');
+            return null;
+        }
+
         $this->logger->info('Successfully parsed response body.');
         return new ParsedAddress(
             $result->Thoroughfarename,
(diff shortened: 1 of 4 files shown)


src/Address/GeopuntAddressParser.php
```diff
@@ -102,6 +112,11 @@ public function parse(string $formattedAddress): ?ParsedAddress
         }
         $result = $data->LocationResult[0];
 
+        if ($result->Thoroughfarename === null) {
```
JonasVHG, line 115: To me, it seemed more logical to catch this here, than in the `RdfProjector`.

LucWollants: I would not make this method return `null` when `Thoroughfarename` is `null`. Maybe other parts of the code need another part, like `Zipcode`, but this no longer works now.

tests/Address/GeopuntAddressParserTest.php
```diff
@@ -316,6 +314,91 @@ public function it_can_handle_an_empty_LocationResult(): void
...
+        $expectedRequest = new Request(
+            'GET',
+            'https://loc.geopunt.be/v4/Location?q=Marguerite%20Durassquare,%201000%20Brussel,%20BE'
+        );
+
+        $mockResponseData = [
+            'LocationResult' => [
+                [
+                    'Municipality' => 'Bruxelles',
+                    'Zipcode' => null,
```
LucWollants, line 379: Maybe the `Zipcode` needs a value in this test?

src/Address/GeopuntAddressParser.php
JonasVHG: I've checked this in the code, and `getPostalCode()` & `getMunicipality()` are used to store it in the cache, but not used anywhere else. Only `getThoroughfare()` & `getHouseNumber()` are used. While I do not think it's likely that GeoPunt would return a `null` on `getThoroughfare()` and an actual value on `getHouseNumber()`, I would consider it incomplete data because a house number without a street isn't really useful information.

LucWollants: But the returned object is `ParsedAddress` with all fields `nullable`, what makes it correct or incorrect?

JonasVHG: Nothing, I doubted between those 2 approaches, but I assumed it would be better to not use a `ParsedAddress` because it is only used for `getThoroughfare()` & `getHouseNumber()`

LucWollants: I would keep it broad so the consumer can decide what to do with the result.

LucWollants (approved):

PR #1348, opened 2023-05-09
III-5545 Refactor `DummyLocationUpdated` to concrete event

LucWollants:
### Changed
- Refactored `DummyLocationUpdated` to concrete event

---
Ticket: https://jira.uitdatabank.be/browse/III-5545


LucWollants:
> Looks good, but we still have to decide how to handle XML's without street and/or housenr.

I added this to the ticket. The existing tests to get the `DummyLocation` also did not take this into account, so this refactor does not introduce a new issue.

PR #1352, opened 2023-05-17
III-5545 Handle missing street and/or number on dummy location

LucWollants:
### Changed
- Handled missing street and/or number on dummy location by converting it to `-`. This simple workaround avoids if statements when `Street` on `Address` would have been made `nullable` for this exceptional case on dummy locations.

---
Ticket: https://jira.uitdatabank.be/browse/III-5545

### DIFF
--- a/tests/Event/samples/event_with_dummy_location.cdbxml.xml
+++ b/tests/Event/samples/event_with_dummy_location.cdbxml.xml
@@ -29,7 +29,6 @@
             <cdb:title>'Bekend met Gent' - quiz</cdb:title>
         </cdb:eventdetail>
     </cdb:eventdetails>
-    <cdb:keywords>gent;Quiz;;Gent on Files</cdb:keywords>
     <cdb:location>
         <cdb:address>
             <cdb:physical>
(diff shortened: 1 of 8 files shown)


tests/Event/samples/event_with_dummy_location.cdbxml.xml
```diff
@@ -0,0 +1,52 @@
...
+    </cdb:categories>
+    <cdb:comments/>
+    <cdb:eventdetails>
+        <cdb:eventdetail lang="nl">
+            <cdb:calendarsummary>vrij 21/11/14 om 20:00 </cdb:calendarsummary>
+            <cdb:shortdescription>Korte beschrijving</cdb:shortdescription>
+            <cdb:title>'Bekend met Gent' - quiz</cdb:title>
+        </cdb:eventdetail>
+    </cdb:eventdetails>
+    <cdb:keywords>gent;Quiz;;Gent on Files</cdb:keywords>
```
simon-debruijn, line 32: Are the 2 `;;` intentional?

simon-debruijn (approved):

tests/Event/samples/event_with_dummy_location.cdbxml.xml
JonasVHG: They are not invalid according to the `XSD`, but it might not yield the expected results, I assume they were copied from other tests.

JonasVHG (approved):

tests/Event/samples/event_with_dummy_location.cdbxml.xml
LucWollants: It was indeed a copy from other tests. I will just remove the keywords.

PR #1362, opened 2023-05-26
III-5613 Refactor `Label\DBALReadRepository`

JonasVHG:
### Added

- DBALReadRepository: `Test` if it can omit excluded `Labels`

### Changed

- `Query`: split up `$suggestion` between `$excludeExcludedLabels` & `$excludeInvalidLabels`
- `Query`: split up `isSuggestion` member between `isExcludedExcluded ` & `isInvalidExcluded`
- `QueryFactory`: `suggestion` is translated to both `excludeExcludedLabels` & `excludeInvalidLabels`
- `Label\DBALReadRepository`: `ExcludedLabelsRepository` removed from `__construct`
-  `Label\DBALReadRepository`: excluded labels work via database Column instead of from `config`
- `Label\SchemaConfigurator`: Add `EXCLUDED_COLUMN `

### Fixed

- Code cleanup

---

### DIFF
--- a/app/Console/Command/AbstractRemoveLabel.php
+++ b/app/Console/Command/AbstractRemoveLabel.php
@@ -6,7 +6,6 @@ namespace CultuurNet\UDB3\Console\Command;
 
 use Broadway\CommandHandling\CommandBus;
 use CultuurNet\UDB3\Label\ReadModels\JSON\Repository\Doctrine\DBALReadRepository;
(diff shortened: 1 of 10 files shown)


LucWollants:
We first discussed taking these steps, but I now notice that we have a breaking change when we commit this PR. So we first need to add the command to mark a label as excluded. Once that is done we can mark labels as excluded on all environments and remove the old repository.

JonasVHG:
This will be fixed in a different PR -> https://github.com/cultuurnet/udb3-backend/pull/1366

PR #1363, opened 2023-05-26
III-5614 Inject broadcasting repo in json projector

LucWollants:
### Changed
- Injected broadcasting repo into JSON projector to avoid the overhead of decorators used for polyfill. These decorators are not needed when updating JSON projections for places and events.

---
Ticket: https://jira.uitdatabank.be/browse/III-5614


simon-debruijn (approved):
The code looks good, It looks like a performance improvement. The functionality should remain the same?

LucWollants:
> The code looks good, It looks like a performance improvement. The functionality should remain the same?

I tried to verify if "everything" was still working. For the pure Redis updates broadcasting is essential. For HTTP requests the entire decorated repos are still used.

I did also run all the acceptance tests.

PR #1366, opened 2023-05-31
III-5613 exclude include label commands

JonasVHG:
### Fixed

---

### DIFF
--- a/src/Label/Commands/ExcludeLabel.php
+++ b/src/Label/Commands/ExcludeLabel.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Label\Commands;
+
+final class ExcludeLabel extends AbstractCommand
+{
+}
--- a/src/Label/Events/Excluded.php
+++ b/src/Label/Events/Excluded.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Label\Events;
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+
+final class Excluded extends AbstractEvent
+{
+    public static function deserialize(array $data): Excluded
+    {
+        return new self(
+            new UUID($data[self::UUID]),
+            $data[self::NAME]
+        );
+    }
+}
--- a/src/Label/Label.php
+++ b/src/Label/Label.php
@@ -25,6 +25,8 @@ class Label extends EventSourcedAggregateRoot
 
     private Privacy $privacy;
 
+    private bool $excluded;
+
     public function getAggregateRootId(): string
     {
         return $this->uuid->toString();
@@ -34,7 +36,8 @@ class Label extends EventSourcedAggregateRoot
         UUID $uuid,
         string $name,
         Visibility $visibility,
-        Privacy $privacy
+        Privacy $privacy,
+        bool $excluded = false
     ): Label {
         $label = new Label();
 
@@ -42,7 +45,8 @@ class Label extends EventSourcedAggregateRoot
             $uuid,
             $name,
             $visibility,
-            $privacy
+            $privacy,
+            $excluded
         ));
 
         return $label;
@@ -102,6 +106,7 @@ class Label extends EventSourcedAggregateRoot
         $this->name = $created->getName();
         $this->visibility = $created->getVisibility();
         $this->privacy = $created->getPrivacy();
+        $this->excluded = $created->isExcluded();
     }
 
     public function applyCopyCreated(CopyCreated $copyCreated): void
--- a/src/Label/Label.php
+++ b/src/Label/Label.php
@@ -7,6 +7,8 @@ namespace CultuurNet\UDB3\Label;
 use Broadway\EventSourcing\EventSourcedAggregateRoot;
 use CultuurNet\UDB3\Label\Events\CopyCreated;
 use CultuurNet\UDB3\Label\Events\Created;
+use CultuurNet\UDB3\Label\Events\Excluded;
+use CultuurNet\UDB3\Label\Events\Included;
 use CultuurNet\UDB3\Label\Events\MadeInvisible;
 use CultuurNet\UDB3\Label\Events\MadePrivate;
 use CultuurNet\UDB3\Label\Events\MadePublic;
@@ -97,6 +99,20 @@ class Label extends EventSourcedAggregateRoot
     {
         if (!$this->privacy->sameAs(Privacy::PRIVACY_PRIVATE())) {
             $this->apply(new MadePrivate($this->uuid, $this->name));
+        }
+    }
+
+    public function include(): void
+    {
+        if ($this->excluded) {
+            $this->apply(new Included($this->uuid, $this->name));
+        }
+    }
+
+    public function exclude(): void
+    {
+        if ($this->excluded === false) {
+            $this->apply(new Excluded($this->uuid, $this->name));
         }
     }
 
--- a/src/Label/Label.php
+++ b/src/Label/Label.php
@@ -149,4 +149,14 @@ class Label extends EventSourcedAggregateRoot
     {
         $this->privacy = Privacy::PRIVACY_PRIVATE();
     }
+
+    public function applyIncluded(): void
+    {
+        $this->excluded = false;
+    }
+
+    public function applyExcluded(): void
+    {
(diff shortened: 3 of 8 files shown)


src/Label/Commands/ExcludeLabel.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Label\Commands;
+
+final class ExcludeLabel extends AbstractCommand
```
LucWollants, line 7: Other commands don't have the `Label` suffix. So `Exclude` could be enough?

src/Label/Events/CopyCreated.php
```diff
@@ -19,9 +19,10 @@ public function __construct(
         string $name,
         Visibility $visibility,
         Privacy $privacy,
-        UUID $parentUuid
+        UUID $parentUuid,
+        bool $excluded = false
```
LucWollants, line 23: I really wonder if we need that parameter here. What is the purpose of creating an excluded label from the start, it should never be created.

src/Label/Events/CopyCreated.php
```diff
@@ -38,7 +39,8 @@ public static function deserialize(array $data): CopyCreated
             $data[self::NAME],
             new Visibility($data[self::VISIBILITY]),
             new Privacy($data[self::PRIVACY]),
-            new UUID($data[self::PARENT_UUID])
+            new UUID($data[self::PARENT_UUID]),
+            (bool) $data[self::EXCLUDED]
```
LucWollants, line 43: I think this will result in a crash when trying to modify old labels which were not created with an excluded field. This is a breaking change.

src/Label/Events/Excluded.php
```diff
@@ -0,0 +1,18 @@
...
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
+
+final class Excluded extends AbstractEvent
+{
+    public static function deserialize(array $data): Excluded
+    {
+        return new self(
+            new UUID($data[self::UUID]),
+            $data[self::NAME]
```
LucWollants, line 15: I don't think we don't need to store the name of the label. This is a bit of the old approach that tries to reuse a common base class. I would just create a very simple class with no inheritance that only stores the UUID of the label.

src/Label/Label.php
```diff
@@ -34,15 +38,17 @@ public static function create(
         UUID $uuid,
         string $name,
         Visibility $visibility,
-        Privacy $privacy
+        Privacy $privacy,
+        bool $excluded = false
```
LucWollants, line 42: I don't think there is a need for this.

src/Label/ReadModels/JSON/Projector.php
```diff
@@ -87,6 +91,16 @@ public function applyMadePrivate(MadePrivate $madePrivate): void
...
     }
 
+    public function applyLabelIncluded(Included $included): void
+    {
+        $this->writeRepository->updateIncluded($included->getUuid());
+    }
+
+    public function applyLabelExcluded(Excluded $excluded): void
+    {
+        $this->writeRepository->updateCountDecrement($excluded->getUuid());
```
LucWollants, line 101: This doesn't look like the correct call `updateCountDecrement`

src/Label/ReadModels/JSON/Repository/BroadcastingWriteRepositoryDecorator.php
```diff
@@ -15,97 +15,83 @@
...
 
-    /**
-     * @inheritDoc
-     */
     public function save(
         UUID $uuid,
         StringLiteral $name,
         Visibility $visibility,
         Privacy $privacy,
+        bool $excluded,
```
LucWollants, line 35: There is no need to pass this parameter, a new label is never excluded.

LucWollants (changes requested):
I think the new `excluded` property should only be exposed on the JSON projection and added as a new command. No need to also expose it on the existing `Created` and `CopyCreated` events. This will even lead to a breaking change.

Note: I did review the unit tests because they will also need a change, after the cleanup I will review the tests.

src/Label/Commands/ExcludeLabel.php
JonasVHG: Problem is that `Include` is a reserved keyword in PHP, so you cannot have a class `Include` `Exclude` is a valid classname but that would make it inconsitent with `Include`.

app/Console/Command/BulkExcludeLabel.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    public function configure(): void
+    {
+        $this->setName('label:bulk-exclude');
+        $this->setDescription('Excludes all labels in a config file');
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $labelIds = file_exists(__DIR__ . '/../../../config.excluded_labels.php') ? require __DIR__ . '/../../../config.excluded_labels.php' : [];
```
LucWollants, line 22: Maybe a warning message is better here?

app/Console/Command/BulkExcludeLabel.php
```diff
@@ -0,0 +1,53 @@
...
+
+        if (!$this->askConfirmation($input, $output, count($labelIds))) {
+            return 0;
+        }
+
+        foreach ($labelIds as $labelId) {
+            $this->commandBus->dispatch(
+                new \CultuurNet\UDB3\Label\Commands\ExcludeLabel(new UUID($labelId))
```
LucWollants, line 30: Make this the FQN

app/Console/Command/BulkExcludeLabel.php
```diff
@@ -0,0 +1,53 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+final class BulkExcludeLabel extends AbstractCommand
+{
+    public function configure(): void
+    {
+        $this->setName('label:bulk-exclude');
+        $this->setDescription('Excludes all labels in a config file');
```
LucWollants, line 17: It's a specific config file.

app/Console/Command/ExcludeLabel.php
```diff
@@ -0,0 +1,31 @@
...
+        $this->setDescription('Excludes a label with the given uuid');
+        $this->addArgument('labelId', InputArgument::REQUIRED, 'Uuid of the label');
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $labelId = $input->getArgument('labelId');
+
+        $this->commandBus->dispatch(
+            new \CultuurNet\UDB3\Label\Commands\ExcludeLabel(new UUID($labelId))
```
LucWollants, line 26: FQN

app/Console/Command/IncludeLabel.php
```diff
@@ -0,0 +1,31 @@
...
+        $this->setDescription('Includes a label with the given uuid');
+        $this->addArgument('labelId', InputArgument::REQUIRED, 'Uuid of the label');
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $labelId = $input->getArgument('labelId');
+
+        $this->commandBus->dispatch(
+            new \CultuurNet\UDB3\Label\Commands\IncludeLabel(new UUID($labelId))
```
LucWollants, line 26: FQN

tests/Label/LabelTest.php
```diff
@@ -45,7 +47,7 @@ public function setUp(): void
             $this->uuid,
             $this->name,
             $this->visibility,
-            $this->privacy
+            $this->privacy,
```
LucWollants, line 50: Remove trailing comma

tests/Label/LabelTest.php
```diff
@@ -254,4 +256,106 @@ public function it_does_not_make_a_label_private_when_already_private(): void
...
+
+    /**
+     * @test
+     */
+    public function it_can_exclude_an_included_label(): void
+    {
+        $this->scenario
+            ->withAggregateId($this->uuid->toString())
+            ->given([$this->created])
+            ->when(function ($label) {
```
LucWollants, line 268: Maybe type-hint the parameter of the anonymous function? I also think that a fat arrow method is possible.

LucWollants (approved):
Only a few remarks left, so I already approved.

PR #1374, opened 2023-06-05
III-5613 Create CLI Command to invalid Labels

JonasVHG:
### Added

- Added CLI Command to Exclude Malformed Labels based on the Regex `\'^[a-zA-Z\d_\-]{2,50}$\'`

### Changed

- Register new Command `label:exclude-invalid` in `ConsoleServiceProvvider`

---

### DIFF
--- a/app/Console/Command/ExcludeInvalidLabels.php
+++ b/app/Console/Command/ExcludeInvalidLabels.php
@@ -72,7 +72,7 @@ final class ExcludeInvalidLabels extends AbstractCommand
 
     private function getAllLabelsCount(): int
     {
-        return $this->connection->createQueryBuilder()->select('*')
+        return $this->connection->createQueryBuilder()->select('uuid_col')
             ->from('labels_json')
             ->where('excluded = :excluded')
             ->setParameter(':excluded', 0)


app/Console/Command/ExcludeMalformedLabel.php
```diff
@@ -0,0 +1,92 @@
...
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+final class ExcludeMalformedLabel extends AbstractCommand
+{
+    private const LABEL_REGEX = '/^[a-zA-Z\d_\-]{2,50}$/';
+
+    private const MAX_RESULTS = 1;
```
LucWollants, line 21: Maybe tweak this value for better performance? There is no need here to make a SQL call for every single label.

app/Console/Command/ExcludeMalformedLabel.php
```diff
@@ -0,0 +1,92 @@
...
+        $offset = 0;
+        do {
+            $labels = $this->getAllLabels($offset);
+
+            foreach ($labels as $label) {
+                $labelId = new Uuid($label['uuid_col']);
+                $labelName = $label['name'];
+                $excluded = (bool) $label['excluded'];
+
+                if (!$excluded && !preg_match(self::LABEL_REGEX, $labelName)) {
```
LucWollants, line 57: Exclude the already excluded labels? So update the methods `getAllLabelsCount` and `getAllLabels` to not take into account excluded labels.

app/Console/Command/ExcludeMalformedLabel.php
```diff
@@ -0,0 +1,92 @@
...
+                    $this->commandBus->dispatch(
+                        new ExcludeLabelCommand($labelId)
+                    );
+                }
+                $progressBar->advance();
+            }
+
+            $offset += count($labels);
+        } while ($offset < $labelsCount);
```
LucWollants, line 66: I would stop the loop if no labels are returned or the count of labels is lower than the max number of labels. Otherwise, this could lead to an infinite loop.

This also makes me wonder, we are working on the JSON projection and not the event store. Maybe a replay of labels is also needed?

app/Console/ConsoleServiceProvider.php
```diff
@@ -76,6 +77,7 @@ final class ConsoleServiceProvider extends AbstractServiceProvider
         'console.organizer:change-owner-bulk',
         'console.label:bulk-exclude',
         'console.label:exclude',
+        'console.label:exclude-malformed',
```
LucWollants, line 80: Maybe a better wording is `exclude-invalid` because they no longer match the required validation rules.

app/Console/Command/ExcludeInvalidLabel.php
```diff
@@ -14,7 +14,7 @@
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
-final class ExcludeMalformedLabel extends AbstractCommand
+final class ExcludeInvalidLabel extends AbstractCommand
```
LucWollants, line 17: ...and it is also plural?

PR #1377, opened 2023-06-06
Cleanup AMQP registration code

LucWollants:
### Changed
- Replaced usages of `StringLiteral` with `string`
- Used local variables where possible
- Removed unused local variables

### Note
- There is only one big commit but the review should be straightforward file by file

---
Ticket: https://jira.uitdatabank.be/browse/III-5630

### DIFF
--- a/src/Broadway/AMQP/EventBusForwardingConsumer.php
+++ b/src/Broadway/AMQP/EventBusForwardingConsumer.php
@@ -12,17 +12,10 @@ use Broadway\EventHandling\EventBus;
 use CultuurNet\UDB3\Deserializer\DeserializerLocatorInterface;
 use PhpAmqpLib\Connection\AMQPStreamConnection;
 use Ramsey\Uuid\UuidFactoryInterface;
-use CultuurNet\UDB3\StringLiteral;
 
-/**
- * Forwards messages coming in via AMQP to an event bus.
- */
 class EventBusForwardingConsumer extends AbstractConsumer
 {
-    /**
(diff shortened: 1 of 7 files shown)


src/Broadway/AMQP/EventBusForwardingConsumer.php
```diff
@@ -12,27 +12,20 @@
 use CultuurNet\UDB3\Deserializer\DeserializerLocatorInterface;
 use PhpAmqpLib\Connection\AMQPStreamConnection;
 use Ramsey\Uuid\UuidFactoryInterface;
-use CultuurNet\UDB3\StringLiteral;
 
-/**
- * Forwards messages coming in via AMQP to an event bus.
- */
 class EventBusForwardingConsumer extends AbstractConsumer
```
JonasVHG, line 16:
```suggestion
final class EventBusForwardingConsumer extends AbstractConsumer
```

tests/Broadway/AMQP/CommandBusForwardingConsumerTest.php
```diff
@@ -17,26 +17,6 @@
 
 class CommandBusForwardingConsumerTest extends TestCase
```
JonasVHG, line 18:
```suggestion
final class CommandBusForwardingConsumerTest extends TestCase
```

JonasVHG (approved):
There are some classes which could be made final, but the rest looks good, so I'll approve.

PR #1382, opened 2023-06-07
III-5648 Setup Behat

LucWollants:
### Added
- Added Behat

---
Ticket: https://jira.uitdatabank.be/browse/III-5648


JonasVHG (approved):
2 remarks:
* I think it would be best to update the `readme.md` and `docker.md` in a separate PR(maybe wait till the Behat-integration is more mature).
* The current setup would not work in a new docker installation. Since this was a quick fix, I've made a separate [PR](https://github.com/cultuurnet/udb3-backend/pull/1384) for it.

Apart from that, it looks great 👍 .
Seeing the current Code, make me think migration might not be as hard as I imagined.
Since my remarks should be done in different PR's I'll approve.

PR #1384, opened 2023-06-09
move config.features.php to features folder

JonasVHG:
### Fixed

- Let `make config` move `config.features.php` to correct folder

---

### DIFF
--- a/docker/config.sh
+++ b/docker/config.sh
@@ -6,6 +6,7 @@ if [ -d "$DIR" ]; then
   cp "$DIR"/* .
   # needed because it is hidden
   cp "$DIR"/.env .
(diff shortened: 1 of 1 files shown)


docker/config.sh
```diff
@@ -6,6 +6,7 @@ if [ -d "$DIR" ]; then
   cp "$DIR"/* .
   # needed because it is hidden
   cp "$DIR"/.env .
+  mv ./config.features.php features/
```
LucWollants, line 9: Is this needed? Does the `cp` not take care of all files and subfolders?

JonasVHG: My mistake, I didn't notice the features folder in `appconfig`

PR #1417, opened 2023-06-29
III-5683 case insensitive label check

JonasVHG:
### Changed

- `AppConfigReadRepositoryDecorator`: make `canUseLabel()` case-insensitive

---
Ticket: https://jira.uitdatabank.be/browse/IIII-5683

### DIFF
--- a/src/Label/ReadModels/JSON/Repository/AppConfigReadRepositoryDecorator.php
+++ b/src/Label/ReadModels/JSON/Repository/AppConfigReadRepositoryDecorator.php
@@ -48,7 +48,7 @@ final class AppConfigReadRepositoryDecorator implements ReadRepositoryInterface
         }
 
         $labels = $config['labels'] ?? [];
-        return in_array($name, $labels);
+        return in_array(strtolower($name), array_map('strtolower', $labels));
     }
 
     public function search(Query $query): array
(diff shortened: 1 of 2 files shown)


tests/Label/ReadModels/JSON/Repository/AppConfigReadRepositoryDecoratorTest.php
```diff
@@ -130,4 +130,21 @@ public function it_returns_true_if_the_decorator_returns_false_but_the_client_ha
         $canUseLabel = $this->appConfigReadRepositoryDecorator->canUseLabel($userId, $label);
         $this->assertTrue($canUseLabel);
     }
+
+    /**
+     * @test
+     */
+    public function it_returns_true_if_the_decorator_returns_false_but_the_client_has_the_label_in_a_different_case_in_the_config(): void
```
LucWollants, line 137: It's the `decoratee` which returns false. The `decoratee` is the class that gets a slightly different implementation by adding decoration around it with a `decorator` class.

But I guess it's fine to make the name of this method more compact and just make the case-insensitive part clear.

src/Label/ReadModels/JSON/Repository/AppConfigReadRepositoryDecorator.php
```diff
@@ -48,7 +48,7 @@ public function canUseLabel(string $userId, string $name): bool
         }
 
         $labels = $config['labels'] ?? [];
-        return in_array($name, $labels);
+        return in_array(strtolower($name), array_map('strtolower', $labels));
```
LucWollants, line 51: I think it's fine to convert the input array to lowercase for now. But in the future, this can become a waste of CPU power. Instead, make sure that the labels inside the config file are always lowercase.

LucWollants (approved):

PR #1421, opened 2023-07-26
III-5749 Add labels data to behat tests and fix failing test

JonasVHG:
### Added

- Added Labels Data setup on `LabelSteps`
- Added `getLabel()` on `LabelSteps`
- Add step  `When I patch the label with :id and :command` on `LabelSteps`

### Changed

- Make sure LabelData is created when running tests for `labels/get`

Related PR: https://github.com/cultuurnet/udb3-acceptance-tests/pull/193

---
Ticket: https://jira.uitdatabank.be/browse/III-5749

### DIFF
--- a/features/Steps/LabelSteps.php
+++ b/features/Steps/LabelSteps.php
@@ -22,9 +22,9 @@ trait LabelSteps
     }
 
     /**
-     * @When I patch the label :id with :command
+     * @When I patch the label with id :id and command :command
      */
-    public function iPatchTheLabelWith(string $id, string $command): void
+    public function iPatchTheLabelWithIdAndCommand(string $id, string $command): void
     {
         $response = $this->getHttpClient()->patchJSON(
             '/labels/' . $id,
--- a/features/Steps/LabelSteps.php
+++ b/features/Steps/LabelSteps.php
@@ -35,6 +35,22 @@ trait LabelSteps
     }
 
     /**
+     * @Given I patch the label :labelId with command :command
+     */
+    public function patchLabel(string $uuid, string $command): void
+    {
+        $response = $this->getHttpClient()->patchJSON(
+            '/labels/' . $uuid,
+            $this->variableState->replaceVariables(
+                Json::encode([
+                    'command' => $command,
+                ])
+            )
+        );
+        $this->responseState->setResponse($response);
+    }
+
+    /**
      * @Given labels test data is available
      */
     public function labelsTestDataIsAvailable(): void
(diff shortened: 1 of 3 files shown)


features/Steps/LabelSteps.php
```diff
@@ -34,6 +34,58 @@ public function iCreateAnInvisibleLabelWithARandomNameOfCharacters(int $nrOfChar
         );
     }
 
+    /**
+     * @Given labels test data is available
+     */
+    public function labelsTestDataIsAvailable(): void
+    {
+        // Create "public-visible" if it doesn't exist yet and (re)set the right privacy and visibility in case its needed
+        $this->createLabel('public-visible', true, true);
```
LucWollants, line 43: I would first check if the label exists and only create it when it does not exist. When it already exists there is no need to create it again, only making sure to set the correct privacy and visibility.

features/label/get.feature
```diff
@@ -4,6 +4,7 @@ Feature: Test the UDB3 labels API
     Given I am using the UDB3 base URL
     And I am using an UiTID v1 API key of consumer "uitdatabank"
     And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And labels test data is available
```
LucWollants, line 7: I would create the labels inside the `beforeSuite` method here https://github.com/cultuurnet/udb3-backend/blob/master/features/Bootstrap/FeatureContext.php#L72

LucWollants (changes requested):
Already some nice progress! I would only tweak it to become more robust and performant.

features/Steps/LabelSteps.php
```diff
@@ -40,48 +40,66 @@ public function iCreateAnInvisibleLabelWithARandomNameOfCharacters(int $nrOfChar
     public function labelsTestDataIsAvailable(): void
     {
         // Create "public-visible" if it doesn't exist yet and (re)set the right privacy and visibility in case its needed
-        $this->createLabel('public-visible', true, true);
         $this->getLabel('public-visible');
+        if ($this->responseState->getStatusCode() === 404) {
+            $this->createLabel('public-visible', true, true);
```
LucWollants, line 45: If the label needs to be created there is no need to patch the privacy and visibility.

JonasVHG: Nice find.

features/Steps/LabelSteps.php
```diff
@@ -50,4 +121,27 @@ private function createLabel(string $name, bool $visible, bool $public): void
...
+    {
+        $response = $this->getHttpClient()->patchJSON(
+            '/labels/' . $uuid,
+            $this->variableState->replaceVariables(
+                Json::encode([
+                    'command' => $command,
+                ])
+            )
+        );
+        $this->responseState->setResponse($response);
```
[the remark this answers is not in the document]
LucWollants: Maybe even better is to change the scenarios that do the patch on a label with this reusable method. Something like `And I patch the label <labelId> with command <command>`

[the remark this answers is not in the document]
JonasVHG: Done, while I am at, I might be useful to also make a step `I get the label <labelId>`?

LucWollants (approved):

PR #1425, opened 2023-07-26
added xdebug to docker

grubolsch:
### Added
Added xdebug to the docker setup

I configured xdebug to only run when enable with a trigger, meaning:
- Running the script in ./bin/start-xdebug.sh for CLI scripts
- Adding XDEBUG_TRIGGER as a GET or POST variable when using Postman

There is a slight performance degradation because of using xdebug, so I am not 100% sure we should include it, but I want to open it for the team. I myself use xdebug a lot, but I can just keep it on my own machine.

Performance tests without xdebug, running "time make ci" on my machine:

make ci  0.10s user 0.23s system 2% cpu 11.059 total
make ci  0.10s user 0.22s system 2% cpu 11.102 total
make ci  0.10s user 0.23s system 2% cpu 11.561 total
make ci  0.10s user 0.22s system 2% cpu 11.572 total

With xdebug:
make ci  0.10s user 0.24s system 2% cpu 13.702 total
make ci  0.11s user 0.26s system 2% cpu 13.497 total
make ci  0.10s user 0.25s system 2% cpu 13.430 total
make ci  0.11s user 0.25s system 2% cpu 13.733 total

You see it adds around 2 seconds of time, even with xdebug trigger off.

Please let me know your opinion, happy to discuss!

### DIFF
--- a/bin/start-xdebug.sh
+++ b/bin/start-xdebug.sh
@@ -1,0 +1,1 @@
+export XDEBUG_MODE=debug XDEBUG_SESSION=1 PHP_IDE_CONFIG="serverName=host.docker.internal"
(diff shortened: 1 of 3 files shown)


LucWollants:
As discussed through Slack, investigate https://docs.docker.com/compose/compose-file/15-profiles/ to have a "normal" container and a container with xdebug.

PR #1432, opened 2023-08-01
III-5751 make invalid label excluded by default

JonasVHG:
### Added

- Unit test for creating invalid labels.
- Acceptance test to create invalid labels.
- `LabelSteps`: Added step to create Label with know `name`

### Changed

- Make a Label that does not confirm to `LabelName::REGEX_SUGGESTIONS` immediately invalid when it is created.
- `CreateLabelRequestHandler`: Post a second `Command` to exclude if `LabelName` is invalid.

### Removed

- `LabelSteps`: Do not exclude `special_label#` & `special_label*` when they need to be created 

---
Ticket: https://jira.uitdatabank.be/browse/III-5751

### DIFF
--- a/features/Steps/LabelSteps.php
+++ b/features/Steps/LabelSteps.php
@@ -22,6 +22,23 @@ trait LabelSteps
     }
 
     /**
+     * @When I patch the label with id :id and command :command
+     */
+    public function iPatchTheLabelWithIdAndCommand(string $id, string $command): void
+    {
+        $response = $this->getHttpClient()->patchJSON(
+            '/labels/' . $id,
+            Json::encode([
+                'command' => $command,
+            ])
+        );
+        $this->responseState->setResponse($response);
+
+        $this->theResponseStatusShouldBe(204);
+    }
+
+
+    /**
      * @Given I create an invisible label with a random name of :nrOfCharacters characters
      */
     public function iCreateAnInvisibleLabelWithARandomNameOfCharacters(int $nrOfCharacters): void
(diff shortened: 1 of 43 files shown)


LucWollants (changes requested):
Is it possible to do this inside the `Label` aggregate? I think it can be possible to call an `apply` with `Excluded` when the name does not match the regex.

features/Steps/LabelSteps.php
```diff
@@ -37,6 +37,18 @@ public function iPatchTheLabelWithIdAndCommand(string $id, string $command): voi
         $this->theResponseStatusShouldBe(204);
     }
 
+    /**
+     * @When I create a label with name :name
+     */
+    public function iCreateALabelWithName(string $name): void
+    {
+        var_dump($this->variableState->replaceVariables($name));
```
LucWollants, line 45: Can be removed.

features/Steps/LabelSteps.php
```diff
@@ -51,6 +63,8 @@ public function iCreateAnInvisibleLabelWithARandomNameOfCharacters(int $nrOfChar
         );
     }
 
+    /**
```
LucWollants, line 66: Can be removed

tests/Http/Label/CreateLabelRequestHandlerTest.php
```diff
@@ -66,4 +66,37 @@ public function it_can_create_a_label(): void
             $this->commandBus->getRecordedCommands()
         );
     }
+
+    /**
+     * @test
+     */
+    public function it_will_exclude_an_invalid_label(): void
```
LucWollants, line 73: Can be removed

LucWollants (approved):

PR #1433, opened 2023-08-01
Fix acceptance test: Create a news article with url that should have been encoded

JonasVHG:
### Changed

- Use `café` instead of `caf%C3%A9` like in original test: https://github.com/cultuurnet/udb3-acceptance-tests/blob/b289e4bb43084603e34958979a6144e24fd89bca/features/curator/create.feature#L94
- Do not escape `%` with `%%` because `sprintf()` of `Ruby` is no longer used.

### Fixed

- `Create a news article with url that should have been encoded` is green

---

### DIFF
--- a/features/curator/create.feature
+++ b/features/curator/create.feature
@@ -113,7 +113,7 @@ Feature: Test the curator API
 
     When I send a GET request to "/news-articles/%{articleId}"
     Then the response status should be "200"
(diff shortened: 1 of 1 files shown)


LucWollants:
I think `Update a news article with an url that should have been encoded` has the same issue?

LucWollants (approved):

PR #1439, opened 2023-08-08
Refactor Sapi3SearchService

grubolsch:
### Added
Unit test for sapi3SearchServiceFactory.

### Changed
Made concrete classes of each configuration of the sapi3 search service (discussed with Luc). This allows the use of fully qualified names instead of usage of strings in the container to find the correct service.

---
Ticket: https://jira.uitdatabank.be/browse/III-5758


simon-debruijn (approved):
Nice improvement 🚀 

You only still have to fix the static analysis errors.

PR #1441, opened 2023-08-08
III-5763 remove bookinginfo when given empty values

grubolsch:
### Changed
When one created an offer without bookingInfo, it does not contain a bookingInfo Node in the Json.
But when you would remove it (empty put) it stayed as an empty array.
Changed so the array is removed when empty from the projector.

---
Ticket: https://jira.uitdatabank.be/browse/III-5763

### DIFF
--- a/features/event/booking-info.feature
+++ b/features/event/booking-info.feature
@@ -254,3 +254,16 @@ Feature: Test the UDB3 events API
       ]
     }
     """
+
+  Scenario: Delete event booking info (with put)
+    Given I create an event from "events/event-with-bookinginfo.json" and save the "url" as "eventUrl"
+    When I set the JSON request payload to:
+    """
+    {
+      "bookingInfo": {}
+    }
+    """
+    And I send a PUT request to "%{eventUrl}/booking-info"
+    Then the response status should be "204"
+    When I get the event at "%{eventUrl}"
+    And the JSON response should not have "bookingInfo"
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -631,7 +631,11 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
 
         $offerLd = $document->getBody();
 
-        $offerLd->bookingInfo = $bookingInfoUpdated->getBookingInfo()->toJsonLd();
+        if (empty($bookingInfoUpdated->getBookingInfo()->toJsonLd())) {
+            unset($offerLd->bookingInfo);
+        } else {
+            $offerLd->bookingInfo = $bookingInfoUpdated->getBookingInfo()->toJsonLd();
+        }
 
         return $document->withBody($offerLd);
     }
--- a/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
+++ b/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
@@ -17,9 +17,6 @@ use CultuurNet\UDB3\SameAsForUitInVlaanderen;
 
 final class PropertyPolyfillOfferRepository extends DocumentRepositoryDecorator
 {
-    // All fields that contain arrays but can be null, should be removed when array is empty
-    private const FILTERED_NULL_LABELS = ['labels', 'hiddenLabels', 'bookingInfo'];
-
     private ReadRepositoryInterface $labelReadRepository;
 
     private OfferType $offerType;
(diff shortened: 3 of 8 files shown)


tests/Event/Events/BookingInfoUpdatedTest.php
```diff
@@ -67,6 +67,18 @@ public function serializationDataProvider(): array
                     )
                 ),
             ],
+            'bookingInfoDeleted' => [
+                [
+                    'item_id' => 'foo',
+                    'bookingInfo' => [
+
+                    ],
```
simon-debruijn, line 75:
```suggestion
                    'bookingInfo' => [],
```

simon-debruijn (approved):

JonasVHG (commented):
I think we also should also add a polyfill to fix offers where the bookinginfo was deleted in the past. Now it will only work on new removals of BookingInfo.

LucWollants:
> I think we also should also add a polyfill to fix offers where the bookinginfo was deleted in the past. Now it will only work on new removals of BookingInfo.

Nice suggestion @JonasVHG! If you need help with this @grubolsch let us know

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -631,7 +631,11 @@ protected function applyBookingInfoUpdated(AbstractBookingInfoUpdated $bookingIn
 
         $offerLd = $document->getBody();
 
-        $offerLd->bookingInfo = $bookingInfoUpdated->getBookingInfo()->toJsonLd();
+        if (empty($bookingInfoUpdated->getBookingInfo()->toJsonLd())) {
```
LucWollants, line 634: I think this is something that can also be tested with a unit test on `OfferLDProjectorTest`?

LucWollants (changes requested):
The fix is in place, but I think 2 extra things can be done:
- Add an extra unit test in `OfferLDProjectorTest`
- Add a polyfil to remove an empty booking request from the JSON projection

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
grubolsch: Not only did I provide a unit test for this, I also added another unit test to add bookinginfo, because this seemed to be missing.

features/event/booking-info.feature
```diff
@@ -253,4 +253,17 @@ Feature: Test the UDB3 events API
         }
       ]
     }
-    """
\ No newline at end of file
+    """
+
+  Scenario: Delete event booking info (with put)
+    Given I create an event from "events/event-with-bookinginfo.json" and save the "url" as "eventUrl"
```
LucWollants, line 259: There is already an event `event-with-all-fields.json`, maybe reuse that one?

src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
```diff
@@ -17,6 +17,9 @@
 
 final class PropertyPolyfillOfferRepository extends DocumentRepositoryDecorator
 {
+    // All fields that contain arrays but can be null, should be removed when array is empty
+    private const FILTERED_NULL_LABELS = ['labels', 'hiddenLabels', 'bookingInfo'];
```
LucWollants, line 21: The name is misleading in the context of `bookingInfo`

src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
```diff
@@ -235,8 +238,9 @@ function (array $json) {
                     return $json;
                 };
 
-                $json = $filterNullLabels($json, 'labels');
-                $json = $filterNullLabels($json, 'hiddenLabels');
+                foreach (self::FILTERED_NULL_LABELS as $filteredNullLabel) {
+                    $json = $filterNullLabels($json, $filteredNullLabel);
```
LucWollants, line 242: Also the method name can be more general, maybe something like `removeEmptyArray`

LucWollants: While having a closer look, this can be dangerous to use in the context of bookingInfo because the `array_filter`

grubolsch: I have made it into a separate method

features/event/booking-info.feature
LucWollants: As discussed through Slack this test existing file also requires images, so we keep the new smaller event with booking info.

LucWollants (approved):

PR #1443, opened 2023-08-09
III-5770: Add info connect docker to frontend

JonasVHG:
### Added

- `docker.md`: added info about how to connect to [udb3-frontend](https://github.com/cultuurnet/udb3-frontend)

---
Ticket: https://jira.uitdatabank.be/browse/III-5770

### DIFF
--- a/docker.md
+++ b/docker.md
@@ -53,3 +53,10 @@ To execute all CI tasks, run the following command:
 ```
 $ make ci
 ```
+
+### Frontend (WIP)
+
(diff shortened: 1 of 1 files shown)


LucWollants:
I think this belongs in `udb3-frontend` because that is the repository that requires this change.

JonasVHG:
> I think this belongs in `udb3-frontend` because that is the repository that requires this change.

Oke, I'll put it there.

PR #1474, opened 2023-09-13
III-3853 StringLiteral cleanup security

JonasVHG:
### Changed
- Refactored the `Security` namespace to use native `string` instead of deprecated `StringLiteral`

---

### DIFF
--- a/src/Security/CommandAuthorizationException.php
+++ b/src/Security/CommandAuthorizationException.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Security;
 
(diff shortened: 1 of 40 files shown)


src/Security/CommandAuthorizationException.php
```diff
@@ -4,25 +4,20 @@
 
 namespace CultuurNet\UDB3\Security;
 
-use CultuurNet\UDB3\StringLiteral;
-
 class CommandAuthorizationException extends \Exception
```
JonasVHG, line 7: This could probably made final.

LucWollants (approved):

PR #1482, opened 2023-09-19
III-5801 Add RDF projection for organizer on an event

LucWollants:
### Added
- Added RDF projection for organizer on an event

---
Ticket: https://jira.uitdatabank.be/browse/III-5801

### DIFF
--- a/src/Event/ReadModel/RDF/RdfProjector.php
+++ b/src/Event/ReadModel/RDF/RdfProjector.php
@@ -141,9 +141,10 @@ final class RdfProjector implements EventListener
             $this->setOrganizer($resource, $event->getOrganizerReference());
         }
 
-        (new WorkflowStatusEditor())->setWorkflowStatus($resource, $event->getWorkflowStatus());
+        $workflowStatusEditor = new WorkflowStatusEditor();
+        $workflowStatusEditor->setWorkflowStatus($resource, $event->getWorkflowStatus());
         if ($event->getAvailableFrom()) {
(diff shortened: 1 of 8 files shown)


src/Event/ReadModel/RDF/RdfProjector.php
```diff
@@ -131,6 +137,10 @@ public function handle(DomainMessage $domainMessage): void
 
         $this->setTerms($resource, $event->getTerms());
 
+        if ($event->getOrganizerReference()) {
+            $this->setOrganizer($resource, $event->getOrganizerReference());
+        }
+
         (new WorkflowStatusEditor())->setWorkflowStatus($resource, $event->getWorkflowStatus());
         if ($event->getAvailableFrom()) {
             (new WorkflowStatusEditor())->setAvailableFrom($resource, $event->getAvailableFrom());
```
grubolsch, line 146: Is there a reason this WorkflowStatusEditor is not a dependency, or only created once?
Maybe you can make it once and then do 2 calls on it?

grubolsch (approved):
Een opmerking, voor de rest ziet er goed uit

src/Event/ReadModel/RDF/RdfProjector.php
LucWollants: There will be only one implementation of the `WorkflowStatusEditor`, another approach could have been a `trait`. But I will store it inside a local variable.

PR #1485, opened 2023-09-20
III-3853 remove legacy address

grubolsch:
### Changed
Removed legacy Address class, replace everywhere with the new UDB3 Address class

Ran E2E test afterwards and no problems came up.

---


LucWollants:
As discussed through Slack replacing the legacy `Address` is risky. Best to keep the focus on "just" replacing `StringLiteral`

PR #1489, opened 2023-09-21
III-3853 Remove stringliteral on legacy postal code

grubolsch:
### Changed
Remove stringliteral on legacy postal code

---

### DIFF
--- a/src/Address/PostalCode.php
+++ b/src/Address/PostalCode.php
@@ -4,12 +4,31 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Address;
 
-use CultuurNet\UDB3\StringLiteral;
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\Trims;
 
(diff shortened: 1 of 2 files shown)


src/Address/PostalCode.php
```diff
@@ -4,12 +4,31 @@
 
 namespace CultuurNet\UDB3\Address;
 
-use CultuurNet\UDB3\StringLiteral;
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\Trims;
```
LucWollants, line 7: We could define extra traits in the `String\Behaviour` namespace, for example:
- `Equals` which implements `sameValueAs`
- `Converts` which implements `toNative`

LucWollants (approved):

src/Address/PostalCode.php
LucWollants: Also fine indeed!

PR #1490, opened 2023-09-21
III-3853 Remove stringliteral on legacy locality

grubolsch:
### Changed
Remove stringliteral on legacy locality

---


LucWollants (approved):
I already approved, but the same remark as on https://github.com/cultuurnet/udb3-backend/pull/1489#discussion_r1332755906

PR #1494, opened 2023-09-21
III-3853 StringLiteral Media/Description

JonasVHG:
### Changed

- Do not use `StringLiteral` in `Media\Description`
- Do not mix `StringLiteral` & `Media\Description`, use `Media\Description` consistently 

### Fixed

- Code Style update

---

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -101,7 +101,7 @@ use CultuurNet\UDB3\Title;
 use DateTimeImmutable;
 use DateTimeInterface;
 
-class Event extends Offer
+final class Event extends Offer
 {
     protected string $eventId;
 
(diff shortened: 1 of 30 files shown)


src/Event/Event.php
```diff
@@ -100,7 +101,7 @@
 use DateTimeImmutable;
 use DateTimeInterface;
 
-class Event extends Offer
+final class Event extends Offer
 {
     protected string $eventId;
```
grubolsch, line 106: Does this need to be protected?

grubolsch (approved):
1 kleine opmerking, eigelijk niets dat jij veranderd hebt maar misschien een chance die je kan meenemen? Voor de rest toppie!

src/Event/Event.php
LucWollants: Correct, many properties have `protected` scope but should have `private` scope. I think we need to have a script to fix those things. Maybe something like https://getrector.com/

PR #1502, opened 2023-09-21
III-3853 Remove stringliteral from EventExport

grubolsch:
### Changed
Remove stringliteral from EventExport

---


src/EventExport/Format/HTML/Properties/Footer.php
```diff
@@ -4,8 +4,17 @@
 
 namespace CultuurNet\UDB3\EventExport\Format\HTML\Properties;
 
-use CultuurNet\UDB3\StringLiteral;
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\IsString;
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\Trims;
 
-class Footer extends StringLiteral
+class Footer
```
JonasVHG, line 10: These & the other classes in the `Format/HTML/Properties`-Folder could probably made final

JonasVHG (approved):
One minor suggestion, otherwise it's ok.

src/EventExport/Format/HTML/Properties/Footer.php
grubolsch: done

PR #1503, opened 2023-09-21
III-3853 Remove StringLiteral from UDB3\UiTID

grubolsch:
### Changed
Remove StringLiteral from UDB3\UiTID

---


src/EventHandling/DelegateEventHandlingToSpecificMethodTrait.php
```diff
@@ -37,7 +37,7 @@ private function getHandleMethodName($event): ?string
 
         $expectedClass = $parameter->getClass();
 
-        if ($expectedClass->getName() !== get_class($event)) {
+        if ($expectedClass !== null && $expectedClass->getName() !== get_class($event)) {
```
LucWollants, line 40: If `$expectedClass` is `null` should this method not return `null`?

tests/Event/ReadModel/Permission/ProjectorTest.php
```diff
@@ -68,7 +68,7 @@ public function it_adds_permission_to_the_user_identified_by_the_createdby_eleme
             $payload
         );
 
-        $userId = new StringLiteral('123');
+        $userId = 'dcd1e123-0608-4824-afe3-99124feda64b';
 
         $this->userIdResolver->expects($this->once())
             ->method('resolveCreatedByToUserId')
```
JonasVHG, line 74: I guess this comes from a mergeConflict, but there is a StringLiteral in the next line, that could be removed.
Same with some other markResourceEditableByUser() in this file

src/EventHandling/DelegateEventHandlingToSpecificMethodTrait.php
grubolsch: Changed it into an or

tests/Event/ReadModel/Permission/ProjectorTest.php
grubolsch: Removed the other string literals

JonasVHG: My mistake, should have been more clear 🫣, they are also in `markResourceEditableByUser()` in `tests/Place/ReadModel/Permission/ProjectorTest.php`

grubolsch: Updated it

LucWollants (approved):

JonasVHG (approved):

PR #1504, opened 2023-09-21
Add Uses comments so the dynamically called methods of applyX can be followed

grubolsch:
### Added
Add Uses comments so the dynamically called methods of applyX can be followed.

There where a lot of methods that showed up as unused in the IDE because they where dynamically called from the handle() method. 
Adding this comment fixes this and allows you to directly click from both directions to the other method: handy for beginners in the codebase (like me)!

I have not done the comments for all 16 classes that use this trait, just for a couple, because I want the group opinion first.


LucWollants (changes requested):
As discussed through Slack, the `@use` comments should be added to the aggregates and the projectors.

PR #1512, opened 2023-09-22
III-3757 fix php stan warnings

grubolsch:
### Changed
Removed phpstan ignore comments where possible.
When not possible I added comments if it could be fixed when upgrading to PHP 8

---
Ticket: https://jira.uitdatabank.be/browse/III-3757


tests/CalendarTest.php
```diff
@@ -107,39 +107,6 @@ public function setUp(): void
         );
     }
 
-    /**
-     * @test
-     */
-    public function time_stamps_need_to_have_type_time_stamp(): void
```
LucWollants, line 113: Why are these tests removed?

LucWollants, line 113: I would keep the tests, what if PHPStan would be removed from the project?

grubolsch: restored tests

grubolsch:
This pr is ready for review. All feedback was processed.

LucWollants (approved):

PR #1514, opened 2023-09-25
III-5813 trait for old stringliterals

grubolsch:
### Changed
Changed all old value objects to use IsString where possible 

---
Ticket: https://jira.uitdatabank.be/browse/III-5813


tests/BackwardsCompatiblePayloadSerializerFactoryTest.php
```diff
@@ -82,7 +82,15 @@ public function it_handles_main_language(
         /** @var EventCreated|PlaceCreated|OrganizerCreatedWithUniqueWebsite $created */
         $created = $this->serializer->deserialize($decoded);
 
-        $this->assertEquals($expectedMainLanguage, $created->getMainLanguage());
+        /*
+         * The signature of getMainLanguage() is not consistent, sometimes this returns a string, sometimes a Language object
```
LucWollants, line 86: This is because the new approach for events is to only use primitive types without validation. Otherwise, if we take the strict validation from the JSON validation or the VO it could lead to runtime exceptions when reading the event stream of older resources.

LucWollants (approved):

PR #1516, opened 2023-09-25
III-5818 event_permission_readmodel check for UPDATE/INSERT

JonasVHG:
### Changed
- `DBALResourceOwnerRepository`: Add check to see if the `resourceId` is present in `event_permission_readmodel`, if it is not present preform an `INSERT` instead of an `UPDATE`. If it already exists, the code remains an `UPDATE`

### Fixed

- When changing the `owner` of an `Offer` with an unknown `creator`(e.g., legacy Offers, with creators who could not be mapped to an `UiTID`), the new `owner` will have edit rights.

---
Ticket: https://jira.uitdatabank.be/browse/III-5818

### DIFF
--- a/src/Security/ResourceOwner/Doctrine/DBALResourceOwnerRepository.php
+++ b/src/Security/ResourceOwner/Doctrine/DBALResourceOwnerRepository.php
@@ -63,22 +63,15 @@ final class DBALResourceOwnerRepository implements ResourceOwnerRepository, Reso
 
     public function markResourceEditableByNewUser(string $resourceId, string $userId): void
     {
-        /*$rowCount = $this->connection->createQueryBuilder()
+        $rowCount = count($this->connection->createQueryBuilder()
             ->select($this->idField)
             ->from($this->tableName)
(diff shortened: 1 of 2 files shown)


src/Security/ResourceOwner/Doctrine/DBALResourceOwnerRepository.php
```diff
@@ -63,10 +63,27 @@ public function markResourceEditableByUser(string $resourceId, string $userId):
 
     public function markResourceEditableByNewUser(string $resourceId, string $userId): void
     {
-        $this->connection->update(
-            $this->tableName,
-            ['user_id' => $userId],
-            [$this->idField => $resourceId]
+        $results = $this->connection->fetchAll(
+            'SELECT * FROM ' . $this->tableName . ' WHERE event_id = :eventId',
```
LucWollants, line 67: Maybe we better stick with the DBAL query language instead of writing SQL statements. For example on line 33.

LucWollants (approved):

PR #1520, opened 2023-09-28
III-5831 Move `Calendar` related classes to the `Calendar` namespace

LucWollants:
### Changed
- Moved `Calendar` related classes to the `Calendar` namespace (this will group new custom exceptions related to invalid calendars inside the `Calendar` namespace)

---
Ticket: https://jira.uitdatabank.be/browse/III-5831


tests/Calendar/week_scheme.xml
JonasVHG, line 1: Probably an oversight, but I think moving this file to tests/Calendar/Samples would be more logical.

JonasVHG (approved):
Just 1 file, that I think should be moved to a deeper subfolder (`test/Calendar/samples` instead of `test/Calendar`)
Rest looks good, so I'll already approve.

PR #1532, opened 2023-11-08
WIP III-5875 Add config to manage if a route requires authentication

grubolsch:
### Changed
Added a configuration parameter to mark which routes are always requiring authentication, or make it toggleable dependent on a GET param.

Example in the config.php file:
```php
'public_routes_with_auth' => [
        '~^/events/[\w\-]+/?$~' => [
            'mode' => 'param',
            'param' => 'embedContributors',
        ],
        '~^/places/[\w\-]+/?$~' => [
            'mode' => 'always',
        ],
    ]
```

**Open questions:**
I am not sure about 2 details yet:

1. Should I remove all the calls to addPublicRoute() and move those to my config file? I am not sure the logic is now doubled.
2. When sending an authentication header the code will always try to authenticate you, even when not required. Is this oké? See RequestAuthenticatorMiddleware.php:88

---
Ticket: https://jira.uitdatabank.be/browse/III-5875

### DIFF
--- a/src/Http/Auth/PublicRouteWithAuthMatcher.php
+++ b/src/Http/Auth/PublicRouteWithAuthMatcher.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Auth;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+final class PublicRouteWithAuthMatcher
+{
+    private const ALWAYS = 'always';
+    private const PARAM = 'param';
+
+    /** @var PublicRouteRule[] */
+    private array $publicRoutes = [];
+
+    private array $publicRoutesWithAuth;
+
+    public function __construct(array $publicRoutesWithAuth)
+    {
+        $this->publicRoutesWithAuth = $publicRoutesWithAuth;
+    }
+
+    public function isAuthenticationRequired(ServerRequestInterface $request): bool
+    {
+        foreach ($this->publicRoutes as $publicRouteRule) {
+            if ($publicRouteRule->matchesRequest($request) && isset($this->publicRoutesWithAuth[$publicRouteRule->getPathPattern()])) {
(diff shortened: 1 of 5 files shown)


LucWollants:
@grubolsch I think we now have 3 kinds of routes:
- public routes without authentication
- public routes with authentication
- private routes

I think with that approach we can do something like:
```
if preflight => always return
if public => but not public with authentication => always return
otherwise => authenticate
```

So I also think that `RequestAuthenticatorMiddleware::addPublicRoute` should work as before and that works in combination with your new `PublicRouteWithAuthMatcher`

src/Http/Auth/PublicRouteWithAuthMatcher.php
```diff
@@ -0,0 +1,58 @@
...
+                if (mb_strtolower($authConfig['mode']) === self::ALWAYS) {
+                    return true;
+                }
+
+                if (mb_strtolower($authConfig['mode']) === self::PARAM) {
+                    return isset($request->getQueryParams()[$authConfig['param']]) && $request->getQueryParams()[$authConfig['param']] !== false;
+                }
+
+                return false;
+            }
```
simon-debruijn, line 48: Maybe these ifs could be less nested by using early returns

src/Http/Auth/PublicRouteWithAuthMatcher.php
```diff
@@ -0,0 +1,58 @@
...
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Http\Auth;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+final class PublicRouteWithAuthMatcher
+{
+    private const ALWAYS = 'always';
+    private const PARAM = 'param';
```
simon-debruijn, line 12: You can make these consts public and use them in the config instead of plain strings

simon-debruijn (approved):

PR #1533, opened 2023-11-08
WIP III-5856 Added extra info to cached addresses

grubolsch:
### Changed
Added extra information to geocaching redis cache, if a feature flag is enabled "address_enrichment".

---
Ticket: https://jira.uitdatabank.be/browse/III-5856


app/Geocoding/GeocodingServiceProvider.php
```diff
@@ -42,7 +44,8 @@ function () use ($container) {
 
                 return new CachedGeocodingService(
                     $geocodingService,
-                    $container->get('cache')('geocoords')
+                    $container->get('cache')('geocoords'),
+                    empty($container->get('config')['address_enrichment']) ? new CoordinateEncoder() : new RichEncoder()
```
JonasVHG, line 48: While a featureflag is a good idea, I would avoid changing the existing code/functionality because that could introduce bugs both with and without the featureflag.

src/Geocoding/CacheEncoder/RichEncoder.php
```diff
@@ -0,0 +1,36 @@
...
+namespace CultuurNet\UDB3\Geocoding\CacheEncoder;
+
+use Geocoder\Location;
+
+class RichEncoder implements CacheEncoder
+{
+    private const POSTFIX = '_enriched';
+
+    public function encode(Location $location): array
+    {
```
JonasVHG, line 14: Since this is for the duplicate places it might be a good idea to check with Frederic how they will get the data.
I doubt the data-team can access the cache. So maybe a logfile or database table?

PR #1540, opened 2023-11-16
III-5875 Added check that contributors are not leaked in sapi3 responses

grubolsch:
### Added
Added check that contributors are not leaked in sapi3 responses

---
Ticket: https://jira.uitdatabank.be/browse/III-5875

### DIFF
--- a/features/search/contributors.feature
+++ b/features/search/contributors.feature
@@ -26,6 +26,7 @@ Feature: Test the Search API v3 contributors
       | q                     | contributors:%{organizerContributorEmail} |
     Then the JSON response at "totalItems" should be 1
     And the JSON response at "member/0/@id" should be "%{organizerUrl}"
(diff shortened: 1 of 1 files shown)


JonasVHG (approved):
Looks good.

Only super minor remark, while you are editing the file. It might be a good idea to also update the error prone  `And I wait 2 seconds` and replace it with `And I wait for the event/place/organizer with url :url to be indexed`.

LucWollants:
> Only super minor remark, while you are editing the file. It might be a good idea to also update the error prone `And I wait 2 seconds` and replace it with `And I wait for the event/place/organizer with url :url to be indexed`.

That is a good suggestion, but unfortunately, this will not work. The wait for indexation can only be used for new items, when an update is done to an item it will return immediately. This is because the code only checks for the availability of the indexed item.

simon-debruijn (approved):

PR #1556, opened 2023-12-15
III-4601 Prevent duplicate places based on unique identifier

grubolsch:
### Changed
- Added a check in the ImportPlaceRequestHandler.php to prevent double places on the same address, based on a unique address identifier in SAPI3.

---


src/Place/ReadModel/Duplicate/LookupDuplicatePlaceWithSapi3.php
```diff
@@ -0,0 +1,73 @@
...
+            } catch (DocumentDoesNotExist $e) {
+            }
+        }
+
+        if ($this->sapi3SearchService->getLastRequestedUri() === null) {
+            // This should never happen, but probably not worth it to crash the Place API request for it
+            return $results->getItems()[0]->getUrl()->toString();
+        }
+
+        return $this->sapi3SearchService->getLastRequestedUri()->__toString();
```
LucWollants, line 71: I would never return the string of the SAPI3 query here. The responsibility for `getDuplicatePlaceUri` is just to return the duplicate place. Suppose there are more than 1 duplicate places throw an error.

In the calling code catch the error and then display the query to the API consumer. That approach also does not require a change to `sapi3SearchService` because the caller knows the query.

src/Place/ReadModel/Duplicate/UniqueAddressIdentifierFactory.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\ReadModel\Duplicate;
+
+use CultuurNet\UDB3\Model\ValueObject\Geography\Address;
+
+class UniqueAddressIdentifierFactory
+{
+    public function hash(string $title, Address $address, string $currentUserId): string
```
LucWollants, line 11: A hash is not created, so I would call the method `create` instead of `hash`

src/Search/ReturnLastRequestedUri.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search;
+
+use Psr\Http\Message\UriInterface;
+
+interface ReturnLastRequestedUri
```
LucWollants, line 9: I don't think this is needed, see https://github.com/cultuurnet/udb3-backend/pull/1556/files#r1455437664

tests/Http/Place/ImportPlaceRequestHandlerTest.php
```diff
@@ -143,7 +144,9 @@ protected function setUp(): void
             ),
             new CallableIriGenerator(fn ($placeId) => 'https://io.uitdatabank.dev/places/' . $placeId),
             $this->commandBus,
-            $this->imageCollectionFactory
+            $this->imageCollectionFactory,
+            true,
+            $this->createMock(LookupDuplicatePlace::class),
```
LucWollants, line 149: This can use some extra unit tests and the feature test can be removed.

LucWollants (changes requested):
I added a few remarks that needs to be addressed first.

src/Place/ReadModel/Duplicate/LookupDuplicatePlaceWithSapi3.php
grubolsch: I am not sure I understand.
Instead of return a string here you want me to throw an exception?
But then how does the exception handler know the query?
That information is encapsulated in sapi3SearchService.

Or do you want me to alter sapi3SearchService to throw an error when more then 1 result is found?
I would have to be careful then, that classes is used in a lot of places.

LucWollants: On line https://github.com/cultuurnet/udb3-backend/pull/1556/files/08862fb23a6cf9b50de2c0e488b190b6e1aa5fc5#diff-979f1a5d2f82f93de80342e71bc787a2748dda27c1bf5d9372fa50c658f4fe2aR38 you create the query so you can use that value to throw a new exception, something like `CouldNotDetermineOriginalPlace`

tests/Http/Place/ImportPlaceRequestHandlerTest.php
grubolsch: Wrote the extra tests (feature test is in a separate pr, and I think both are useful)

src/Place/ReadModel/Duplicate/LookupDuplicatePlaceWithSapi3.php
```diff
@@ -0,0 +1,71 @@
...
+    public function getDuplicatePlaceUri(Place $place): ?string
+    {
+        $query = 'unique_address_identifier:' .
+            $this->addressIdentifierFactory->create(
+                $place->getTitle()->getTranslation($place->getMainLanguage())->toString(),
+                $place->getAddress()->getTranslation($place->getMainLanguage()),
+                $this->currentUserId
+            );
+
+        $results = $this->sapi3SearchService->search(
```
LucWollants, line 45: If you set a limit here of `1` the total count will either be `0` or `1`, so you will never reach line 58. This needs some tweaking. It does not result in an error in the unit tests because there the search service is mocked and the limit parameter is discarded.

LucWollants (changes requested):
There is still an issues with the search query, there is a hardcoded limit of 1

grubolsch:
> There is still an issues with the search query, there is a hardcoded limit of 1

I removed the limit, the default is now 30 - that should be enough

I also remove the logic for duplicateBy, and will re-introduce this later when we can write a feature test for this.

PR #1578, opened 2024-01-17
III-4601 Add feature test for duplicate places

grubolsch:
### Added
Added a feature test for the duplicate places check

---


features/Steps/UtilitySteps.php
```diff
@@ -37,4 +39,51 @@ public function iWaitSeconds(int $seconds): void
...
+
+        $config['prevent_duplicate_creation'] = $bool;
+
+        file_put_contents('config.php', '<?php' . PHP_EOL . PHP_EOL . 'return ' . $this->shorthandVarExport($config) . ';');
+
+        return true;
+    }
+
+    // var_export uses old school array() style, this function converts this to the modern [] standard
+    private function shorthandVarExport(array $expression): string
```
LucWollants, line 79: To avoid this you could also do a simple text replacement and don't convert to an array.

Replace `'prevent_duplicate_creation' => true` with `'prevent_duplicate_creation' => false` to turn the feature off.

 Replace `'prevent_duplicate_creation' => false` with `'prevent_duplicate_creation' => true` to turn the feature on.

features/place/duplicate.feature
```diff
@@ -7,12 +7,12 @@ Feature: Test creating places
     And I send and accept "application/json"
 
   Scenario: Be prevented from creating a new place if we already have one on that address
-    Given I create a minimal place and save the "id" as "placeId" then I should get a "201" response code
-    Given I create a minimal place and save the "id" as "placeId" then I should get a "303" response code
-    Then the JSON response should be:
+    Given I prevent duplicate creation
+    Given I create a random name of 6 characters and keep it as "name"
```
LucWollants, line 11: This call is no longer needed, the duplicate error comes from executing lines 12 and 13

LucWollants (changes requested):
See my added remarks to finetune this PR

features/Steps/UtilitySteps.php
grubolsch: I didn't want to do a string conversion because that might be more vulnerable for slight differences in syntacs, for example somebody using double quotes for the property, or not using spaces.

features/place/duplicate.feature
grubolsch: I don't think so, because the default place that is always created (den engel) will probably already exist because of a previous feature test. So I need to make sure the first place can be created.

features/data/places/place-with-required-fields-and-variable-name.json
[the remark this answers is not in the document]
grubolsch: See above, I think it is needed to have a place with a variable name.

features/place/duplicate.feature
LucWollants: Ok, I made a mistake here because I did not find any usages of `place-with-required-fields-and-variable-name.json`

features/data/places/place-with-required-fields-and-variable-name.json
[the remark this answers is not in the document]
LucWollants: A made a wrong assumption.

grubolsch:
processed all feedback

grubolsch:
Pr feedback has been processed

features/Steps/UtilitySteps.php
```diff
@@ -37,4 +39,44 @@ public function iWaitSeconds(int $seconds): void
...
     }
+
+    /**
+     * @Given /^I prevent duplicate creation$/
+     */
+    public function iPreventDuplicateCreation(): void
+    {
+        $configFile = file_get_contents('config.php');
+
+        if (str_contains($configFile, "'prevent_duplicate_creation' => true")) {
```
simon-debruijn, line 50: Wouldn't it be safer to load the config object into a variable, then check and modify the object, and then serialize it back into the file?
Using `str_contains` and `str_replace` seems a bit more error prone

simon-debruijn (approved):

features/Steps/UtilitySteps.php
grubolsch: That was what I did first, but Luc advised me to change this to make it more simple.

simon-debruijn: Haha ok never mind then 😉 I didn't look at the resolved comments

LucWollants: With the approach of loading the file into a variable an extra conversion was needed when streaming the contents back to the file. See this method https://github.com/cultuurnet/udb3-backend/pull/1578/commits/bc53661bbce982b3f18de08585cf6314045d0ff2#diff-c4ca10529a97c122ff5754d13c511ce2c822c4576c46a797a80c14a1335562a8L78-L88

features/Steps/UtilitySteps.php
```diff
@@ -37,4 +39,44 @@ public function iWaitSeconds(int $seconds): void
...
+
+        file_put_contents('config.php', $configFile);
+
+        $this->initialPreventDuplicateCreationValue = false;
+    }
+
+    /**
+     * @Then /^I allow duplicate creation$/
+     */
+    public function iAllowDuplicateCreation(): void
```
LucWollants, line 70: This method name is not correct, it does restore the prevent duplication config.

features/place/duplicate.feature
```diff
@@ -7,12 +7,13 @@ Feature: Test creating places
...
+    Given I create a minimal place and save the "id" as "originalPlaceId" then I should get a "201" response code
+    Then I wait for the place with url "/places/%{originalPlaceId}" to be indexed
+    Given I create a minimal place and save the "originalPlace" as "newPlaceUri" then I should get a "409" response code
+    Then the JSON response at "originalPlace" should be:
     """
-    {
-      "message": "A place with this address / location name combination already exists. Please use the existing place for your purposes.",
-      "placeId": "%{placeId}",
-    }
+    "http://host.docker.internal:9000/places/%{originalPlaceId}"
```
LucWollants, line 17: Very brittle to use a hardcoded value here!

features/Steps/AuthorizationSteps.php
```diff
@@ -14,6 +14,8 @@ trait AuthorizationSteps
     public function iAmUsingTheUDB3BaseURL(): void
     {
         $this->variableState->setVariable('baseUrl', $this->config['base_url']);
+        $this->variableState->setVariable('base_url_sapi3', $this->config['base_url_sapi3']);
```
LucWollants, line 17:
```suggestion
        $this->variableState->setVariable('baseUrlSapi3', $this->config['base_url_sapi3']);
```

LucWollants (approved):

PR #1579, opened 2024-01-17
III-4601 Duplicate place redis cache

grubolsch:
### Added
III-4601 Duplicate place redis cache

---


LucWollants (changes requested):
I think this approach could lead to issues if a place gets a modified address. The old address is still inside the cache and would prevent the creation of a place with the old address.

PR #1581, opened 2024-01-18
III-5887 - Add DeleteCurrentOrganizerHandler to catch Full updates of Events with TicketSales

JonasVHG:
### Added

- `CommandHandler\DeleteCurrentOrganizerHandler` & test: Added to inject `EventHasTicketSalesGuard`

### Changed

- `OfferServiceProvider.php`: add `DeleteCurrentOrganizerHandler` to `container`

- `CommandBusServiceProvider`: subscribe `CommandBus` to `DeleteCurrentOrganizerHandler`

- `OfferCommandHelper` & Children `EventCommandHelper`, `PlaceCommandHelper` & `ItemCommandHelper`: remove `getDeleteCurrentOrganizerClassName()` `handleDeleteCurrentOrganizer()`

- Made `Offer\DeleteCurrentOrganizer` `final` instead of `abstract`

- refacored `tests`

### Removed

- No longer needed `childClasses` of `Offer\DeleteCurrentOrganizer`

### Fixed

- One can no longer delete the `Organizer`, when an `Event` has `TicketSales` during a full-update

- Some boyScouting, made `Classes` final where appropriate.

---
Ticket: https://jira.publiq.be/browse/III-5887


src/Offer/CommandHandlers/DeleteCurrentOrganizerHandler.php
```diff
@@ -0,0 +1,39 @@
...
+        OfferRepository $offerRepository,
+        EventHasTicketSalesGuard $organizerChangeAllowedBasedOnTicketSales
+    ) {
+        $this->offerRepository = $offerRepository;
+        $this->organizerChangeAllowedBasedOnTicketSales = $organizerChangeAllowedBasedOnTicketSales;
+    }
+
+    public function handle($command): void
+    {
+        if (!($command instanceof DeleteCurrentOrganizer)) {
```
grubolsch, line 28: Technicly you don't need the inner - (), instanceof has a higher priority than the !.

src/Offer/CommandHandlers/DeleteCurrentOrganizerHandler.php
```diff
@@ -0,0 +1,39 @@
...
+            return;
+        }
+
+        $offer = $this->offerRepository->load($command->getItemId());
```
grubolsch, line 32: Are we sure this always gives an offer back? load() can throw an EventStreamNotFoundException, should we handle that here?

grubolsch (approved):

src/Offer/CommandHandlers/DeleteCurrentOrganizerHandler.php
JonasVHG: Nicely spotted.

src/Offer/CommandHandlers/DeleteCurrentOrganizerHandler.php
JonasVHG: Yes, if you include an unknown `id`, a new `Offer` is created, and you cannot delete an Organizer during a `create` only during an `update` so the `OfferRepository` should always return an `Offer`.
Whether this logic desirable is another matter, but I thought that was for legacy reasons like `OMD`.

LucWollants: Throwing the exception `AggregateNotFoundException` is already handled inside the `ApiProblemFactory`. Inside the command handlers, there is no guarantee that the ID passed is that of an existing item.

PR #1584, opened 2024-01-23
III-5979/bugfix updating place with event

grubolsch:
### Added
Add check that you cannot get a place with a valid event id, or an event with a valid place id.
This could be because the consumer of the API switched up the endpoints of the API.

---
Ticket: https://jira.uitdatabank.be/browse/III-5979/bugfix


simon-debruijn:
Is it still a WIP? Then you can set it to draft

features/Steps/EventSteps.php
```diff
@@ -58,6 +58,19 @@ public function iGetTheEventAt(string $url): void
         $this->theResponseBodyShouldBeValidJson();
     }
 
+    /**
+     * @Given I fail to get the event at :url
+     */
+    public function iFailToGetTheEventAt(string $url): void
```
JonasVHG, line 64: I think the term `FailToGet` is too broad, is it because it does not exist, you do not have permission, it is in the past,...
Since it's a `404` maybe something along the line of `iDoNotFindtheEvent()`

features/Steps/PlaceSteps.php
```diff
@@ -92,6 +92,19 @@ public function iGetThePlaceAt(string $url): void
         $this->theResponseBodyShouldBeValidJson();
     }
 
+    /**
+     * @Given I fail to get the place at :url
+     */
+    public function iFailToGetThePlaceAt(string $url): void
```
JonasVHG, line 98: Same as above

src/Http/CheckTypeOfOfferMiddleware.php
```diff
@@ -0,0 +1,47 @@
...
+            $offerType = $routeParameters->getOfferType();
+        } catch (RuntimeException $e) {
+            return $handler->handle($request);
+        }
+
+        $offerId = $routeParameters->getOfferId();
+
+        if (OfferType::event()->sameAs($offerType)) {
+            $this->eventRepository->fetch($offerId);
+        } elseif (OfferType::place()->sameAs($offerType)) {
```
JonasVHG, line 41: The elseif isn't necessary.
```suggestion
        }
        if (OfferType::place()->sameAs($offerType)) {
```
has the same logic since $offerType can't be both.

app/Authentication/AuthServiceProvider.php
```diff
@@ -42,6 +43,13 @@ public function register(): void
 
         CurrentUser::configureGodUserIds($container->get('config')['user_permissions']['allow_all']);
 
+        $container->addShared(CheckTypeOfOfferMiddleware::class, function () use ($container): CheckTypeOfOfferMiddleware {
```
LucWollants, line 46: This new class `CheckTypeOfOfferMiddleware` that is now part of the container also needs to be exposed with `getProvidedServiceNames`.

I also think a better location is `PsrRouterServiceProvider` or `OfferServiceProvider` because this new middleware is not related to authentication.

features/Steps/EventSteps.php
LucWollants: I would say there is even no reason to write a new step.

The scenario should be:
- Create a new event and store the UUID
- Get a place with that UUID and check if the error response is 404

LucWollants: Maybe a step like `iGetTheEventAtAndTheResponseIs` could help.

features/place/read.feature
```diff
@@ -0,0 +1,19 @@
...
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I send and accept "application/json"
+
+  @bugfix # https://jira.uitdatabank.be/browse/III-5979
+  Scenario: Try to get a event that actually is a place
+    Given I create a minimal place and save the "url" as "placeUrl"
+    # The line below gets an event which is actually a place - this is on purpose and should fail
```
LucWollants, line 12: When using a step like `iGetTheEventAtAndTheResponseIs` the comment is not needed. Furthermore the test is marked with a bug ticket so all is clear from the ticket.

features/place/read.feature
```diff
@@ -0,0 +1,19 @@
...
+    Then I fail to get the event at "%{placeUrl}"
+
+  @bugfix # https://jira.uitdatabank.be/browse/III-5979
```
LucWollants, line 15: This test should move to the event features

features/Steps/PlaceSteps.php
```diff
@@ -92,6 +92,32 @@ public function iGetThePlaceAt(string $url): void
...
+        );
+
+        $this->theResponseStatusShouldBe($statusCode);
+        $this->theResponseBodyShouldBeValidJson();
+    }
+
+    /**
+     * @Given I fail to get the place at :url
+     */
+    public function iFailToGetThePlaceAt(string $url): void
```
LucWollants, line 111: This method is no longer used

features/event/read.feature
```diff
@@ -0,0 +1,12 @@
...
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I send and accept "application/json"
+
+  @bugfix # https://jira.uitdatabank.be/browse/III-5979
+  Scenario: Try to get a event that actually is a place
+    Given I create a minimal place and save the "id" as "placeId"
+    Then I get the event at "http://host.docker.internal:8000/events/%{placeId}" and get response code "404"
```
LucWollants, line 12: The URL should not be hardcoded because this will give issues if we run them in the cloud.
A better approach is to do something like `iGetTheRdfOfEventWithId` and make a method like `iGetTheEventWithIdTheResponseCodeShouldBe`

src/Security/PermissionVoterCommandBusSecurity.php
```diff
@@ -13,7 +13,7 @@ final class PermissionVoterCommandBusSecurity implements CommandBusSecurity
     private PermissionVoter $permissionVoter;
 
     public function __construct(
-        ?string $userId,
+        string $userId,
```
LucWollants, line 16: Is this change needed in the context of this pull request?

tests/Http/CheckTypeOfOfferMiddlewareTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $placeRepository = new InMemoryDocumentRepository();
+        $placeRepository->save(new JsonDocument(self::PLACE_ID, '{}'));
+
+        $eventRepository = new InMemoryDocumentRepository();
+        $eventRepository->save(new JsonDocument(self::EVENT_ID, '{}'));
+
+        $middleware = new CheckTypeOfOfferMiddleware($placeRepository, $eventRepository);
+
+        if ($expectException) {
```
LucWollants, line 40: The happy paths don't have an assert or expect so you introduced risky tests, see https://github.com/cultuurnet/udb3-backend/actions/runs/7627184604/job/20775262612?pr=1584#step:8:91

LucWollants (changes requested):
This still requires some cleanup work.

src/Security/PermissionVoterCommandBusSecurity.php
grubolsch: No, but i am really sure this is a bug, for 2 reasons:
- It makes no sense that the property declaration and the constructor are different.
- When I did the opposite - make it both nullable, it just failed with other dependencies down the line.

If you want I can cherrypick this into the worlds smallest PR? ;-)

tests/Http/CheckTypeOfOfferMiddlewareTest.php
grubolsch: Problem is I don't really know what to test here. Before I mocked the EventRepo and placeRepo, and then I had ->expects($this->once()) and the correct repo, so at least I tested that.
I lost that because I now removed the mock.

I can reintroduce the mock if you want?

LucWollants: I think there is no need to test the happy paths, those are handled in various other unit tests

src/Security/PermissionVoterCommandBusSecurity.php
LucWollants: But the `getId` method from `CurrentUser` returns `?string` see https://github.com/cultuurnet/udb3-backend/blob/b491c80878447306225bac139da04f6e10deccc3/src/User/CurrentUser.php#L24 so method also has a wrong return type?

features/event/read.feature
```diff
@@ -0,0 +1,13 @@
...
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I send and accept "application/json"
+
+  @bugfix # https://jira.uitdatabank.be/browse/III-5979
+  Scenario: Try to get a event that actually is a place
+    Given I create a minimal place and save the "id" as "placeId"
+    Then I send a GET request to "/events/%{placeId}"
```
LucWollants, line 12:
```suggestion
    When I send a GET request to "/events/%{placeId}"
```

features/place/read.feature
```diff
@@ -0,0 +1,14 @@
...
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I send and accept "application/json"
+
+  @bugfix # https://jira.uitdatabank.be/browse/III-5979
+  Scenario: Try to get a place that actually is an event
+    Given I create a minimal place and save the "url" as "placeUrl"
+    Given I create a minimal permanent event and save the "id" as "eventId"
```
LucWollants, line 12:
```suggestion
    And I create a minimal permanent event and save the "id" as "eventId"
```

features/place/read.feature
```diff
@@ -0,0 +1,14 @@
...
+    Then I send a GET request to "/places/%{eventId}"
```
LucWollants, line 13:
```suggestion
    When I send a GET request to "/places/%{eventId}"
```

tests/Http/CheckTypeOfOfferMiddlewareTest.php
grubolsch: Removed the happy paths

LucWollants (approved):

PR #1589, opened 2024-01-25
III-5995 Fix typo in PrijsInfo RDF namespace

LucWollants:
### Fixed
- Fix typo in PrijsInfo RDF namespace

---
Ticket: https://jira.uitdatabank.be/browse/III-5995

### DIFF
--- a/features/data/events/rdf/event-with-all-fields.ttl
+++ b/features/data/events/rdf/event-with-all-fields.ttl
@@ -11,7 +11,7 @@
 @prefix schema: <https://schema.org/> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix labeltype: <https://data.cultuurparticipatie.be/id/concept/LabelType/> .
-@prefix cpp: <https://data.vlaanderen.be/ns/cultuurparticipatie#Prijsinfo> .
+@prefix cpp: <https://data.vlaanderen.be/ns/cultuurparticipatie#Prijsinfo.> .
 
 <http://host.docker.internal:8080/events/%{eventId}>
   a cidoc:E7_Activity ;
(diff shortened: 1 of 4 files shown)


grubolsch (approved):

features/data/events/rdf/event-with-all-fields.ttl
```diff
@@ -11,7 +11,7 @@
 @prefix schema: <https://schema.org/> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix labeltype: <https://data.cultuurparticipatie.be/id/concept/LabelType/> .
-@prefix cpp: <https://data.vlaanderen.be/ns/cultuurparticipatie#Prijsinfo> .
+@prefix cpp: <https://data.vlaanderen.be/ns/cultuurparticipatie#Prijsinfo.> .
```
simon-debruijn, line 14: The fix is correct. But isn't this prefix a bit misleading? `cpp` for `cultuurparticipatie`. But we only expose Prijsinfo subfields. Shouldn't the prefix be something like priceInfo following the labeltype prefix?

simon-debruijn (approved):

features/data/events/rdf/event-with-all-fields.ttl
LucWollants: I will have to pass this question to Stan. But I no longer find his GitHub user to mention him here, so passing it on through Slack.

JonasVHG (approved):

features/data/events/rdf/event-with-all-fields.ttl
Stiksels: hi @simon-debruijn ; good suggestion and something we could look into in the future. For now it's ok to keep using this prefix; it only exists in the named graph and is used for abbreviation of uri's at read-time. When storing RDF information in a triple store, all uri's are expanded in full. IF you, for example, copy/paste our turtle representation and convert it again to turtle with the online tool https://www.easyrdf.org/converter, you'll see they'll create new namespaces like 'ns0', 'ns1'...

simon-debruijn: @Stiksels I see, interesting. Thanks for the explanation!

PR #1596, opened 2024-01-30
III-6010 Handle Full update with unkown Organizer

JonasVHG:
### Added

- `GuardOrganizer`: trait to check if `Organizer` exists
-  new data files for Acceptance tests: `event-with-non-existing-organizer.json` & `place-with-non-existing-organizer.json`

### Changed

- `ImportEventRequestHandler` & `ImportPlaceRequestHandler`: Added `OrganizerDocumentRepository` in constructor & check if `Organizer` exists.
- `EventRequestHandlerServiceProvider` & `PlaceRequestHandlerServiceProvider`:  Injected `organizer_jsonld_repository` into `ImportEventRequestHandler` & `ImportPlaceRequestHandler`
- `features/event/create.feature`:  Added test for creating an `event` with an non-existing `Organizer`
- `features/place/create.feature`:  Added test for creating a `place` with an non-existing `Organizer`
- `ImportEventRequestHandlerTest.php` & `ImportPlaceRequestHandlerTest`:  Added Unit Tests for creating `Offers` with a non-existing `Organizer`

### Fixed

- One can no longer Create or Update `Offers` with a non-existing `Organizer`

---
Ticket: https://jira.publiq.be/browse/III-6010

### DIFF
--- a/src/Http/Event/ImportEventRequestHandler.php
+++ b/src/Http/Event/ImportEventRequestHandler.php
@@ -74,6 +74,7 @@ final class ImportEventRequestHandler implements RequestHandlerInterface
     private CommandBus $commandBus;
     private ImageCollectionFactory $imageCollectionFactory;
     private DocumentRepository $locationDocumentRepository;
+    private DocumentRepository $organizerDocumentRepository;
 
     public function __construct(
         Repository $aggregateRepository,
@@ -83,7 +84,8 @@ final class ImportEventRequestHandler implements RequestHandlerInterface
         RequestBodyParser $combinedRequestBodyParser,
         CommandBus $commandBus,
         ImageCollectionFactory $imageCollectionFactory,
-        DocumentRepository $locationDocumentRepository
+        DocumentRepository $locationDocumentRepository,
+        DocumentRepository $organizerDocumentRepository
     ) {
         $this->aggregateRepository = $aggregateRepository;
         $this->uuidGenerator = $uuidGenerator;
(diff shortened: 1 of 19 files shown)


src/Http/Event/ImportEventRequestHandler.php
```diff
@@ -270,7 +273,17 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         // which might cause race conditions if we're still dispatching other commands here as well.
         $organizerId = $eventAdapter->getOrganizerId();
         if ($organizerId) {
-            $commands[] = new UpdateOrganizer($eventId, $organizerId);
+            try {
+                $this->organizerDocumentRepository->fetch($organizerId);
+                $commands[] = new UpdateOrganizer($eventId, $organizerId);
+            } catch (DocumentDoesNotExist $e) {
```
LucWollants, line 279: I think that the `DocumentDoesNotExist` exception gets converted to a not found and then a 404 gets returned which is also fine in this context. So I don't think it's needed to catch the `DocumentDoesNotExist` and then convert it to `bodyInvalidData`

LucWollants: I gave this a quick try, but the resulting error is not so clear because it uses a general message `Document with id ... not found.`

LucWollants: Maybe to make it more readable put this in a `gaurdOrganization` method?

LucWollants (approved):

PR #1600, opened 2024-01-31
III-4601 Made format error message RFC7807 compliant for duplicate places

grubolsch:
### Changed
I Made format error message RFC7807 compliant for duplicate places
I now throw ApiProblem for this new format.

Minor bonus: changed all the status codes in ApiProblem to use constants.

Related: https://github.com/cultuurnet/apidocs/pull/359

---

### DIFF
--- a/features/place/duplicate.feature
+++ b/features/place/duplicate.feature
@@ -11,7 +11,7 @@ Feature: Test creating places
     Given I create a random name of 6 characters and keep it as "name"
     Given I create a minimal place and save the "id" as "originalPlaceId" then I should get a "201" response code
     Then I wait for the place with url "/places/%{originalPlaceId}" to be indexed
-    Given I create a minimal place and save the "query" as "query" then I should get a "409" response code
+    Given I create a minimal place then I should get a "409" response code
     Then the JSON response should be:
     """
     {
--- a/features/place/duplicate.feature
+++ b/features/place/duplicate.feature
@@ -15,7 +15,7 @@ Feature: Test creating places
     Then the JSON response should be:
     """
     {
-        "query": "%{baseUrl}/place/%{originalPlaceId}",
+        "duplicatePlaceUri": "%{baseUrl}/place/%{originalPlaceId}",
         "type": "https://api.publiq.be/probs/url/status-conflict",
         "title": "Status conflict",
         "status": 409,
--- a/features/place/duplicate.feature
+++ b/features/place/duplicate.feature
@@ -19,7 +19,7 @@ Feature: Test creating places
         "type": "https://api.publiq.be/probs/url/status-conflict",
         "title": "Status conflict",
         "status": 409,
-        "detail": "This place already exists. Use the attached query to get existing place(s) for the place you tried to create."
+        "detail": "A place with this address / name combination already exists. Please use the existing place for your purposes."
     }
     """
     Then I restore the duplicate configuration
--- a/features/place/duplicate.feature
+++ b/features/place/duplicate.feature
@@ -11,9 +11,15 @@ Feature: Test creating places
     Given I create a random name of 6 characters and keep it as "name"
     Given I create a minimal place and save the "id" as "originalPlaceId" then I should get a "201" response code
     Then I wait for the place with url "/places/%{originalPlaceId}" to be indexed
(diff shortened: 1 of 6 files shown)


features/place/duplicate.feature
```diff
@@ -11,9 +11,15 @@ Feature: Test creating places
     Given I create a random name of 6 characters and keep it as "name"
     Given I create a minimal place and save the "id" as "originalPlaceId" then I should get a "201" response code
     Then I wait for the place with url "/places/%{originalPlaceId}" to be indexed
-    Given I create a minimal place and save the "originalPlace" as "newPlaceUri" then I should get a "409" response code
-    Then the JSON response at "originalPlace" should be:
+    Given I create a minimal place and save the "query" as "query" then I should get a "409" response code
```
LucWollants, line 14: Is it still needed to save this `query` value? I get the feeling that we could do without this extra custom step by just calling the existing steps like, for example:

```
When I send a POST request to "/places/"
Then the response status should be "409"
And I keep the value of the JSON response at "id" as "originalPlaceId"
```

src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -283,4 +263,26 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+        if (! $this->preventDuplicatePlaces) {
+            return;
+        }
+
+        try {
+            $duplicatePlaceId = $this->lookupDuplicatePlace->getDuplicatePlaceUri($place);
+            if ($duplicatePlaceId !== null) {
+                throw ApiProblem::statusConflict(
+                    MultipleDuplicatePlacesFound::ERROR_MSG,
+                    ['query' => $duplicatePlaceId]
```
LucWollants, line 278: When there is only one duplicate place, is it then still a query?

tests/Http/Place/ImportPlaceRequestHandlerTest.php
```diff
@@ -5197,55 +5205,66 @@ public function testHandleWithDuplicatePreventionEnabled(Repository $aggregateRe
             ->withJsonBodyFromArray($this->getSimplePlace())
             ->build('POST');
 
+        $this->expectExceptionObject(ApiProblem::statusConflict(
+            MultipleDuplicatePlacesFound::ERROR_MSG,
+            ['instance' => $originalPlaceUri]
```
LucWollants, line 5210: Here `instance` is still used but the guard throws `query`, which makes me think there is still something else going on.

src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -283,4 +263,26 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+    {
+        if (! $this->preventDuplicatePlaces) {
+            return;
+        }
+
+        try {
+            $duplicatePlaceId = $this->lookupDuplicatePlace->getDuplicatePlaceUri($place);
+            if ($duplicatePlaceId !== null) {
+                throw ApiProblem::statusConflict(
+                    MultipleDuplicatePlacesFound::ERROR_MSG,
```
LucWollants, line 277: It feels weird to get the error message from the `MultipleDuplicatePlacesFound` although there is exactly one place returned as duplicate.

src/Http/Place/ImportPlaceRequestHandler.php
grubolsch: What would you prefer as the param name?

LucWollants: Something like the parameter and method name, maybe`duplicatePlaceUri`?

tests/Http/Place/ImportPlaceRequestHandlerTest.php
LucWollants: This is not verified with `expectExceptionObject` I think it is better to use `AssertApiProblemTrait`

grubolsch: done

src/Http/Place/ImportPlaceRequestHandler.php
grubolsch: changed it

LucWollants (approved):

PR #1608, opened 2024-02-06
III-6047 Implement completeness on organizer

LucWollants:
### Added
- Implemented completeness on organizer

### Note
This pull request is required for the config https://github.com/cultuurnet/appconfig/pull/557

---
Ticket: https://jira.uitdatabank.be/browse/III-6047


app/Organizer/OrganizerJSONLDServiceProvider.php
```diff
@@ -55,7 +55,8 @@ function () use ($container) {
                     ),
                     new CdbXMLImporter(
                         $container->get(CdbXMLToJsonLDLabelImporter::class)
-                    )
+                    ),
+                    $container->get('config')['completeness']['organizer']
```
JonasVHG, line 59: Don't know about this one,
On the one hand you provide a fallback with an empty array in the config: 
`file_exists(__DIR__ . '/config.completeness.php') ? require __DIR__ . '/config.completeness.php' : [],
`
But if the empty array doesn't contain `organizer`, it could still fail because you get a null. Though I still doubt whether it would be good to provide another fallback or fall hard

JonasVHG (commented):
I have some doubts about the `$container->get('config')['completeness']['organizer']` 
If `organizer` doesn't exist(or a typo was introduced), it will return a `null` instead of an empty array. While such a thing should still be catched during a PR, the mixing of empty arrays as fallback & the missing empty array here could introduce unexpected behaviour.

src/Organizer/OrganizerLDProjector.php
```diff
@@ -687,4 +693,29 @@ private function updateModified(JsonDocument $jsonDocument, DomainMessage $domai
...
+
+        $completeness = 0;
+        foreach ($this->weights as $key => $weight) {
+            if (isset($body[$key])) {
+                if ($key === 'contactPoint' && $this->isContactPointEmpty($body[$key])) {
+                    continue;
+                }
+
+                $completeness += $weight;
+            }
```
simon-debruijn, line 709:
```suggestion
            if (!isset($body[$key])) {
                continue;
            }
            
            if ($key === 'contactPoint' && $this->isContactPointEmpty($body[$key])) {
                continue;
            }

           $completeness += $weight;
```

app/Organizer/OrganizerJSONLDServiceProvider.php
simon-debruijn: I'm leaning towards expecting there will always be a completeness config for organizer. So I would not fallback with an empty array, but fail hard instead.

simon-debruijn (approved):

app/Organizer/OrganizerJSONLDServiceProvider.php
LucWollants: See also the discussion here https://github.com/cultuurnet/appconfig/pull/557, I think it's indeed much better to fail hard instead of having unexpected side effects.

JonasVHG: I agree with the fail hard, just because we always failed softly doesn't mean we should continue with that approach.
This was just a special case where failing soft & failing hard were mixed.

JonasVHG (approved):

PR #1609, opened 2024-02-08
III-6047 Calculate completeness score for places and events

LucWollants:
### Added
- Calculated completeness score for places and events

---
Ticket: https://jira.uitdatabank.be/browse/III-6047

### DIFF
--- a/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
+++ b/src/Offer/ReadModel/JSONLD/OfferLDProjector.php
@@ -100,6 +100,8 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
 
     protected VideoNormalizer $videoNormalizer;
 
+    private array $weights;
+
     private ?int $playhead = null;
 
     /**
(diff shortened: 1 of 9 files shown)


src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -100,6 +100,8 @@ abstract class OfferLDProjector implements OrganizerServiceInterface
 
     protected VideoNormalizer $videoNormalizer;
 
+    private array $weights;
```
grubolsch, line 103: Can we add a docblock here what type of array this is? I assume an array of int?

grubolsch (approved):

src/Offer/ReadModel/JSONLD/OfferLDProjector.php
```diff
@@ -123,6 +126,7 @@ public function __construct(
         $this->mediaObjectSerializer = $mediaObjectSerializer;
         $this->basePriceTranslations = $basePriceTranslations;
         $this->videoNormalizer = $videoNormalizer;
+        $this->weights = $weights;
```
grubolsch, line 129: Given that this should be an array of integers, should we either check the inserted weights, or cast them all to integers?

`$this->weights = array_map('intval', $weights);`

LucWollants: I will create a class `Weights` that throws on wrong input types. I will do this in a separate pull request.

PR #1611, opened 2024-02-08
III-5912 Make Title a string in EventCreated

grubolsch:
### Changed
III-5912 Make Title a string in EventCreated

This is to prepare for the title length limit in the commands later.

---

### DIFF
--- a/src/Address/Parser/GeopuntAddressParser.php
+++ b/src/Address/Parser/GeopuntAddressParser.php
@@ -68,7 +68,12 @@ final class GeopuntAddressParser implements AddressParser, LoggerAwareInterface
                     'items' => [
                         'type' => 'object',
                         'properties' => [
-                            'Municipality' => ['type' => 'string'],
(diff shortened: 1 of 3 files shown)


tests/Event/EventCommandHandlerTest.php
```diff
@@ -99,7 +99,7 @@ public function it_should_create_a_new_event(): void
         $this->scenario
             ->withAggregateId($id)
             ->when($command)
-            ->then([new EventCreated($id, $language, $title, $type, $location, $calendar, $theme, $now)]);
+            ->then([new EventCreated($id, $language, $title->toString(), $type, $location, $calendar, $theme, $now)]);
```
LucWollants, line 102: Could make sense here to change the construction of the $`title` to be a primitive instead of the`Title` class?

LucWollants (approved):
The code changes look good. I did not run any acceptance tests.

PR #1613, opened 2024-02-09
III-5912 event major info updated make title string

grubolsch:
### Changed
For the event majorInfoUpdated , changed LegacyTitle to string to prepare for length limitation.

---


tests/Event/EventCommandHandlerTest.php
```diff
@@ -141,7 +141,7 @@ public function it_can_update_major_info_of_an_event(): void
             ->when(
                 new UpdateMajorInfo($id, $title, $eventType, $location, $calendar)
             )
-            ->then([new MajorInfoUpdated($id, $title, $eventType, $location, $calendar)]);
+            ->then([new MajorInfoUpdated($id, $title->toString(), $eventType, $location, $calendar)]);
```
LucWollants, line 144: I think the constructor can be changed for `$title` to a primitive string instead of the `Title` class

LucWollants (approved):
Nice work

PR #1619, opened 2024-02-12
III-5912 AbstractTitleUpdated - changed LegacyTitle to string

grubolsch:
### Changed
AbstractTitleUpdated - changed LegacyTitle to string

---

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -587,7 +587,7 @@ final class Event extends Offer
 
     protected function createTitleUpdatedEvent(Title $title): TitleUpdated
     {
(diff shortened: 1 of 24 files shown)


src/Event/Event.php
```diff
@@ -587,7 +587,7 @@ protected function createTitleTranslatedEvent(LegacyLanguage $language, Title $t
 
     protected function createTitleUpdatedEvent(Title $title): TitleUpdated
```
LucWollants, line 588: Does it make sense to change the interface of this method and change the type of `$title` to `string`?

LucWollants (approved):
Only a minor remark so I already approved.

PR #1620, opened 2024-02-12
III-5912 change legacy title to new title for create event

grubolsch:
### Changed
I changed the CreateEvent / CreatePlace command to use the new Title object, instead of the LegacyTitle object. 
This in preparation of the length-restriction.

### Removed
I see that the abstract class AbstractCreateCommand is only used for the class CreateEvent
I remove the abstract class, to simpiify the code.
I assume it used to also extend to the CreatePlace command, but this is no longer the case.

---


src/Http/Place/ImportPlaceRequestHandler.php
```diff
@@ -161,7 +161,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             $placeAggregate = PlaceAggregate::create(
                 $placeId,
                 $mainLanguage,
-                $title,
+                new Title($title->toString()),
```
LucWollants, line 164: This seems weird, is this a conversion between the same types?

LucWollants (approved):
Only one small question, so I already approved.

src/Http/Place/ImportPlaceRequestHandler.php
grubolsch: it is not, its a conversion between the old LegacyTitle and the new Title.

I will merge this now, but this will be fixed in the next PR I am going to create in a minute, scope wise I thought it wa s best to do this in a seperate PR.

PR #1624, opened 2024-02-13
III-5912 add length check trait

grubolsch:
### Added
Added a trait to check for length in the title.
The change in the Title itself is in a different PR, so so far this "does nothing".

---


src/Model/ValueObject/String/Behaviour/HasMaxLength.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Model\ValueObject\String\Behaviour;
+
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Http\ApiProblem\SchemaError;
+
+trait HasMaxLength
+{
+    private function hasMaxLength(string $value, int $maxLength, string $path='/'): void
```
LucWollants, line 12: Other behaviors use a `guard` method, for example `guardNotEmpty`

src/Model/ValueObject/String/Behaviour/HasMaxLength.php
```diff
@@ -0,0 +1,20 @@
...
+    {
+        if (mb_strlen($value) > $maxLength) {
+            throw ApiProblem::bodyInvalidData(
```
LucWollants, line 15: This should not throw an `ApiProblem` because it is not inside the HTTP context, but a `InvalidArgumentException`, also see `guardNotEmpty`.

tests/Model/ValueObject/String/Behaviour/MockHasMaxLength.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Model\ValueObject\String\Behaviour;
+
+class MockHasMaxLength
```
LucWollants, line 7: Check how `MockNotEmptyString` handles this

tests/Model/ValueObject/String/Behaviour/MockHasMaxLength.php
```diff
@@ -0,0 +1,24 @@
...
+    private bool $success;
+
+    public function __construct(string $value, int $maxLength)
+    {
+        $this->hasMaxLength($value, $maxLength);
+
+        $this->success = true;
+    }
+
+    public function isSuccess(): bool
```
LucWollants, line 20: There is no need for this, the throw of the exception is what should be checked.

tests/Model/ValueObject/String/Behaviour/MockHasMaxLengthTest.php
```diff
@@ -0,0 +1,35 @@
...
+    public function testMaxLengthValidation(string $value, int $maxLength, bool $expectException): void
+    {
+        if ($expectException) {
+            $this->expectException(ApiProblem::class);
+        }
+
+        $mock = new MockHasMaxLength($value, $maxLength);
+
+        // Check for the happy path
+        $this->assertTrue($mock->isSuccess());
```
LucWollants, line 24: This is a side-effect that is only available on the Mock and not in a real scenario were the trait would be used.

LucWollants (changes requested):
I would mimic `IsNotEmpty` more

tests/Model/ValueObject/String/Behaviour/MockHasMaxLengthTest.php
grubolsch: I know, but I did not know how else to check the happy patchs, because they don't return anything. I could make hasMaxLength() always make return a bool if you prefer? Or if you have another solution?

tests/Model/ValueObject/String/Behaviour/MockHasMaxLength.php
grubolsch: Looks the same to me, how would you prefer it? Something like "MockNotTooLong" ?

src/Model/ValueObject/String/Behaviour/HasMaxLength.php
grubolsch: I first did that, but the thing is, if I do it that way I will require a try/catch on every command-handler where this is used for a clean error message.
Now I get all of that for free out of the box. 

Is this not cleaner this way?

LucWollants: There is code that catches general or specific exceptions and converts them to an `ApiProblem` to indeed avoid this `try/catch` combination. The code for this can be found inside `ApiProblemFactory::createFromThrowable`. So it's possible to throw an exception here, otherwise all the other guards from the behaviors would have the same issue.

tests/Model/ValueObject/String/Behaviour/MockHasMaxLength.php
LucWollants: I guess it is `MockHasMaxLengthString`

tests/Model/ValueObject/String/Behaviour/MockHasMaxLengthTest.php
LucWollants: Assert the value from the string, like `it_should_throw_an_exception_if_an_empty_string_is_given`

src/Model/ValueObject/String/Behaviour/HasMaxLength.php
grubolsch: I made it throw a custom Exception that I am catching in ApiProblemFactory like discussed

grubolsch:
Feedback processed

src/Model/ValueObject/String/Behaviour/HasMaxLength.php
```diff
@@ -4,14 +4,14 @@
...
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\Exception\MaxLengthExceeded;
 
 trait HasMaxLength
 {
-    private function guardTooLong(string $value, int $maxLength, string $path='/'): void
+    private function guardTooLong(string $value, int $maxLength): void
     {
         if (mb_strlen($value) > $maxLength) {
-            throw new InvalidArgumentException(sprintf('Given string should not be longer than %d characters.', $maxLength));
+            throw MaxLengthExceeded::maxLengthExceeded($maxLength);
```
LucWollants, line 14: I think it is fine to use the constructor here, that would also be more logical to read?

src/Model/ValueObject/String/Behaviour/Exception/MaxLengthExceeded.php
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public static function maxLengthExceeded(int $maxLength): self
+    {
+        return new self(sprintf('Given string should not be longer than %d characters.', $maxLength));
+    }
+
+    public function getPath(): string
+    {
+        return '/0/title';
```
LucWollants, line 23: This makes the assumption that this trait is used inside the `Title` value object, but it is possible to use it also in the `CopyrightHolder` so then the path would be wrong.

I would not worry too much about the "correct" exposure of this error to the API layer, that will be done with the JSON schema rules.

src/Model/ValueObject/String/Behaviour/HasMaxLength.php
grubolsch: I always prefer to make the constructor private for custom exceptions, and put the error message (the string) inside the exception itself. 
This gives a couple of benefits:

- Gives a nice summary of all possible errors this Exception is used
- Gives another developer the reuse the exact the same message, and not have 5 variants of the same error floating in the code
- Forced any developer to make a new type of error explicit

This is also a recommended way i books I have read about this subject.

So, I would prefer to keep it this way if it is oke for you?

LucWollants: Fine for me to keep it. But you could also change the constructor to only accept the length and use the fixed string inside the constructor.

I think the examples you are talking about are:

```
EventCannotBeAddedToProduction::becauseItAlreadyBelongsToAnotherProduction
EventCannotBeAddedToProduction::becauseSomeEventsBelongToAnotherProduction
```

In the current implementation, there is only one flavor, and something like `LengthIsInvalid::becauseMaxLengthIsExceeded` could work which reads more logically than `becauseSomeEventsBelongToAnotherProduction`

PR #1626, opened 2024-02-15
Just a handy script to copy config from apidocs to UDB3 codebase

grubolsch:
### Added
I added a small script in the bin/ dir to copy the json schemas from the apidocs repo to our codebase, handy for local development.
The script does assume apidocs & udb3-backend share the same root directory.


LucWollants:
The approach we take is:
- create a feature branch on the package that needs to be changed
- depend on the feature branch of the package
- run a `composer update` to get the feature branch

PR #1630, opened 2024-02-22
III-6098 Add purge command for `*permission_readmodel`

JonasVHG:
### Changed

- `ReplayCommand`: Add purge for `events`, `places` & `organizers` when replaying a item.

### Fixed

- `Offers` & `Organizers` can now be replayed without throwing an `UniqueConstraintException`

### Note:
I have also checked `app/Console/Command/PurgeModelCommand.php`, but that still seems up-to-date.

---
Ticket: https://jira.publiq.be/browse/III-6098

### DIFF
--- a/app/Console/Command/ReplayCommand.php
+++ b/app/Console/Command/ReplayCommand.php
@@ -132,7 +132,7 @@ final class ReplayCommand extends AbstractCommand
 
         if ($cdbids !== null) {
             foreach ($cdbids as $cdbid) {
-                $this->purgePermissionReadmodels($cdbid);
+                $this->purgeReadmodels($cdbid);
             }
         }
 
@@ -246,7 +246,7 @@ final class ReplayCommand extends AbstractCommand
         return $eventStream;
     }
 
-    private function purgePermissionReadmodels(string $cdbid): void
+    private function purgeReadmodels(string $cdbid): void
     {
         foreach (self::TABLES_TO_PURGE as $tableName => $columnName) {
             $this->connection->delete(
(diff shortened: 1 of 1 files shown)


grubolsch (approved):

grubolsch:
Would it be a good idea to also purge the entire table if we do a replay without a specific id?
Now the bug would still trigger if you would do a general replay.

LucWollants (approved):

app/Console/Command/ReplayCommand.php
```diff
@@ -121,6 +130,12 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $cdbids = $input->getOption(self::OPTION_CDBID);
 
+        if ($cdbids !== null) {
+            foreach ($cdbids as $cdbid) {
```
LucWollants, line 134: This could be an issue when we have multiple cdbids, then we purge everything instead of doing it one by one before the replay starts for one item.

LucWollants: I will try to explain the issue with an example:
- A user provides the cdbids 1, 2, 3, 4 and 5
- The scripts starts and purges 1 till 5
- The script will now publish the event streams
- There is an issue and the script fails on id 2
- The ids 3, 4 and 5 are already purged but not replayed
- So we are in a situation where the state is worse then before

JonasVHG: Indeed, after discussing this in Slack we decided that we will only purge the first item because this Command is mostly run by helpdesk via Jenkins, and in Jenkins it is limited to 1 cdbid.

LucWollants (approved):

LucWollants:
I approved, but now I wonder if we need to validate the user input. What happens if I enter a `*` as the cdbid, will all rows be deleted?

JonasVHG:
> I approved, but now I wonder if we need to validate the user input. What happens if I enter a `*` as the cdbid, will all rows be deleted?

As discussed in Slack neither `*` nor `%` delete anything, so this shouldn't be an issue.

PR #1631, opened 2024-02-22
III-5912 Remove Title from Apply methods

grubolsch:
### Changed
Removed the typed Title from all apply() methods, instead working with string.

### How I tested?
- Created an org/place/event with a very long title. (disabled length check)
- Enable length check
- Try to update org/place/event -> success!

---


LucWollants (approved):
Only 2 small remarks so I already approved.

PR #1632, opened 2024-02-22
III-6094 Update place geocode command

grubolsch:
### Changed
Added an option called "all", that forces the script to update all entities geocordinates, instead of only the ones with missing coordinates.

The following commands have changed:
- place:geocode
- event:geocode
- organizer:geocode

---
Ticket: https://jira.uitdatabank.be/browse/III-6094

### DIFF
--- a/app/Console/Command/AbstractGeocodeCommand.php
+++ b/app/Console/Command/AbstractGeocodeCommand.php
@@ -19,6 +19,8 @@ use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 abstract class AbstractGeocodeCommand extends AbstractCommand
 {
+    private const OPT_ALL = 'all';
+
     private ResultsGeneratorInterface $searchResultsGenerator;
 
     private DocumentRepository $documentRepository;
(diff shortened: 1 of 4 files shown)


LucWollants (approved):

JonasVHG (commented):
I wonder if we should not add more logic to the `all` option?
What if the command crashes halfway, would we have to restart from the beginning?
I think it might be an improved if we could at a query to run the `all` option in batches.
Best to discuss this with who will run it on production.

grubolsch:
> I wonder if we should not add more logic to the `all` option? What if the command crashes halfway, would we have to restart from the beginning? I think it might be an improved if we could at a query to run the `all` option in batches. Best to discuss this with who will run it on production.

Good point!
I could enter a start and limit, if I read the SAPI3 docs that is an option? https://docs.publiq.be/docs/uitdatabank/search-api/pagination

Maybe I could keep the current start in redis and add a restart option to start from the last start-pointer.
What do you think? @LucWollants @JonasVHG

LucWollants:
As discussed through Slack, the process to fix all places should only be run once. Because of this one-time nature and the potential risk of failing midway, it seems better to use a list of IDs to feed this script. Something similar has already been done by @willaerk for putting messages on AMQP.

PR #1642, opened 2024-02-28
III-6109 Make batchscript to loop over other script

grubolsch:
### Added
I made a batchscript that accepts a csv, loops over it and then runs a different shell command for each row in the file.
It is possible to pass columns as arguments by using a placeholder %1, %2, ...

Example:   bin/udb3.php execute-command-from-csv input.csv "place:delete %1 %2"

### Why?
Idea is for duplicate places to take the CSV from Frederic and delete all places in clusters (except the canonical), and we are going to do this one at a time. So other command "place:delete".
I did write this so it can be reused for different commands.

---
Ticket: https://jira.uitdatabank.be/browse/III-6109

### DIFF
--- a/app/Console/Command/ExecuteCommandFromCsv.php
+++ b/app/Console/Command/ExecuteCommandFromCsv.php
@@ -8,6 +8,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Process\Process;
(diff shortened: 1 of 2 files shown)


app/Console/Command/ExecuteCommandFromCsv.php
```diff
@@ -0,0 +1,80 @@
...
+            $output->writeln(sprintf("<error>Could not find file '%s'</error>", $file));
+            return 0;
+        }
+
+        $handle = fopen($file, 'rb');
+
+        $progressBar = new ProgressBar($output, count(file($file)));
+        $progressBar->start();
+
+        $firstRow = true;
```
LucWollants, line 43: Maybe make it an option that the CSV file has a header or not?

grubolsch: done!

LucWollants (approved):

PR #1643, opened 2024-02-28
III-6069 Add endpoint to request ownership

LucWollants:
### Added
- Added a new endpoint `POST /ownerships` to request an ownership on a certain item of a certain type.
- Added a new endpoint `GET /ownerships/{ownershipId}` to get an ownership by id

### Note
- Some missing features for a follow-up pull request
  - validation (are all fields correct)
  - authorization (can this user request)
  - state (what is the state of the request)
  - add Stoplight docs

---
Ticket: https://jira.uitdatabank.be/browse/III-6069


app/Ownership/OwnershipServiceProvider.php
```diff
@@ -0,0 +1,53 @@
...
+                $container->get('event_store_factory')(AggregateType::ownership()),
+                $container->get(EventBus::class),
+                [
+                    $container->get('event_stream_metadata_enricher'),
+                ]
+            )
+        );
+
+        $container->addShared(
+            'ownership_jsonld_repository',
```
grubolsch, line 40: Can we get this as a constant somewhere? Either on the repo class or the OwnershipServiceProvider.
I want to avoid more strings as dependency keys in our code base.

src/Model/ValueObject/Identity/UserId.php
```diff
@@ -0,0 +1,20 @@
...
+use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\IsString;
+
+final class UserId
+{
+    use IsString;
+    use IsNotEmpty;
+
+    public function __construct(string $value)
+    {
+        $this->guardNotEmpty($value);
```
grubolsch, line 17: Would it be a good idea to add some regex here to check if it has a valid structure of an id?

src/Ownership/OwnershipRepository.php
```diff
@@ -0,0 +1,34 @@
...
+            Ownership::class,
+            new PublicConstructorAggregateFactory(),
+            $eventStreamDecorators
+        );
+    }
+
+    public function load($id): Ownership
+    {
+        /** @var Ownership $ownership */
+        $ownership = parent::load($id);
```
grubolsch, line 31: Maybe add an instanceof check here, to make sure it is an Ownership? You can pass an id to it, so you never know when somebody makes a mistake.
You can then drop the @var comment.

src/Http/Ownership/GetOwnershipRequestHandler.php
```diff
@@ -0,0 +1,38 @@
...
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $ownershipId = $routeParameters->getOwnershipId();
+
+        try {
+            return new JsonLdResponse(
+                $this->ownershipRepository->fetch($ownershipId)->getRawBody()
+            );
+        } catch (DocumentDoesNotExist $exception) {
```
grubolsch, line 34: You could add an @todo here that once we upgrade to PHP8, we can remove the exception variable.

grubolsch (changes requested):

src/Model/ValueObject/Identity/UserId.php
LucWollants: I think that could become a difficult regex because we have several formats to support, for example:
- `google-oauth2|102486314601596809843`
- `auth0|813ff497-f0b8-4a59-bd44-dc2646ad98ff`
- `auth0|631748dba64ea78e3983b207`
- `9e68dafc-01d8-4c1c-9612-599c918b981d`

But I could create some basic if statements to check prefix and/or uuid?

src/Http/Ownership/GetOwnershipRequestHandler.php
LucWollants: Nice suggestion, but I think we can do that with a tool like https://github.com/rectorphp/rector for all occurrences

src/Model/ValueObject/Identity/UserId.php
simon-debruijn: If we add validation I would opt for if checks for the different prefixes. To support these different patterns a regex would be to fragile.

grubolsch: maybe check either if it is in a list of allowed prefixes
and then if nothing else if it is a regex?

JonasVHG: The problem is that the list of allowed prefixes is not static, we have `auth0`, `google-oauth2`, `facebook`, `apple`, `csam` in the past we also had `linkedin`. In the future, new services might be added, so just a valid string seems the logical approach to me.

LucWollants: Good to know this! But I guess this can be something we expose somewhere while doing the migration to Keycloack?

grubolsch (approved):

PR #1650, opened 2024-03-06
Fix geocoding command when place has no address

grubolsch:
### Changed
There is a minor bug with the geocoding event, some places do not have an address and the script throws an ugly error about this.
I now catch this in a more elegant solution.

### DIFF
--- a/app/Console/Command/GeocodePlaceCommand.php
+++ b/app/Console/Command/GeocodePlaceCommand.php
@@ -34,11 +34,16 @@ class GeocodePlaceCommand extends AbstractGeocodeCommand
 
         $jsonLd = Json::decodeAssociatively($document->getRawBody());
 
+        if (!isset($jsonLd['address'])) {
+            $output->writeln("Skipping {$placeId}. (JSON-LD does not contain an address.)");
+            return;
+        }
+
         $addressLanguage = $jsonLd->mainLanguage ?? 'nl';
 
         if (!isset($jsonLd['address'][$addressLanguage])) {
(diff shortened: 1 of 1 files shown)


app/Console/Command/GeocodePlaceCommand.php
```diff
@@ -38,7 +38,7 @@ protected function dispatchGeocodingCommand(string $placeId, OutputInterface $ou
 
         if (!isset($jsonLd['address'][$addressLanguage])) {
             // Some places have an address in another language then the main language or `nl`
-            $addressLanguage = array_key_first($jsonLd['address']);
+            $addressLanguage = array_key_first($jsonLd['address'] ?: []);
```
LucWollants, line 41: Not sure if this will fix this error https://publiq-vzw.sentry.io/issues/5027675756/?project=5435528&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=3

grubolsch: You are right. It solved the fatal error, but there is still a warning. The check with ?? is better here, that supresses both errors.

LucWollants: For me it feels more logical to add an early return for when the `address` property is missing. When using the query it is not possible to have a missing address, now places can have a missing address.

```
if (!isset($jsonLd['address'])) {
 // Log missing address
 // Do early return
}
```

LucWollants (approved):

PR #1651, opened 2024-03-06
III-6117 Add id to response create saved searches

grubolsch:
### Changed
I changed the output of the Saved Searches endpoint.
It used to return a NULL, which I found very confusing and not consistent with most of our create statements.
(I thought that the script just failed the first time I called it.)

I refactored the endpoint to now return the id of the created saved search.

I also need this for a feature test for the next feature I am creating, a PUT command for Saved Searches.

Screenshot 2024-03-06 at 12 43 58

---
Ticket: https://jira.uitdatabank.be/browse/III-6117


LucWollants (approved):
Nice work!

Make sure to add a note about the changed response so the UI can also be tested.

One thing I wonder about are the errors inside `app/SavedSearches/SavedSearchesServiceProvider.php`, does this not result in an error when running the acceptance tests?

simon-debruijn (approved):

PR #1652, opened 2024-03-06
III-6069 Add check ownership already exists

LucWollants:
### Changed
- Prevent creating ownership for the given item and given owner when this ownership already exists

### Added
- Added projector that creates an "index" inside a MySQL table with owner and item

### Note
- The ownership inside the search table needs to be cleared when an ownership gets rejected to make sure it can be requested again
- Follow-up validation in next pull requests
  - Check if the item exists, see https://github.com/cultuurnet/udb3-backend/pull/1653
  - Check if the current user has permission to request ownership, see https://github.com/cultuurnet/udb3-backend/pull/1654
  - Check if the owner exists

---
Ticket: https://jira.uitdatabank.be/browse/III-6069


src/Http/Ownership/RequestOwnershipRequestHandler.php
```diff
@@ -51,6 +57,22 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
         $requestOwnership = $requestBodyParser->parse($request)->getParsedBody();
 
+        // Make sure there is no open request for this item and owner
+        try {
+            $ownershipItem = $this->ownershipSearchRepository->getByItemIdAndOwnerId(
+                $requestOwnership->getItemId()->toString(),
+                $requestOwnership->getOwnerId()->toString()
+            );
+        } catch (OwnershipItemNotFound $e) {
+            $ownershipItem = null;
```
grubolsch, line 67: Why the extra step of putting it to null and then throwing an error? getByItemIdAndOwnerId() can never return a null, so this is the only case the check below could trigger.

grubolsch (changes requested):

JonasVHG (approved):

src/Http/Ownership/RequestOwnershipRequestHandler.php
LucWollants: Indeed, this flow can be improved, see these changes https://github.com/cultuurnet/udb3-backend/pull/1652/commits/5f1c3f1b67d093ccb7f391f126cc114194466ed7

grubolsch (approved):

PR #1657, opened 2024-03-11
III-6117 BUGFIX put saved search gave error when saving twice

grubolsch:
### Fixed
Fixed a bug with the PUT of saved searches.
If you saved a saved search without modifying any value, it would throw an `ApiProblem::savedSearchNotFound()`.

This is because I listened to the "affected rows", which will return 0 if the update does not have to change anything in the db.

I now use a proper select query.

I provided an extra feature test to prove the bug is fixed.

---
Ticket: https://jira.uitdatabank.be/browse/III-6117

### DIFF
--- a/features/saved_searches/update.feature
+++ b/features/saved_searches/update.feature
@@ -33,8 +33,33 @@ Feature: Test the UDB3 saved searches API
 
     When I send a GET request to "/saved-searches/v3"
     Then the response status should be "200"
-    And the response body should be valid JSON
+#    And the response body should be valid JSON
     And the JSON response should include:
       """
       {"name":"%{name}","query":"Lessen in de avond","id":"%{id}"}
       """
+
+  #Bugfix https://jira.publiq.be/browse/III-6019
+  Scenario: I update a saved search twice
+    Given I create a random name of 12 characters
+    And I set the JSON request payload to:
+       """
+       {"name":"This should change","query":"Lessen in de ochtend"}
+       """
+    When I send a POST request to "/saved-searches/v3/"
+    Then the response status should be "201"
+    And I keep the value of the JSON response at "id" as "id"
+
+    And I set the JSON request payload to:
+       """
+       {"name":"%{name}","query":"Lessen in de avond"}
+       """
+    When I send a PUT request to "/saved-searches/v3/%{id}"
+    Then the response status should be "204"
+
+    And I set the JSON request payload to:
+       """
+       {"name":"%{name}","query":"Lessen in de avond"}
+       """
+    When I send a PUT request to "/saved-searches/v3/%{id}"
+    Then the response status should be "204"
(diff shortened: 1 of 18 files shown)


JonasVHG (approved):

src/SavedSearches/WriteModel/SavedSearchRepositoryInterface.php
```diff
@@ -20,10 +21,12 @@ public function update(
...
         QueryString $queryString
-    ): int;
+    ): void;
 
     public function delete(
         string $userId,
         string $searchId
     ): void;
+
+    public function findSavedSearchOwnedByCurrentUser(string $id): ?SavedSearch;
```
LucWollants, line 31: This interface is confusing me, all the other interfaces pass in an `id` and a `userId`. I think it's best to stick with that approach and change the method to `findByIdAndUserId`.

I would also remove the `userId` property from the constructor of `UDB3SavedSearchRepository`

features/saved_searches/update.feature
```diff
@@ -33,8 +33,33 @@ Feature: Test the UDB3 saved searches API
...
+#    And the response body should be valid JSON
     And the JSON response should include:
       """
       {"name":"%{name}","query":"Lessen in de avond","id":"%{id}"}
-      """
\ No newline at end of file
+      """
+
+  #Bugfix https://jira.publiq.be/browse/III-6019
+  Scenario: I update a saved search twice
```
LucWollants, line 43: This makes me wonder if we need to make a test where one user would be updating the saved search of another user?

src/SavedSearches/WriteModel/SavedSearchRepositoryInterface.php
LucWollants: I now see that this new method is inside the wrong interface, there is a read and a write interface. This method should be part of `CultuurNet\UDB3\SavedSearches\ReadModel\SavedSearchRepositoryInterface`

LucWollants (changes requested):
The new method is not inside the correct interface, it should be added to `CultuurNet\UDB3\SavedSearches\ReadModel\SavedSearchRepositoryInterface`

src/SavedSearches/WriteModel/SavedSearchRepositoryInterface.php
grubolsch: Moved it to a separate SavedSearchReadRepository class.

features/saved_searches/update.feature
grubolsch: Added a test for this, and also for when the saved search does not exist

grubolsch:
I processed all feedback, and added some new feature / unit tests.
Also rerun all features test related to saved searches.

LucWollants (approved):
Nice work!

PR #1658, opened 2024-03-11
[opinionated] Improve performance - do not use array_merge in loop

grubolsch:
### Changed
I want to remove array_merge in a loop, because it is very inefficent. Array_merge always copies the entire array in memory instead of modifying the existing one, making it expensive to run in a loop.

I suggest an old school loop to improve the performance.


LucWollants (approved):
Fine for me to change it but this is really a micro optimisation because the amount of data that is processed. I even think that this is now slower because only one copy is needed 😏

PR #1660, opened 2024-03-18
III-6120 Clean place names before sending api call

grubolsch:
### Added
Clean place names to before sending them to google maps based on the following rules:

- Lowercase all letters
- Remove dots
- Remove accents
- Replace these symbols ["'?!()&_,:] by a white space
- Remove duplicate words
- Remove city name out of location name

---
Ticket: https://jira.uitdatabank.be/browse/III-6120

### DIFF
--- a/src/Place/ReadModel/Duplicate/CleanPlaceName.php
+++ b/src/Place/ReadModel/Duplicate/CleanPlaceName.php
@@ -1,0 +1,258 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\ReadModel\Duplicate;
+
+use CultuurNet\UDB3\Model\Place\Place;
+use CultuurNet\UDB3\Model\ValueObject\Geography\Address;
+
+/**
+ * This class converts place names (title) with the following rules:
+ * Lowercase all letters
+ * Remove dots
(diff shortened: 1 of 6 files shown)


src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,251 @@
...
+ * */
+class CleanPlaceName
+{
+    public static function transform(Address $address, string $title): string
+    {
+        $title = str_replace(['.', ''], '', mb_strtolower(trim($title)));
+
+        $title = self::removeAccents($title);
+
+        $title = str_replace(['"', '\'', '?', '!', '(', ')', '&', ',', '_', ':'], ' ', $title);
```
JonasVHG, line 26: I know the ticket only mentioned these characters. But is there a reason why `#`, `$`, `%`, ...  should not be removed? Maybe discuss this with @corneelwille 
If so `$title = preg_replace('/[^a-zA-Z0-9]+/', ' ', $title);` might be a more stable solution.

grubolsch: I already asked Frederic this question :-) We know the answer on wednesday.

PR #1661, opened 2024-03-19
III-6126 - Add force option to place:process-duplicates

grubolsch:
### Added
Add force option to place:process-duplicates

---
Ticket: https://jira.uitdatabank.be/browse/III-6125

### DIFF
--- a/app/Console/Command/ProcessDuplicatePlaces.php
+++ b/app/Console/Command/ProcessDuplicatePlaces.php
@@ -25,6 +25,7 @@ use Symfony\Component\Console\Question\ConfirmationQuestion;
 final class ProcessDuplicatePlaces extends AbstractCommand
 {
     private const ONLY_RUN_CLUSTER_ID = 'only-run-cluster-id';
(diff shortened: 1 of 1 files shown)


app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -173,6 +175,11 @@ public function execute(InputInterface $input, OutputInterface $output): int
 
     private function askConfirmation(InputInterface $input, OutputInterface $output, int $count): bool
     {
+        if ($input->getOption(self::FORCE) === true) {
+            $output->writeln("This action will process {$count} clusters");
```
JonasVHG, line 179: I do not think this feedback is very useful in this case, `force` options should only be run via automated `cronjob` or via `Jenkins`. Besides, it cannot be stopped. We also do not give this type of feedback on other `force` commands.

JonasVHG (approved):
Just a minor preference remark, but you can leave it in, if you find it useful.

PR #1662, opened 2024-03-19
III-6120/clean place names before sending api call

grubolsch:
### Added
Added cleanup of location name towards geocoordinates API call - using slide 3.

**Before I merge this - I want to check tomorrow with Frederic if everything is implemented as he intended!**

---
Ticket: https://jira.uitdatabank.be/browse/III-6120

### DIFF
--- a/src/Place/ReadModel/Duplicate/CleanPlaceName.php
+++ b/src/Place/ReadModel/Duplicate/CleanPlaceName.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Place\ReadModel\Duplicate;
 
+use CultuurNet\UDB3\Json;
 use CultuurNet\UDB3\Model\ValueObject\Geography\Address;
 
 class CleanPlaceName
--- a/src/Place/ReadModel/Duplicate/CleanPlaceName.php
+++ b/src/Place/ReadModel/Duplicate/CleanPlaceName.php
@@ -17,235 +17,28 @@ use CultuurNet\UDB3\Model\ValueObject\Geography\Address;
  * */
 class CleanPlaceName
 {
+    private const MAX_LENGTH_TITLE = 150;
+    private const NONE = 'none';
+
     public static function transform(Address $address, string $title): string
     {
-        $title = str_replace(['.', ''], '', mb_strtolower(trim($title)));
-
-        $title = self::removeAccents($title);
-
-        $title = str_replace(['"', '\'', '?', '!', '(', ')', '&', ',', '_', ':'], ' ', $title);
-
-        $title = self::removeCityName($address, $title);
-
-        $title = self::removeDuplicateWords($title);
-
-        return self::reduceMultipleSpaces($title);
-    }
-
-    private static function removeAccents(string $title): string
-    {
-        // believe it or not - this function was "loaned" from the Wordpress engine (GPLv2)
-
-        if (!preg_match('/[\x80-\xff]/', $title)) {
-            return $title;
+        if (mb_strtolower($title) === self::NONE) {
+            return '';
         }
 
-        return strtr($title, [
-            // Decompositions for Latin-1 Supplement
-            chr(195) . chr(128) => 'A',
-            chr(195) . chr(129) => 'A',
-            chr(195) . chr(130) => 'A',
-            chr(195) . chr(131) => 'A',
-            chr(195) . chr(132) => 'A',
-            chr(195) . chr(133) => 'A',
-            chr(195) . chr(135) => 'C',
-            chr(195) . chr(136) => 'E',
-            chr(195) . chr(137) => 'E',
-            chr(195) . chr(138) => 'E',
-            chr(195) . chr(139) => 'E',
-            chr(195) . chr(140) => 'I',
-            chr(195) . chr(141) => 'I',
-            chr(195) . chr(142) => 'I',
-            chr(195) . chr(143) => 'I',
-            chr(195) . chr(145) => 'N',
-            chr(195) . chr(146) => 'O',
-            chr(195) . chr(147) => 'O',
-            chr(195) . chr(148) => 'O',
-            chr(195) . chr(149) => 'O',
-            chr(195) . chr(150) => 'O',
-            chr(195) . chr(153) => 'U',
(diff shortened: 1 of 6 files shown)


src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,36 @@
...
+
+        if (mb_strlen($title) > self::MAX_LENGTH_TITLE) {
+            return '';
+        }
+
+        if (str_contains($title, $address->getStreet()->toString())) {
+            return '';
+        }
+
+        $title = htmlspecialchars($title);
```
LucWollants, line 29: On the slides is the opposite, the conversion of `&amp` to `&`

src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if (mb_strtolower($title) === self::NONE) {
+            return '';
+        }
+
+        if (mb_strlen($title) > self::MAX_LENGTH_TITLE) {
+            return '';
+        }
+
+        if (str_contains($title, $address->getStreet()->toString())) {
```
LucWollants, line 25: I think when it's a full match only and not when it contains it?

src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,36 @@
...
+            return '';
+        }
+
+        $title = htmlspecialchars($title);
+
+        $title = str_replace([' BE', 'BE ', ' NL', 'NL '], [' Belgium', 'Belgium ', ' Netherlands', 'Netherlands '], $title);
```
LucWollants, line 31: I think this replacement is done on the address part and not on the location name.

LucWollants (changes requested):
I think there are still some essential differences between this implementation and the implementation of the data team. I guess a longer list of examples is the best way to discuss this.

src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,36 @@
...
+
+        //Decode the unicode characters
+        return Json::decodeAssociatively('"' . $title . '"');
```
LucWollants, line 34: There is also the method `Json::decode`.

But it feels weird to use JSON conversion to do Unicode conversion. I guess these are the quirks of PHP. I would expect something like `mb_convert_encoding($unicodeString, 'ASCII//TRANSLIT', 'UTF-8')`

src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,36 @@
...
+use CultuurNet\UDB3\Model\ValueObject\Geography\Address;
+
+class CleanPlaceName
+{
+    private const MAX_LENGTH_TITLE = 150;
+    private const NONE = 'none';
+
+    public static function transform(Address $address, string $title): string
+    {
+        if (mb_strtolower($title) === self::NONE) {
```
LucWollants, line 17: I guess this should match various casing?

grubolsch: it does now? Or do you mean it should not?

src/Place/ReadModel/Duplicate/CleanPlaceName.php
grubolsch: No, if you check the python version I am correct

src/Place/ReadModel/Duplicate/CleanPlaceName.php
grubolsch: I tried, but it gave problems with characters like é, it converted them to ?

LucWollants: Thanks for trying!

src/Place/ReadModel/Duplicate/CleanPlaceName.php
grubolsch: it is not, it is done on location name to prevent double information in the google api call, given that we just concat location name and address.
I did make the code work for more countries.

grubolsch:
In the end we decided to remove a lot of the edge cases. I also did manual research, I had no problems fetching info for the Kouter tennisclub, and I did not find a single place with the location name "none".

src/Place/ReadModel/Duplicate/CleanPlaceName.php
```diff
@@ -0,0 +1,78 @@
...
+
+        $title = self::replaceCountryCodes($title);
+
+        // Decode the unicode characters
+        return Json::decode('"' . $title . '"');
+    }
+
+    private static function replaceCountryCodes(string $title): string
+    {
+        $countries = [
```
LucWollants, line 35: If we maintain a big list like this I would move it to a config file. Otherwise I would only support BE, NL, FR and DE

grubolsch: Both are fine for me. In practice I think BE and NL will cover 99% of all places.
should I just shrink the array?

tests/Place/ReadModel/Duplicate/CleanPlaceNameTest.php
```diff
@@ -0,0 +1,46 @@
...
+            ['Café de lindekens', 'Café de lindekens'],
+            ['<b>Café de lindekens</b>', 'Café de lindekens'],
+            ['\u0062\u0061\u0072 la cantina', 'bar la cantina'],
+            ['aan de hoek van het kruispunt om 15 uur vertrekken we aan de hoek van het kruispunt om 15 uur vertrekken we aan de hoek van het kruispunt om 15 uur vertrekken we', ''],
+            ['Speelplaats aan de Kerkstraat 1', ''],
+            ['Café de lindekens BE', 'Café de lindekens Belgium'],
+            ['BE Café de lindekens', 'Belgium Café de lindekens'],
+            ['dit is belachelijk om te veranderen', 'dit is belachelijk om te veranderen'],
+            ['FR', ''],
+            ['France', ''],
```
LucWollants, line 43: I'm not sure if this is needed.

LucWollants (approved):

PR #1664, opened 2024-03-27
III-6030 Rename duplicate place creation flag

grubolsch:
### Changed
Renamed flag "prevent_duplicate_creation" to "prevent_duplicate_places_creation"

### Related

---
Ticket: https://jira.uitdatabank.be/browse/III-6030


LucWollants (changes requested):
This will break the existing code. I think you will need to do this in 2 steps. There are 2 ways to work around this:
- Check in the code and support both `prevent_duplicate_creation` and `prevent_duplicate_places_creation`, release the code, than release the config, than fix the code
- Change the config and keep `prevent_duplicate_creation` and `prevent_duplicate_places_creation`, release the config, release the code, than fix the config

LucWollants (approved):
The config was updated to support the old and new name of the flag

PR #1665, opened 2024-03-28
Log changes in coordinates or title for Place

grubolsch:
### Added
Logging calls to changes in coordinates or title for Place.

Currently does not do a lot but is a placeholder for fetching the geo coordinates for the new index for the data team.

---
Ticket: https://jira.uitdatabank.be/browse/III-6133


LucWollants (approved):
Once the real implementation is done, a better name can be chosen.

PR #1668, opened 2024-04-02
III-6072 Add endpoint to reject ownership

LucWollants:
### Added
- Added endpoint `POST /ownerships/{ownershipId}/reject` to reject an ownership

### Note
- There is some overlapping code in the request handler, I will change that in a follow-up pull request

---
Ticket: https://jira.uitdatabank.be/browse/III-6072

### DIFF
--- a/features/Steps/OwnershipSteps.php
+++ b/features/Steps/OwnershipSteps.php
@@ -38,7 +38,7 @@ trait OwnershipSteps
         );
         $this->responseState->setResponse($response);
 
-        $this->theResponseStatusShouldBe(202);
+        $this->theResponseStatusShouldBe(204);
         $this->theResponseBodyShouldBeValidJson();
     }
 
@@ -53,7 +53,7 @@ trait OwnershipSteps
         );
         $this->responseState->setResponse($response);
 
-        $this->theResponseStatusShouldBe(202);
+        $this->theResponseStatusShouldBe(204);
         $this->theResponseBodyShouldBeValidJson();
     }
 
--- a/features/Steps/OwnershipSteps.php
+++ b/features/Steps/OwnershipSteps.php
@@ -43,6 +43,21 @@ trait OwnershipSteps
     }
 
     /**
+     * @When I reject the ownership with ownershipId :ownershipId
+     */
+    public function iRejectTheOwnershipWithOwnershipId(string $ownershipId): void
+    {
+        $response = $this->getHttpClient()->postJSON(
+            '/ownerships/' . $this->variableState->replaceVariables($ownershipId) . '/reject',
+            ''
+        );
+        $this->responseState->setResponse($response);
+
+        $this->theResponseStatusShouldBe(202);
+        $this->theResponseBodyShouldBeValidJson();
+    }
+
+    /**
      * @When I get the ownership with ownershipId :ownershipId
      */
     public function iGetTheOwnershipWithOwnershipId(string $ownershipId): void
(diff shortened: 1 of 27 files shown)


features/Steps/OwnershipSteps.php
```diff
@@ -42,6 +42,21 @@ public function iApproveTheOwnershipWithOwnershipId(string $ownershipId): void
...
     }
 
+    /**
+     * @When I reject the ownership with ownershipId :ownershipId
+     */
+    public function iRejectTheOwnershipWithOwnershipId(string $ownershipId): void
+    {
+        $response = $this->getHttpClient()->postJSON(
+            '/ownerships/' . $this->variableState->replaceVariables($ownershipId) . '/reject',
+            ''
```
grubolsch, line 52: Why do we have to send an empty body? Seems not elegant?

src/Http/Ownership/RejectOwnershipRequestHandler.php
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            $ownership = $this->ownershipSearchRepository->getById($ownershipId);
+        } catch (OwnershipItemNotFound $exception) {
+            throw ApiProblem::ownershipNotFound($ownershipId);
+        }
+
+        // Make sure the current user can reject the ownership
+        // This means that the current user is a god user
+        // Or that the current user is the owner of the item
+        if (!$this->isAllowedToRejectOwnership($ownership)) {
```
grubolsch, line 59: Same remark as other pr - maybe move this comment block to the correct method

src/Http/Ownership/RejectOwnershipRequestHandler.php
```diff
@@ -0,0 +1,85 @@
...
+            throw ApiProblem::forbidden('You are not allowed to reject this ownership');
+        }
+
+        $this->commandBus->dispatch(
+            new RejectOwnership(new UUID($ownershipId), new UserId($this->currentUser->getId()))
+        );
+
+        return new JsonResponse(
+            [],
```
grubolsch, line 68: This is just going to respond an empty body?
Should we respond with a 204 instead then?

I think it would be better to respond with something relevant, like the ownershipId.
In general we often respond with the id after an action so that would keep it consistent.

grubolsch (changes requested):

src/Http/Ownership/RejectOwnershipRequestHandler.php
LucWollants: A 204 could indeed be better. But I don't think we need to pass a response body because the id would already be known.

features/Steps/OwnershipSteps.php
LucWollants: Do you mean inside requesting code here?

src/Ownership/Commands/ApproveOwnership.php
```diff
@@ -10,21 +10,21 @@
 final class ApproveOwnership
 {
     private UUID $id;
-    private UserId $approverId;
+    private UserId $requesterId;
 
     public function __construct(UUID $id, UserId $requesterId)
```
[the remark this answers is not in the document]
LucWollants: I added it because an ownership request can be done for someone else. As an admin, I can request the ownership of an item for someone else. But for approving or rejecting it is only the action itself that needs to be guarded with the current user and that is validated inside the HTTP request handler.

grubolsch (approved):

PR #1669, opened 2024-04-02
Add tag to sapi3 feature tests

grubolsch:
### Added
I added a tag to all feature tests related to SAPI3, to make it easier to change code in the sapi3 code base.
I also added a new Make command to run commands with a specific tag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,11 @@ test-group:
 feature-init:
 	docker exec -it php.uitdatabank composer feature -- --tags @init
 
+feature-tag:
+	docker exec -it php.uitdatabank composer feature -- --tags $(tag)
+
 feature:
-	docker exec -it php.uitdatabank composer feature  -- --tags "~@init"
+	docker exec -it php.uitdatabank composer feature -- --tags "~@init"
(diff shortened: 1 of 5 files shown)


LucWollants (commented):
In the past we made use of tags, but they were hard to maintain and let to unexpected issues.

I also think there are tests that indirectly use SAPI3, for example exports.

LucWollants:
The more I think about it, `sapi3` can not be a tag because permission voters are handled with SAPI3 requests, so almost all requests require SAPI3 in one or another way.

grubolsch:
> The more I think about it, `sapi3` can not be a tag because permission voters are handled with SAPI3 requests, so almost all requests require SAPI3 in one or another way.

I understand. But is there a middleroad where we could run a specific set of feature tests when working on SAPI3?
Once we do 1 permission check with sapi3 we don't need to run the entire suit of tests.

Just looking if we can get a sensible compromise? 
I found the tag really helpful during my work on the PHP8 SAPI3 upgrade.

LucWollants (approved):

PR #1679, opened 2024-04-05
III-6075 Avoid permission on places

LucWollants:
### Fixed
- Avoid permission on places associated with an organizer with approved ownership by adding the `_type:event` advanced query

---
Ticket: https://jira.uitdatabank.be/browse/III-6075

### DIFF
--- a/features/ownership/permission.feature
+++ b/features/ownership/permission.feature
@@ -92,7 +92,7 @@ Feature: Test permissions based on ownership
     And I send a PUT request to "/events/%{eventId}/name/nl"
     Then the response status should be "403"
 
-  Scenario: Approving the ownership of an organizer gives permission no permission on the place associated with the organizer
+  Scenario: Approving the ownership of an organizer doesn't give permission on the place associated with the organizer
     Given I create a minimal organizer and save the "id" as "organizerId"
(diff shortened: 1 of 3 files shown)


features/ownership/permission.feature
```diff
@@ -90,4 +90,26 @@ Feature: Test permissions based on ownership
         {"name": "madewithlove"}
     """
     And I send a PUT request to "/events/%{eventId}/name/nl"
+    Then the response status should be "403"
+
+  Scenario: Approving the ownership of an organizer gives permission no permission on the place associated with the organizer
```
JonasVHG, line 95:
```suggestion
  Scenario: Approving the ownership of an organizer doesn't give permission on the place associated with the organizer
```
`gives permission no permission ` is a bit difficult to understand if you are not familiar with the background.

JonasVHG (approved):
Code looks fine, just 1 remark about the scenario description, that might be hard to understand without background.

features/ownership/permission.feature
LucWollants: That was indeed a copy/paste error

simon-debruijn (approved):

PR #1681, opened 2024-04-08
III-6079 movie fetcher repository

JonasVHG:
### Added

- `MovieRepository`: Create & check relation between Kinepolis Id's & publiq Id's

---

### DIFF
--- a/app/Migrations/Version20240402125210.php
+++ b/app/Migrations/Version20240402125210.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20240402125210 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable('movie_mapping');
+
+        $table->addColumn('event_id', Types::GUID)->setLength(36)->setNotnull(true);
+        $table->addColumn('movie_id', Types::STRING)->setNotnull(true);
+
+        $table->setPrimaryKey(['event_id']);
+        $table->addIndex(['movie_id']);
+        $table->addUniqueIndex(['movie_id', 'event_id']);
+    }
+
+    public function down(Schema $schema): void
+    {
+        $schema->dropTable('movie_mapping');
+    }
+}
--- a/app/Migrations/Version20240402125210.php
+++ b/app/Migrations/Version20240402125210.php
@@ -12,7 +12,7 @@ final class Version20240402125210 extends AbstractMigration
 {
     public function up(Schema $schema): void
     {
-        $table = $schema->createTable('movie_mapping');
+        $table = $schema->createTable('kinepolis_movie_mapping');
 
         $table->addColumn('event_id', Types::GUID)->setLength(36)->setNotnull(true);
         $table->addColumn('movie_id', Types::STRING)->setNotnull(true);
@@ -24,6 +24,6 @@ final class Version20240402125210 extends AbstractMigration
 
     public function down(Schema $schema): void
     {
-        $schema->dropTable('movie_mapping');
+        $schema->dropTable('kinepolis_movie_mapping');
     }
 }
--- a/src/Movie/MovieRepository.php
+++ b/src/Movie/MovieRepository.php
@@ -10,7 +10,7 @@ final class MovieRepository
 {
     private Connection $connection;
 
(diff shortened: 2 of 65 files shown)


app/Migrations/Version20240402125210.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable('movie_mapping');
+
+        $table->addColumn('event_id', Types::GUID)->setLength(36)->setNotnull(true);
+        $table->addColumn('movie_id', Types::STRING)->setNotnull(true);
+
+        $table->setPrimaryKey(['event_id']);
+        $table->addIndex(['movie_id']);
```
simon-debruijn, line 21: I think this index is unneeded. SInce the composite unique index covers searching on movie_id

src/Movie/MovieRepository.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $this->connection = $connection;
+    }
+
+    public function getEventIdByMovieId(string $movieId): ?string
+    {
+        $result = $this->connection->createQueryBuilder()
+            ->select('event_id')
+            ->from(self::TABLE)
+            ->where('movie_id' . ' = :movie_id')
```
simon-debruijn, line 25:
```suggestion
            ->where('movie_id = :movie_id')
```

tests/Movie/MovieRepositoryTest.php
```diff
@@ -0,0 +1,95 @@
...
+use CultuurNet\UDB3\DBALTestConnectionTrait;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Types;
+use PHPUnit\Framework\TestCase;
+
+final class MovieRepositoryTest extends TestCase
+{
+    use DBALTestConnectionTrait;
+    public const TABLE_NAME = 'movie_mapping';
```
simon-debruijn, line 16:
```suggestion
    
    public const TABLE_NAME = 'movie_mapping';
```

tests/Movie/MovieRepositoryTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $this->movieRepository = new MovieRepository(
+            $this->getConnection()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_get_the_event_id_for_a_given_movie(): void
```
simon-debruijn, line 62:
```suggestion
    public function it_gets_the_event_id_for_a_given_movie(): void
```

tests/Movie/MovieRepositoryTest.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->assertEquals(
+            $this->eventId,
+            $this->movieRepository->getEventIdByMovieId($this->movieId)
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_returns_null_for_an_new_movie_id(): void
```
simon-debruijn, line 73:
```suggestion
    public function it_returns_null_for_a_new_movie_id(): void
```

simon-debruijn (approved):

app/Migrations/Version20240402125210.php
LucWollants: I'm not sure about this. I think the combination index is separate from the single-column index.

src/Movie/MovieRepository.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Movie;
+
+use Doctrine\DBAL\Connection;
+
+final class MovieRepository
```
LucWollants, line 9: This name is too general, I would make a domain folder called `Kinepolis`

app/Migrations/Version20240402125210.php
simon-debruijn: https://planetscale.com/learn/courses/mysql-for-developers/indexes/duplicate-indexes?autoplay=1

LucWollants: In short: if the column is the first column inside the unique index no separate index is needed.

See: https://dev.mysql.com/doc/refman/8.0/en/mysql-indexes.html

>If the table has a multiple-column index, any leftmost prefix of the index can be used by the optimizer to look up rows. For example, if you have a three-column index on (col1, col2, col3), you have indexed search capabilities on (col1), (col1, col2), and (col1, col2, col3). For more information, see [Section 10.3.6, “Multiple-Column Indexes”](https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html).

LucWollants (approved):

PR #1682, opened 2024-04-08
III-6079 moviefetcher-cli-base

JonasVHG:
### Added

- `FetchMoviesFromKinepolisApi`: CLI-Command to read & create `Events` from `movies`
- `ParsedMovie`: Class to facilitate converting the KinepolisModel of a Movie to the UiTdatabankModel of an Event
- `AutenticatedKinepolisClient` & interface `KinepolisClient`: Client to call the various endpoints of the `Kinepois`-api
- `KinepolisDateParser` & interface `DateParser`: Converts the dates from the external Kinepolis API to a format that can be used in `UiTdatabank`
- `KinepolisParser` & interface `Parser`: basic Parser for Kinepolis Json result.
- `KinepolisService`: wrapper for `KinepolisService`, `KinepolisClient`, `Parser` & `MovieRepository`

### Changed

- `ConsoleServiceProvider`: Registered `FetchMoviesFromKinepolisApi`-command

### Fixed

- Movies can be created & updated after fetching the external API.

### TODO

- Price & Image

---


src/Kinepolis/AutenticatedKinepolisClient.php
```diff
@@ -0,0 +1,86 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Kinepolis;
+
+use CultuurNet\UDB3\Json;
+use GuzzleHttp\Psr7\Request;
+use Psr\Http\Client\ClientInterface;
+
+final class AutenticatedKinepolisClient implements KinepolisClient
```
LucWollants, line 11:
```suggestion
final class AuthenticatedKinepolisClient implements KinepolisClient
```

src/Kinepolis/KinepolisClient.php
```diff
@@ -0,0 +1,16 @@
...
+
+interface KinepolisClient
+{
+    public function getToken(): string;
+
+    public function getMovies(string $token): array;
+
+    public function getMovieDetail(string $token, int $mid): array;
+
+    public function getHeaders(string $token = null): array;
```
LucWollants, line 15: I think that this method needs to be private and can be called `createHeaders`

app/Console/Command/FetchMovies.php
```diff
@@ -0,0 +1,110 @@
...
+use CultuurNet\UDB3\Kinepolis\KinepolisService;
+use CultuurNet\UDB3\Kinepolis\MovieMappingRepository;
+use CultuurNet\UDB3\Language as LegacyLanguage;
+use CultuurNet\UDB3\Kinepolis\ParsedMovie;
+use CultuurNet\UDB3\Offer\Commands\UpdateCalendar;
+use CultuurNet\UDB3\Event\Commands\UpdateDescription;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class FetchMovies extends AbstractCommand
```
LucWollants, line 22: I would make the name again more specific, this can certainly refer to Kinepolis

src/Kinepolis/KinepolisService.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Kinepolis;
+
+final class KinepolisService
```
LucWollants, line 7: The reason for a service is to combine all the work that is done inside the CLI command and inject this with interfaces. So getting movies, parsing those movies, and dispatching commands should all happen here.

Because the client and parser are interfaces they can be mocked and the commands passed on the recording command bus can be validated.

src/Kinepolis/DateParser.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Kinepolis;
+
+interface DateParser
```
LucWollants, line 7: There is no real benefit to this being an interface, it's better to make the full parser an interface.

tests/Kinepolis/KinepolisDateParserTest.php
```diff
@@ -0,0 +1,972 @@
...
+     */
+    public function it_converts_dates_from_kinepolis(array $data, int $length, array $result): void
+    {
+        $this->assertEquals(
+            $this->dateParser->processDates($data, $length),
+            $result
+        );
+    }
+
+    public function dateDataProvider(): array
```
LucWollants, line 36: Very nice to see these detailed tests, but maybe the movie arrays can be placed inside simple PHP files that return an array?

tests/Kinepolis/samples/example.json
```diff
@@ -0,0 +1,509 @@
+{
```
LucWollants, line 1: Maybe give this a more explicit name?

LucWollants (changes requested):
Overall a very good approach on fetching an parsing movies from Kinepolis. I added a few remarks to make the code more clear with better naming and some design changes to making testing easier.

I would also add some logging to a dedicated logger file.

src/Kinepolis/DateParser.php
JonasVHG: After discussing this, we decided all the Parsers need interfaces, both the main & the subParsers.

app/Console/Command/FetchMoviesFromKinepolisApi.php
```diff
@@ -0,0 +1,57 @@
...
+    private KinepolisService $service;
+
+    public function __construct(KinepolisService $service)
+    {
+        parent::__construct();
+        $this->service = $service;
+    }
+    public function configure(): void
+    {
+        $this->setName('movies:fetch');
```
LucWollants, line 25: Maybe `kinepolis:fetch-movies`?

app/Console/Command/FetchMoviesFromKinepolisApi.php
```diff
@@ -0,0 +1,57 @@
...
+            null,
+            InputOption::VALUE_NONE,
+            'Skip confirmation.'
+        );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        if ($this->askConfirmation($input, $output)) {
+            $this->service->fetch();
```
LucWollants, line 38: It does more than only a fetch, I think import is a better name?

src/Kinepolis/KinepolisService.php
```diff
@@ -0,0 +1,130 @@
...
+        $this->commandBus = $commandBus;
+        $this->aggregateRepository = $aggregateRepository;
+        $this->client = $client;
+        $this->parser = $parser;
+        $this->movieMappingRepository = $movieMappingRepository;
+        $this->uuidGenerator = $uuidGenerator;
+        $this->logger = $logger;
+    }
+
+    public function getClient(): KinepolisClient
```
LucWollants, line 52: There is no need for this method and also no use for `getParser`

src/Kinepolis/KinepolisService.php
```diff
@@ -0,0 +1,130 @@
...
+                $this->dispatch($parsedMovie);
+            }
+        }
+    }
+
+    private function dispatch(ParsedMovie $parsedMovie): void
+    {
+        $commands = [];
+        $eventId = $this->movieMappingRepository->getByMovieId($parsedMovie->getExternalId());
+        $eventExists = $eventId !== null;
```
LucWollants, line 91: This is only used once, so need for a local variable.

src/Kinepolis/ParsedMovie.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Kinepolis;
+
+use CultuurNet\UDB3\Description;
```
LucWollants, line 7: I would try to avoid using the legacy classes and only do the conversion from the new classes to the old classes where it is really needed. So in this case I would use the new description here and only do the conversion in the service. If we do the refactor of the `UpdateDescription` only one classes needs to be changed.

src/Kinepolis/KinepolisDateParser.php
```diff
@@ -0,0 +1,87 @@
...
+
+    private function getFromTime(string $day, string $time): \DateTimeImmutable
+    {
+        $timeZoneBrussels = new \DateTimeZone('Europe/Brussels');
+        $timeZoneUtc = new \DateTimeZone('UTC');
+        $dt = \DateTimeImmutable::createFromFormat('Y-m-d H:i:s', $day . ' ' . $time, $timeZoneBrussels);
+        return $dt->setTimezone($timeZoneUtc);
+    }
+
+    private function getFormat(array $formats): string
```
LucWollants, line 67: I first thought this referred to the date time format, but this refers to being either 2D or 3D

src/Kinepolis/KinepolisDateParser.php
```diff
@@ -0,0 +1,87 @@
...
+        return $this->timeTableList;
+    }
+
+    private function processDay(string $day, array $timeList, int $length): void
+    {
+        foreach ($timeList as $info) {
+            $format = $this->getFormat($info['format']);
+            $from = $this->getFromTime($day, $info['time']);
+            $to = $this->getToTime($from, $length);
+            $this->timeTableList[$info['tid']][$format][] = new SubEvent(
```
LucWollants, line 46: Do we need to stick with this format or could we also make it a typed object like `ParsedMovie`?

JonasVHG: I might be to hard to include it in the first release, but using using a typed Object like `ParsedMovie` could be an optimalization in the future.

src/Kinepolis/KinepolisDateParser.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->timeTableList = [];
+
+        foreach ($dates as $day => $timeList) {
+            $this->processDay($day, $timeList, $length);
+        }
+
+        return $this->timeTableList;
+    }
+
+    private function processDay(string $day, array $timeList, int $length): void
```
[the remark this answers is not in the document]
JonasVHG: As discussed via Slack. this method doesn't really create subEvents but takes the Dimension & Location into consideration.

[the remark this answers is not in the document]
LucWollants: As discussed during pair programming, nice to have but not for now. Also see https://github.com/cultuurnet/udb3-backend/pull/1682#discussion_r1565479489

LucWollants (approved):
Thanks for all the renames, refactors and optimisations!

PR #1683, opened 2024-04-09
III-6079 movie-fetcher price

JonasVHG:
### Added

- `ParsedPriceForATheater`: To store the prices of a specific theater
- `KinepolisPriceParser` & interface `PriceParser`: The parse the prices in the format of the external API.

### Changed

- `KineepolisClient`: added `getTheaters()` & `getPricesForATheater()` to get the price information from the external API.
- `ParsedMovie`: Added PriceInfo
- `KinepolisService`: injected `PriceParser`
- `KinepolisService`: also dispatch command `UpdatePriceInfo`

### Fixed

- Fetched movies will now have a Price

---


src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +72,34 @@ public function getMovieDetail(string $token, int $mid): array
...
+        $request = new Request(
+            'GET',
+            $this->movieApiBaseUrl . 'services/content/1.1/theaters',
+            $this->createHeaders($token)
+        );
+
+        $response = $this->client->sendRequest($request)->getBody()->getContents();
+        $contents = Json::decodeAssociatively($response);
+
+        return $contents['theatres'];
```
grubolsch, line 86: Should we do a check here if theatres exist?

```suggestion
        return $contents['theatres'] ?? [];
```

In general I am missing a check if the API request failed? How would that error be capture if for example the API is down?

src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +72,34 @@ public function getMovieDetail(string $token, int $mid): array
         return $contents['movies'][0];
     }
 
+    public function getTheaters(string $token): array
```
grubolsch, line 75: I prefer to call methods that actually fetch information from API's something like "fetchTheaters" instead of the get-prefix.
The get-prefix often communicates that all the information is already in memory, that is not the case here.

src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +72,34 @@ public function getMovieDetail(string $token, int $mid): array
...
+        $request = new Request(
+            'GET',
+            $this->movieApiBaseUrl . 'services/content/1.1/theaters/' . $tid,
+            $this->createHeaders($token)
+        );
+
+        $response = $this->client->sendRequest($request)->getBody()->getContents();
+        $contents = Json::decodeAssociatively($response);
+
+        return $contents['theatres'][0]['tariffs'][0]['tarifs'];
```
grubolsch, line 100: Add isset() check if this long chain actually exists?

src/Kinepolis/AuthenticatedKinepolisClient.php
grubolsch: Would it be better instead of returning an array directly from the API to create a "Theater" DTO that wraps all the info? That way when the API ever changes it's format we don't have to change all the underlying code.
Also will make for better typehints.

src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +72,34 @@ public function getMovieDetail(string $token, int $mid): array
...
+            $this->createHeaders($token)
+        );
+
+        $response = $this->client->sendRequest($request)->getBody()->getContents();
+        $contents = Json::decodeAssociatively($response);
+
+        return $contents['theatres'];
+    }
+
+    public function getPricesForATheater(string $token, string $tid): array
```
grubolsch, line 89: Can we get a docblock what the structure is of this array, or convert it to a dto?

src/Kinepolis/KinepolisParser.php
```diff
@@ -54,6 +58,10 @@ public function getParsedMovies(array $moviesToParse): array
                 $is3D = $dimension === '3D';
                 $title = $is3D ? $title . ' 3D' : $title;
 
+                $isLong = $length >= 135;
+
+                $parsedPrice = $parsedPrices[$theatreId];
+
                 $calendar = sizeof($subEvents) === 1 ?
```
grubolsch, line 65: count() is the more used method (3 times as much in the UDB3 code base), I would use that as a standard.

src/Kinepolis/KinepolisPriceParser.php
```diff
@@ -0,0 +1,53 @@
...
+        $studentPrice = 0;
+        $surchargeForLongMovie = 0;
+        $surchargeFor3D = 0;
+
+        // This is done because the array of prices of Kinepolis
+        // is always random and includes prices that are irrelevant for us
+        // like prices for various models of 3d glasses
+        // We assume the values below are always present
+        foreach ($theaterPrices as $theaterPrice) {
+            if ($theaterPrice[1] === 'Normaal tarief') {
```
grubolsch, line 22: This entire structure could be a switch, maybe also move the strings to constants?

src/Kinepolis/KinepolisPriceParser.php
```diff
@@ -0,0 +1,53 @@
...
+            $studentPrice,
+            $surchargeForLongMovie,
+            $surchargeFor3D
+        );
+    }
+
+    // All their prices are in the format of € 0,00
+    private function normalizePrice(string $price): int
+    {
+        return (int) (preg_replace('/[^0-9]/', '', $price));
```
grubolsch, line 51: Prices never have a decimal like 10.50 ?

src/Kinepolis/KinepolisService.php
```diff
@@ -58,6 +63,34 @@ public function import(): void
             $this->logger->error('Problem with Kinepolis Service Authentication: ' . $exception->getMessage());
             return;
         }
+
+        try {
+            $theaters = $this->client->getTheaters($token);
+        } catch (\Exception $exception) {
```
grubolsch, line 69: Is there a more specific exception we could listen to instead of the generic Exception?
I would also just import the namespace.

src/Kinepolis/Parser.php
```diff
@@ -7,7 +7,8 @@
 interface Parser
```
grubolsch, line 7: Should this be a more specific name like movieParser?
I know it is in the Kinepolis namespace but this feels to generic.
If you don't feel that way - feel free to ignore this comment.

src/Kinepolis/KinepolisParser.php
```diff
@@ -54,6 +58,10 @@ public function getParsedMovies(array $moviesToParse): array
                 $is3D = $dimension === '3D';
                 $title = $is3D ? $title . ' 3D' : $title;
 
+                $isLong = $length >= 135;
```
grubolsch, line 61: Magic value, move to a constant

src/Kinepolis/KinepolisService.php
```diff
@@ -58,6 +63,34 @@ public function import(): void
...
+
+        /**
+         * @var ParsedPriceForATheater[] $parsedPrices
+         */
+        $parsedPrices = [];
+        foreach ($theaters as $theater) {
+            try {
+                $pricesToParse = $this->client->getPricesForATheater($token, $theater['tid']);
+            } catch (\Exception $exception) {
+                $this->logger->error('Problem with fetching the prices from Theater ' . $theater['tid'] . ': ' . $exception->getMessage());
```
grubolsch, line 82: Really not a fan of the  $theater['tid'] structure, would move this to a DTO

grubolsch (changes requested):

src/Kinepolis/KinepolisPriceParser.php
JonasVHG: There are strings with a `€`-sign and a comma never a dot(see my comment)
Since our `Money` object expects cents and not Euro's return the value in cents.

grubolsch: Maybe mention in the comment the return value is in cents?

src/Kinepolis/AuthenticatedKinepolisClient.php
LucWollants: The parsers are used for this conversion. The API client and parsers are used in the service.

src/Kinepolis/Parser.php
LucWollants: I also think that a sub folder is needed with all the Parsers

src/Kinepolis/KinepolisService.php
LucWollants: No strong opinion on this one. Because it will always be a conversion from an array to something more typed and that is now done inside the parsers. This way we have a nice separation for concerns. The API client just gets the arrays, the parsers convert them and the service binds it all together.

src/Kinepolis/AuthenticatedKinepolisClient.php
JonasVHG: If the entire KinepolisAPI is down, we could not fetch a `token` and thus not continue.
In most cases, including this one, I am in favour of failing hard & fast.
If we cannot fetch the theaters, we cannot fetch the Prices, so we could not add `PriceInfo`.

grubolsch: There might be a situation where the API call succeeds, but returns something unexpected.
I am fine with throwing an exception in such case, but do think we should test for it.
Maybe add that catched exception to a log?

src/Kinepolis/KinepolisService.php
LucWollants: There is a `ParsedMovie` class.

src/Kinepolis/KinepolisParser.php
LucWollants: Let us also create a follow-up ticket to remove all occurrences of `sizeof` and add a code style rule to always use `count` otherwise this change is meaningless.

JonasVHG: https://jira.publiq.be/browse/III-6150

src/Kinepolis/KinepolisService.php
JonasVHG: And a `ParsedPriceForATheater` which is the only thing we need from the KinepolisAPI.
All necessary information for a Theater are stored inside the Places, which are already in UiTdatabank.
Only the Price is Needed because there format with surcharges depending on the length or dimension is incompatible with our PriceInfo format.

src/Kinepolis/AuthenticatedKinepolisClient.php
JonasVHG: I agree, any exception thrown by the `Client` should be caught in the `KinepolisService` & written to the log+sentry.

src/Kinepolis/AuthenticatedKinepolisClient.php
JonasVHG: I think the task of the Client should just be to fetch the Json. The Parsers should Convert the Json to Typed Objects.

grubolsch (approved):

PR #1686, opened 2024-04-15
III-6068 newsletter api

grubolsch:
### Added
I added a new endpoint to subscribe to a newsletter, removing the need for the current mail-microservice.
Contains a unit test and a feature test to prove it works.

I also tested it on the acc env of mailjet and saw the new contacts in the export.

It will also need a change in the frontend variable ["newsletterApiUrl"](https://github.com/cultuurnet/udb3-frontend/blob/main/src/hooks/api/newsletter.ts#L7C3-L7C101) (Thanks to @simon-debruijn)

Related to the config changes in https://github.com/cultuurnet/appconfig/pull/588

---
Ticket: https://jira.uitdatabank.be/browse/III-6068

### DIFF
--- a/app/Http/PsrRouterServiceProvider.php
+++ b/app/Http/PsrRouterServiceProvider.php
@@ -142,6 +142,7 @@ use CultuurNet\UDB3\Http\Productions\SearchProductionsRequestHandler;
 use CultuurNet\UDB3\Http\Productions\SkipEventsRequestHandler;
 use CultuurNet\UDB3\Http\Productions\SuggestProductionRequestHandler;
 use CultuurNet\UDB3\Http\Proxy\ProxyRequestHandler;
+use CultuurNet\UDB3\Mailinglist\SubscribeUserToMailinglistRequestHandler;
 use CultuurNet\UDB3\UiTPASService\Controller\AddCardSystemToEventRequestHandler;
 use CultuurNet\UDB3\UiTPASService\Controller\DeleteCardSystemFromEventRequestHandler;
 use CultuurNet\UDB3\UiTPASService\Controller\GetCardSystemsFromEventRequestHandler;
@@ -219,6 +220,8 @@ final class PsrRouterServiceProvider extends AbstractServiceProvider
                 $this->bindUiTPASLabels($router);
 
                 $this->bindUiTPASOrganizers($router);
+
+                $this->bindMailinglist($router);
 
                 // Proxy GET requests to /events, /places, /offers and /organizers to SAPI3.
                 $router->get('/events/', ProxyRequestHandler::class);
@@ -579,5 +582,10 @@ final class PsrRouterServiceProvider extends AbstractServiceProvider
         $router->group('uitpas/organizers', function (RouteGroup $routeGroup): void {
             $routeGroup->get('{organizerId}/card-systems/', GetCardSystemsFromOrganizerRequestHandler::class);
         });
+    }
+
+    private function bindMailinglist(Router $router): void
+    {
+        $router->put('mailinglist/{email}/{mailingListId}/', SubscribeUserToMailinglistRequestHandler::class);
     }
 }
--- a/features/mailinglist/put.feature
+++ b/features/mailinglist/put.feature
@@ -1,0 +1,13 @@
+Feature: Test the UDB3 labels API
+
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I send and accept "application/json"
+
+  Scenario: Subscribe to newsletter
+    When I create a random name of 10 characters
+    And I send a PUT request to "mailinglist/%{name}@test.be/1746977"
+    Then the response status should be "200"
+    And the JSON response at "status" should be "ok"
--- a/src/Mailinglist/Client/MailingJetClient.php
+++ b/src/Mailinglist/Client/MailingJetClient.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailinglist\Client;
+
+use Mailjet\Client;
+use Mailjet\Resources;
+
+class MailingJetClient implements MailinglistClient
+{
+    private Client $client;
+
+    public function __construct(string $apiKey, string $apiSecret)
+    {
+        $this->client = new Client(
+            $apiKey,
+            $apiSecret
+        );
+    }
+
+    public function subscribe(string $email, string $mailingListId): void
+    {
+        $mailjetResponse = $this->client->post(
+            Resources::$ContactslistManagecontact,
+            [
+                'id' => $mailingListId,
+                'body' => [
+                    'Email' => $email,
+                    'Action' => 'addnoforce',
+                ],
+            ]
+        );
+
+        if (!$mailjetResponse->success()) {
+            throw new MailinglistSubscriptionFailed($mailjetResponse->getReasonPhrase());
+        }
+    }
+}
--- a/src/Mailinglist/Client/MailinglistClient.php
+++ b/src/Mailinglist/Client/MailinglistClient.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailinglist\Client;
+
+interface MailinglistClient
+{
+    public function subscribe(string $email, string $mailingListId): void;
+}
--- a/src/Mailinglist/Client/MailinglistSubscriptionFailed.php
+++ b/src/Mailinglist/Client/MailinglistSubscriptionFailed.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailinglist\Client;
+
+class MailinglistSubscriptionFailed extends \DomainException
+{
+}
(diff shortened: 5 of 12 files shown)


app/Http/PsrRouterServiceProvider.php
```diff
@@ -580,4 +583,9 @@ private function bindUiTPASOrganizers(Router $router): void
             $routeGroup->get('{organizerId}/card-systems/', GetCardSystemsFromOrganizerRequestHandler::class);
         });
     }
+
+    private function bindMailinglist(Router $router): void
+    {
+        $router->put('mailinglist/{email}/{mailingListId}/', SubscribeUserToMailinglistRequestHandler::class);
```
LucWollants, line 589: The order of the parts of the URL path is a bit weird, `mailinglist/{mailingListId}/{email}` seems more logical.
Even better would be `PUT mailinglist/{mailingListId}` and the body contains the email.

If the front end is the only consumer, it could be worth changing?

src/Mailinglist/Client/MailinglistClient.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailinglist\Client;
+
+interface MailinglistClient
+{
+    public function subscribe(string $email, string $mailingListId): void;
```
LucWollants, line 9: There is a `EmailAddress` value object, maybe good to use it to have a bit of validation?

features/mailinglist/put.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Test the UDB3 labels API
+
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I send and accept "application/json"
+
+  Scenario: Subscribe to newsletter
```
JonasVHG, line 9: I am in doubt this needs a feature test, because (like we discussed last Friday) this pollutes a testing environment.  If we decide to keep, I would also include an unhappy path with an email like `mailformedmailadres@` or something like that.

LucWollants:
Once this is deployed we also need a cleanup task. Archiving the repositories, cleaning up Jenkins and the server, etc.

src/Mailinglist/SubscribeUserToMailinglistRequestHandler.php
```diff
@@ -0,0 +1,42 @@
...
+        $mailingListId = (new RouteParameters($request))->get('mailingListId');
+
+        try {
+            $this->client->subscribe($email, $mailingListId);
+        } catch (MailinglistSubscriptionFailed $e) {
+            $this->logger->error('Failed to subscribe to newsletter: ' . $e->getMessage());
+            return new JsonResponse(['status' => $e->getMessage()], StatusCodeInterface::STATUS_BAD_REQUEST);
+        }
+
+        return new JsonResponse(['status' => 'ok']);
```
LucWollants, line 40: A typical response for our API would be a [204](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204), like for example on all [organizer endpoints](https://publiq.stoplight.io/docs/uitdatabank/entry-api/reference/operations/update-a-organizer-label).

If you think about it, what is the added value of returning a body if the status code is already indicating that the call was successful.

src/Mailinglist/Client/MailingJetClient.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function __construct(string $apiKey, string $apiSecret)
+    {
+        $this->client = new Client(
+            $apiKey,
+            $apiSecret
+        );
+    }
+
+    public function subscribe(string $email, string $mailingListId): void
```
JonasVHG, line 22: Maybe we could add a check to see if `$email` is a valid e-mail, I'm sure Mailjet does it too, but maybe it's best we do not create too much errors on the external server.

src/Mailinglist/Client/MailinglistSubscriptionFailed.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailinglist\Client;
+
+class MailinglistSubscriptionFailed extends \DomainException
```
JonasVHG, line 7: For the front-end, it might be better to work with an `ApiProblem` for consistency?

src/Mailinglist/Client/MailinglistClient.php
grubolsch: The old code didn't do any validation, and the frontend does not check any response, it does it own validation of the e-mail.
So I am happy to add it but without frontend work it would be quite pointless, and would never happen because there is already a JS check done before.

src/Mailinglist/Client/MailingJetClient.php
LucWollants: I think that makes sense, see https://github.com/cultuurnet/udb3-backend/pull/1686#discussion_r1565558363

src/Mailinglist/Client/MailinglistSubscriptionFailed.php
LucWollants: If we add API calls to `udb3-backend` we should indeed try to apply the guides and rules that were decided on.

src/Mailinglist/Client/MailinglistClient.php
LucWollants: Why is it pointless to add validation to an API that is public?

features/mailinglist/put.feature
grubolsch: Added a tag, and changed the Makefile so it auto skips this test

src/Mailinglist/Client/MailingJetClient.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailinglist\Client;
+
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddress;
+use Mailjet\Client;
+use Mailjet\Resources;
+
+class MailingJetClient implements MailinglistClient
```
LucWollants, line 11: The name of the tool is Mailjet

src/Http/ApiProblem/ApiProblem.php
```diff
@@ -552,4 +552,14 @@ public static function invalidJsonData(string $detail): self
             $detail
         );
     }
+
+    public static function failedToSubscribeToNewsletter(string $detail): self
+    {
+        return self::create(
+            'https://api.publiq.be/probs/body/failed-subscription-to-newsletter',
```
LucWollants, line 559: I think "body" is not a good classification, not sure if we have a better one already.

src/Mailinglist/SubscribeUserToMailinglistRequestHandler.php
```diff
@@ -0,0 +1,61 @@
...
+            throw ApiProblem::bodyInvalidSyntax('json');
+        }
+
+        if (empty($body['email'])) {
+            throw ApiProblem::requiredFieldMissing('email');
+        }
+
+        try {
+            $this->client->subscribe(new EmailAddress($body['email']), $mailingListId);
+        } catch (InvalidArgumentException $e) {
```
LucWollants, line 52: I guess it is possible to just catch all exceptions here?

tests/Mailinglist/SubscribeUserToMailinglistRequestHandlerTest.php
```diff
@@ -0,0 +1,105 @@
...
+        );
+
+        $this->request = (new Psr7RequestBuilder())
+            ->withUriFromString('/mailinglist/' . self::MAILINGLIST_ID)
+            ->withRouteParameter('mailingListId', (string) self::MAILINGLIST_ID)
+            ->withJsonBodyFromArray(['email' => self::EMAIL])
+            ->build('PUT');
+    }
+
+    public function testHandleSuccess(): void
```
LucWollants, line 56: In udb3-backend we still use it_handles_subscribing

tests/Mailinglist/SubscribeUserToMailinglistRequestHandlerTest.php
```diff
@@ -0,0 +1,105 @@
...
+    private $mailinglistClient;
+
+    /**
+     * @var MockObject|LoggerInterface
+     */
+    private $logger;
+
+    private ServerRequestInterface $request;
+
+    protected function setUp(): void
```
LucWollants, line 39: I think this misses a test for when the MailinglistClient would throw an exception

tests/Mailinglist/SubscribeUserToMailinglistRequestHandlerTest.php
```diff
@@ -0,0 +1,105 @@
...
+
+    public function testHandleFailureInvalidEmail(): void
+    {
+        $request = (new Psr7RequestBuilder())
+            ->withUriFromString('/mailinglist/' . self::MAILINGLIST_ID)
+            ->withRouteParameter('mailingListId', (string) self::MAILINGLIST_ID)
+            ->withJsonBodyFromArray(['email' => 'koen'])
+            ->build('PUT');
+
+        $this->assertCallableThrowsApiProblem(ApiProblem::failedToSubscribeToNewsletter('Given string is not a valid e-mail address.'), function () use ($request) {
```
LucWollants, line 75: This should be an invalid body API problem

simon-debruijn (approved):

src/Mailinglist/SubscribeUserToMailinglistRequestHandler.php
grubolsch: Not the ApiProblems, because they are processed elsewhere, but I am catching the error exceptions now.

PR #1687, opened 2024-04-16
III-6077 Added basic search on requested ownerships

LucWollants:
### Added
- Added endpoint `GET /ownerships?state=requested` to search on all ownerships that have state `requested`

### Note
- Limit and offset are not yet taken into account
- The endpoint response is a flat array, in a follow-up pull request this will be changed to return a `PagedCollection` and take the offset and limit into account

---
Ticket: https://jira.uitdatabank.be/browse/III-6077


tests/Ownership/Readmodels/OwnershipSearchProjectorTest.php
```diff
@@ -42,26 +45,83 @@ public function it_handles_ownership_requested(): void
...
+            ->with(
+                $ownershipDeleted->getId(),
+                OwnershipState::deleted()
+            );
+
+        $this->ownershipSearchProjector->handle($this->createDomainMessage($ownershipDeleted));
+    }
+
+    /**
+     * @param OwnershipRequested|OwnershipApproved|OwnershipRejected|OwnershipDeleted $event
```
JonasVHG, line 115: This is the only part of the PR, that made me doubt, Chaining 4 classes seems a bit too much to me.
But since this is the only place it's needed, introducing a BaseClass would not be a good idea.

JonasVHG (approved):
Looks great!

tests/Ownership/Readmodels/OwnershipSearchProjectorTest.php
LucWollants: I could just remove the type hint because it is inside a unit test. In a normal class I would make a factory method for all the different types, but that is overkill here.

PR #1693, opened 2024-04-17
Make the ApiProblem for Rdf Creation more clear

grubolsch:
### Changed
I changed the ApiProblem::invalidJsonData method name to more clearly reflect this does not contain generic invalid json errors, but is specific for the RDF creation.

### DIFF
--- a/src/Http/ApiProblem/ApiProblem.php
+++ b/src/Http/ApiProblem/ApiProblem.php
@@ -546,7 +546,7 @@ final class ApiProblem extends Exception
     public static function invalidJsonDataForRdfCreation(string $detail): self
     {
         return self::create(
-            'https://api.publiq.be/probs/body/invalid-json-data',
+            'https://api.publiq.be/probs/body/invalid-json-data-for-rdf-creation',
             'Invalid JSON data for RDF creation',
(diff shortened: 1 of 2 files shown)


src/Http/ApiProblem/ApiProblem.php
```diff
@@ -543,7 +543,7 @@ public static function imageMustBeLinkedToResource(string $imageId): self
         );
     }
 
-    public static function invalidJsonData(string $detail): self
+    public static function invalidJsonDataForRdfCreation(string $detail): self
     {
         return self::create(
             'https://api.publiq.be/probs/body/invalid-json-data',
```
LucWollants, line 549: Should this URL match the new naming better?

grubolsch: Change the URI.
However this error is not documented if you actually follow the URL, would you like me to document this error or is this not needed because internal?

simon-debruijn (approved):

src/Http/ApiProblem/ApiProblem.php
LucWollants: This is indeed an internal error, so I would not expose it for now.

LucWollants (approved):

PR #1698, opened 2024-05-07
III-6079 moviefetcher image

JonasVHG:
### Changed

- `KinepolisClient`: Added  `getImage()` to fetch the Image from a movie
- `ParsedMovie`: Added `$poster` as property
- `KinepolisService`: Upload Poster as Image to `UiTdatabank`, when creating a new movie.

### Fixed

- Movies now have an Image.

---

### DIFF
--- a/src/Kinepolis/AuthenticatedKinepolisClient.php
+++ b/src/Kinepolis/AuthenticatedKinepolisClient.php
@@ -75,9 +75,7 @@ final class AuthenticatedKinepolisClient implements KinepolisClient
 
     public function getImage(string $token, string $path): UploadedFile
     {
-        $lastSlashPosition = strrpos($path, '/');
-
-        $fileName = substr($path, $lastSlashPosition + 1);
+        $fileName = substr($path, strrpos($path, '/') + 1);
 
         $headers = $this->createHeaders($token);
         $headers['Accept'] = 'image/jpeg';
--- a/src/Kinepolis/AuthenticatedKinepolisClient.php
+++ b/src/Kinepolis/AuthenticatedKinepolisClient.php
@@ -6,6 +6,7 @@ namespace CultuurNet\UDB3\Kinepolis;
 
 use CultuurNet\UDB3\Json;
 use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\UploadedFile;
 use Psr\Http\Client\ClientInterface;
 
 final class AuthenticatedKinepolisClient implements KinepolisClient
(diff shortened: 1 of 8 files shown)


src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +73,33 @@ public function getMovieDetail(string $token, int $mid): array
         return $contents['movies'][0];
     }
 
+    public function getImage(string $token, string $path): UploadedFile
+    {
+        $lastSlashPosition = strrpos($path, '/');
```
grubolsch, line 78: Variable can move inline

src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +73,33 @@ public function getMovieDetail(string $token, int $mid): array
...
+
+        $fileName = substr($path, $lastSlashPosition + 1);
```
grubolsch, line 80: I would move this a separate function with a function name that explains what / why you are doing this.

src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +73,33 @@ public function getMovieDetail(string $token, int $mid): array
...
+        $headers['Accept'] = 'image/jpeg';
+        $request = new Request(
+            'GET',
+            $this->movieApiBaseUrl . $path,
+            $headers
+        );
+
+        $response = $this->client->sendRequest($request);
+        $imageStream = $response->getBody();
+        $mimeType = $response->getHeaderLine('Content-Type');
```
grubolsch, line 92: Can be moved inline

grubolsch (approved):

src/Kinepolis/ParsedMovie.php
```diff
@@ -27,14 +27,17 @@ final class ParsedMovie
...
     public function __construct(
         string $externalId,
         Title $title,
         LocationId $locationId,
         Description $description,
         Theme $theme,
         Calendar $calendar,
-        PriceInfo $priceInfo
+        PriceInfo $priceInfo,
+        string $poster
```
LucWollants, line 40: What is poster? Is that the URL of the image?

JonasVHG: Indeed,  I kept the naming consistent with the naming in the JSON API from Kinepolis.

src/Kinepolis/AuthenticatedKinepolisClient.php
```diff
@@ -72,6 +73,33 @@ public function getMovieDetail(string $token, int $mid): array
...
+
+        return new UploadedFile(
```
LucWollants, line 94: I'm a bit confused by the usage of the `UploadedFile` class here. Normally an `UploadedFile` is something that was uploaded to the UDB3 server and is physically present on the file system. In this case, it is a stream from another online URL, will this work?

JonasVHG: I tested it locally, and it worked, I will have to test it again, when it is on the server.

src/Kinepolis/KinepolisService.php
```diff
@@ -173,6 +182,19 @@ private function createNewMovie(ParsedMovie $parsedMovie): string
...
+        // to avoid multiple needless updates, we only add an image during the creation
+        $uploadedImage = $this->client->getImage($token, $parsedMovie->getPoster());
+        $imageId = $this->imageUploader->upload(
+            $uploadedImage,
+            new MediaDescription($parsedMovie->getTitle()->toString()),
+            new CopyrightHolder('Kinepolis'),
+            new LegacyLanguage('nl')
+        );
+        $addImage = new AddImage($eventId, $imageId);
+        $this->commandBus->dispatch($addImage);
```
LucWollants, line 196: It's a bit weird that the `createNewMovie` also does the dispatch of the `AddImage` command. Maybe that needs to become a separate method that can be called inside the `process` method. That new method can do the upload of the file and returns the `AddImage` command.

src/Kinepolis/ParsedMovie.php
LucWollants: I have no strong opinion about it, but is image more clear in the UDB3 context?

src/Kinepolis/AuthenticatedKinepolisClient.php
LucWollants: Then it will be fine because it is a stream object.

LucWollants (approved):

PR #1700, opened 2024-05-15
III-6175 migrate moviedata

JonasVHG:
### Added

- `ImportMovieIdsFromCsv`: `Command` to import movieIds from the legacy application via a `csv`.

### Changed

- `ConsoleServiceProvider`: Register new Command `ImportMovieIdsFromCsv`

---
Ticket: https://jira.publiq.be/browse/III-6175

### DIFF
--- a/app/Console/Command/ImportMovieIdsFromCsv.php
+++ b/app/Console/Command/ImportMovieIdsFromCsv.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use CultuurNet\UDB3\Kinepolis\MovieMappingRepository;
+use League\Csv\Reader;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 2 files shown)


app/Console/Command/ImportMovieIdsFromCsv.php
```diff
@@ -0,0 +1,52 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $csvReader = Reader::createFromPath(
+            $input->getArgument(self::CSV_FILE)
+        );
+
+        $records = $csvReader->getRecords();
+        $output->writeln('Starting import.');
+        foreach ($records as $record) {
+            $this->mappingRepository->create($record[0], $record[1]);
```
LucWollants, line 45: Should validation be done to verify that the event exists inside Redis or inside the event store?

JonasVHG: The legacy application worked via EntryAPI, so if we find a `uuid` that doesn't exist in the `event_store` something would be really wrong, still might be worth it.

LucWollants (approved):

PR #1704, opened 2024-05-21
III-6079 BUGFIX: moviefetcher publish

JonasVHG:
### Changed

- `KinpelosService`:  Dispatch a `Publish`, when creating a new `Movie`

### Fixed

- New movies are no longer in `DRAFT`

---

### DIFF
--- a/src/Kinepolis/KinepolisService.php
+++ b/src/Kinepolis/KinepolisService.php
@@ -10,6 +10,7 @@ use Broadway\UuidGenerator\UuidGeneratorInterface;
 use CultuurNet\UDB3\Calendar\Calendar as LegacyCalendar;
 use CultuurNet\UDB3\Description;
 use CultuurNet\UDB3\Event\Commands\AddImage;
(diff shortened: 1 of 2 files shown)


src/Kinepolis/KinepolisService.php
```diff
@@ -152,6 +153,7 @@ private function process(ParsedMovie $parsedMovie, string $token): void
 
         if ($eventId === null) {
             $eventId = $this->createNewMovie($parsedMovie);
+            $commands[] = new Publish($eventId);
```
LucWollants, line 156: Is the order important? Should it be the last command?

LucWollants (approved):

simon-debruijn (approved):

src/Kinepolis/KinepolisService.php
JonasVHG: I don't see why it would matter, you can update events regardless they are in `DRAFT` or `READY_FOR_VALIDATION`. The Normal EntryAPI also does it as a first `Command`.

PR #1705, opened 2024-05-21
III-6079 BUGFIX: link movie to production

JonasVHG:
### Changed

-  `KinepolisService`: newly created movies are linked to a `production`, if no matching `production` is found, a new one is created for the movie.

### Fixed

- Movies are no correctly linked to a production.

---

### DIFF
--- a/src/Kinepolis/KinepolisService.php
+++ b/src/Kinepolis/KinepolisService.php
@@ -165,7 +165,7 @@ final class KinepolisService
             $addImage = $this->uploadImage($token, $parsedMovie, $eventId);
             $commands[] = $addImage;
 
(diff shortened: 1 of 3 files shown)


src/Kinepolis/KinepolisService.php
```diff
@@ -228,4 +240,14 @@ private function uploadImage(string $token, ParsedMovie $parsedMovie, string $ev
         );
         return new AddImage($eventId, $imageId);
     }
+
+    private function linkToProduction(string $title): ProductionId
```
LucWollants, line 244: This searches the production rather than doing the actual linking which happens with the command.

LucWollants (approved):

PR #1709, opened 2024-05-24
III-6181 Movies with empty description

JonasVHG:
### Changed

- Refactoring: Renamed `Parser` to `MovieParser` to better differentiate between the different Parsers.
- `ParsedMovie`: Make `Description` nullable
- `KinepolisMovieParser`: handle Empty Description
- `KinepolisService`: do not dispatch a `UpdateDescription` when `Description` is `null`

- `KinepolisMovieParserTest`:  Added test for `ParsedMovie` with empty Description
- `KinepolisServiceTest`: Added test for `ParsedMovie` with empty Description

### Fixed

- The `KinepolisService` now handles movies with an empty description

---
Ticket: https://jira.publiq.be/browse/III-6181

### DIFF
--- a/src/Kinepolis/ValueObject/ParsedMovie.php
+++ b/src/Kinepolis/ValueObject/ParsedMovie.php
@@ -19,7 +19,7 @@ final class ParsedMovie
 
     private LocationId $locationId;
 
-    private Description $description;
+    private ?Description $description;
 
     private Theme $theme;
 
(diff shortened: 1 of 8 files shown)


src/Kinepolis/ValueObject/ParsedMovie.php
```diff
@@ -33,7 +33,7 @@ public function __construct(
         string $externalId,
         Title $title,
         LocationId $locationId,
-        Description $description,
+        ?Description $description,
```
LucWollants, line 36: Another approach is only to have the required or non-nullable parameters inside the constructor and add a method `withDescription` which does the typical cloning, setting, and returning.

LucWollants (approved):
I already approved. Adding the `withDescription` method is not "blocking".

PR #1710, opened 2024-05-27
move getDuplicatePlacesRemovedFromCluster outside cluster loop

grubolsch:
### Fixed

- Move `getDuplicatePlacesRemovedFromCluster` outside the cluster loop, it is now generating way too many messages on the queue.

### DIFF
--- a/app/Console/Command/ProcessDuplicatePlaces.php
+++ b/app/Console/Command/ProcessDuplicatePlaces.php
@@ -131,20 +131,8 @@ final class ProcessDuplicatePlaces extends AbstractCommand
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -170,7 +158,12 @@ public function execute(InputInterface $input, OutputInterface $output): int
             }
         }
 
-        return 0;
+        if (! $onlySetCanonical) {
```
LucWollants, line 161: We also need to test for `dryRun` here?

LucWollants: I see now that it is handled inside the new method `reindexPlaces`

LucWollants (approved):

PR #1717, opened 2024-06-14
III-6212 Bootstrap JWT V2 validator dynamically for Keycloak

LucWollants:
### Changed
- Bootstrap JWT V2 validator dynamically for Keycloak

### Note
- Requires config change https://github.com/cultuurnet/appconfig/pull/671

---
Ticket: https://jira.uitdatabank.be/browse/III-6212

### DIFF
--- a/app/Authentication/AuthServiceProvider.php
+++ b/app/Authentication/AuthServiceProvider.php
@@ -20,6 +20,7 @@ use CultuurNet\UDB3\Impersonator;
 use CultuurNet\UDB3\Role\UserPermissionsServiceProvider;
 use CultuurNet\UDB3\Role\ValueObjects\Permission;
 use CultuurNet\UDB3\User\CurrentUser;
+use League\Container\DefinitionContainerInterface;
 
 final class AuthServiceProvider extends AbstractServiceProvider
 {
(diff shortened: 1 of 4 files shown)


app/Authentication/AuthServiceProvider.php
```diff
@@ -180,4 +177,15 @@ static function () use ($container): ?Consumer {
             fn () => new Impersonator()
         );
     }
+
+    private function createUitIdV2JwtValidator(DefinitionContainerInterface $container): UitIdV2JwtValidator
+    {
+        $version = $container->get('config')['keycloak']['enabled'] ? 'keycloak' : 'v2';
```
simon-debruijn, line 183: Can't we replace the technology specific naming of keycloak by something else?

It makes sense to have a KEYCLOAK_ENABLED toggle. But maybe other config can be less specific

simon-debruijn (approved):

app/Authentication/AuthServiceProvider.php
LucWollants: I was also thinking about this, but then I thought `v1 and `v2` are more confusing than `uitid`, `auth0`, and `keycloak. It's only needed to know this when bootstrapping, for the rest of the code this is somewhat hidden.

JonasVHG (approved):

PR #1728, opened 2024-07-08
III-6228 Create script that can move events

grubolsch:
### Added
I added a CLI script to move events to another place based on a SAPI3 query.

---
Ticket: https://jira.uitdatabank.be/browse/III-6228

### DIFF
--- a/app/Console/Command/MoveEvents.php
+++ b/app/Console/Command/MoveEvents.php
@@ -94,7 +94,7 @@ class MoveEvents extends AbstractCommand
                 $this->commandBus->dispatch($command);
                 $output->writeln('Dispatching UpdateLocation for event with id ' . $command->getItemId());
             } catch (Exception $exception) {
-                $exceptions[] = sprintf("Event with id: %s caused an exception: %s", $event->getId(), $exception->getMessage());
+                $exceptions[] = sprintf('Event with id: %s caused an exception: %s', $event->getId(), $exception->getMessage());
             }
 
             $progressBar->advance();
@@ -104,7 +104,7 @@ class MoveEvents extends AbstractCommand
 
         $output->writeln('');
         foreach ($exceptions as $exception) {
-            $output->writeln(sprintf("<error>%s</error>", $exception));
+            $output->writeln(sprintf('<error>%s</error>', $exception));
         }
 
         return self::SUCCESS;
--- a/app/Console/Command/MoveEvents.php
+++ b/app/Console/Command/MoveEvents.php
@@ -1,0 +1,130 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Event\Commands\UpdateLocation;
+use CultuurNet\UDB3\Event\ValueObjects\LocationId;
+use CultuurNet\UDB3\Search\ResultsGenerator;
+use CultuurNet\UDB3\Search\ResultsGeneratorInterface;
+use CultuurNet\UDB3\Search\SearchServiceInterface;
+use CultuurNet\UDB3\Search\Sorting;
+use Exception;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+/**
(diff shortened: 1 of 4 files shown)


app/Console/Command/MoveEvents.php
```diff
@@ -0,0 +1,130 @@
...
+                self::QUERY,
+                null,
+                'SAPI3 query for which events to move.'
+            )
+            ->addOption(
+                self::FORCE,
+                null,
+                InputOption::VALUE_NONE,
+                'Skip confirmation.'
+            );
```
JonasVHG, line 64: Maybe a dry-run option can be usefull especially since a sapi3-query is rather powerful?

JonasVHG (approved):

app/Console/Command/MoveEvents.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): ?int
+    {
+        $placeUuid = $input->getArgument(self::PLACE_UUID);
+        $query = $input->getArgument(self::QUERY);
+
+        if ($placeUuid === null || $query === null) {
+            $output->writeln('<error>Missing argument, the correct syntax is: event:move "place_uuid_to_move_to" "sapi3_query"</error>');
```
simon-debruijn, line 80: Can we use an error method instead of `->writeln('<error>`?

tests/Console/Command/MoveEventsTest.php
```diff
@@ -0,0 +1,106 @@
...
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+
+class MoveEventsTest extends TestCase
+{
+    private TraceableCommandBus $commandBus;
+
+    /**
+     * @var SearchServiceInterface|(SearchServiceInterface&object&MockObject)|(SearchServiceInterface&MockObject)|(object&MockObject)|MockObject
```
simon-debruijn, line 28: Can't this just be `SearchServiceInterface&MockObject` or `SearchServiceInterface|MockObject`?

simon-debruijn (approved):

app/Console/Command/MoveEvents.php
grubolsch: I have no knowledge of this method. Can you give an example?

app/Console/Command/MoveEvents.php
```diff
@@ -0,0 +1,139 @@
...
+            new Sorting('created', 'asc'),
+            100
+        );
+    }
+
+    public function configure(): void
+    {
+        $this
+            ->setName('event:move')
+            ->setDescription('Move an event from one place to another based on a SAPI3 query')
```
LucWollants, line 49: I would write something like "Update the location of all events from the given SAPI3 query to the given new location"

app/Console/Command/MoveEvents.php
```diff
@@ -0,0 +1,139 @@
...
+        if ($placeUuid === null || $query === null) {
+            $output->writeln('<error>Missing argument, the correct syntax is: event:move "place_uuid_to_move_to" "sapi3_query"</error>');
+            return self::FAILURE;
+        }
+
+        $count = $this->searchResultsGenerator->count($query);
+
+        if ($count <= 0) {
+            $output->writeln('<error>No events found</error>');
+            return self::FAILURE;
```
LucWollants, line 88: Not sure if this is a failure, same remark applies for line 92

app/Console/Command/MoveEvents.php
```diff
@@ -0,0 +1,139 @@
...
+        $progressBar = new ProgressBar($output, $count);
+
+        $exceptions = [];
+        foreach ($this->searchResultsGenerator->search($query) as $event) {
+            try {
+                $command = new UpdateLocation($event->getId(), new LocationId($placeUuid));
+                if (!$input->getOption(self::DRY_RUN)) {
+                    $this->commandBus->dispatch($command);
+                }
+                $output->writeln('Dispatching UpdateLocation for event with id ' . $command->getItemId());
```
LucWollants, line 104: This should be written before the dispatch, otherwise there is no output when an exception happens.

app/Console/Command/MoveEvents.php
```diff
@@ -0,0 +1,139 @@
...
+            } catch (Exception $exception) {
+                $exceptions[] = sprintf('Event with id: %s caused an exception: %s', $event->getId(), $exception->getMessage());
```
LucWollants, line 106: I would prefer to write the exception when it occurs so it is possible to abort the command when running it supervised.

PR #1729, opened 2024-07-08
III-6124 Include completness in canonical, when multiple museum places

grubolsch:
### Changed
Currently when we calculate a canonical for the duplicate places clusters, we skip clusters with multiple MuseumPass places.
Now that we have the "completeness" inside the backend of UiTdatabank, it can be very useful to use that and to make sure we use data rich places as a canonical.
So we are going to take the place with the highest completeness in these cases and select it as canonical.

Comes with a test :-)

---
Ticket: https://jira.uitdatabank.be/browse/III-6124

### DIFF
--- a/src/Place/Canonical/CanonicalService.php
+++ b/src/Place/Canonical/CanonicalService.php
@@ -45,7 +45,7 @@ class CanonicalService
             return $placesWithMuseumpas[array_key_first($placesWithMuseumpas)];
         }
         if (count($placesWithMuseumpas) > 1) {
-            throw new MuseumPassNotUniqueInCluster($clusterId, count($placesWithMuseumpas));
+            return $this->getPlaceWithHighestCompleteness($placesWithMuseumpas);
         }
 
         $placesWithMostEvents = $this->getPlacesWithMostEvents($placeIds);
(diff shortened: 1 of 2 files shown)


src/Place/Canonical/CanonicalService.php
```diff
@@ -94,4 +93,23 @@ private function getOldestPlace(array $placeIds): string
...
+
+    private function getPlaceWithHighestCompleteness(array $placeIds): string
+    {
+        $maxCompleteness = -1;
+        $placeWithMaxCompleteness = '';
+
+        foreach ($placeIds as $placeId) {
+            $jsonDocument = $this->placeRepository->fetch($placeId);
+            $body = $jsonDocument->getBody();
+            $completeness = $body->completeness;
```
JonasVHG, line 105: There are still 129 places without a `completeness`-score, but since none of them are `Museumpas`, a fallback might not be needed.

JonasVHG (approved):

src/Place/Canonical/CanonicalService.php
grubolsch: That is really interesting to know, thank you!

PR #1730, opened 2024-07-08
[performance] Improve performance of canonical sorting methods

grubolsch:
## Context
I was in this class for another ticket and saw some performance opportunities.

### Changed
We currently first loop trough all places for both `getPlacesWithMostEvents()`  and `getOldestPlace()`, and then sort the arrays based on some field. 
By directly comparing event counts, we avoid the need to sort the array, reducing complexity, and also lowering the RAM usage because we don't build a tmp array to sort and then throw away..

Because sorting in PHP uses quicksort, which is O(n log n). 
This means this change reduces the complexity to a linear O(n)

### DIFF
--- a/src/Place/Canonical/CanonicalService.php
+++ b/src/Place/Canonical/CanonicalService.php
@@ -68,30 +68,42 @@ class CanonicalService
 
     private function getPlacesWithMostEvents(array $placeIds): array
     {
-        $eventCountsWithPlaceId = [];
+        $maxEventCount = -1;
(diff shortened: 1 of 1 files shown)


LucWollants (approved):
Nice improvement.

I must admit that with changes like these I always like to see timings to avoid premature performance optimisations.

JonasVHG (approved):

simon-debruijn (approved):

PR #1732, opened 2024-07-16
III-5990 incorrect events in Production

JonasVHG:
### Added

- `BulkRemoveFromProduction`: CLI-Command, to bulk remove Events from a production, without overloading the queues.
- `RemoveEventsFromProduction`: Command to remove multiple Events from a production.

### Changed

- `ConsoleServiceProvider`: Registered `BulkRemoveFromProduction`
- `CopyEventHandler`: Make it optional to add a copied event to the same production as the original.
- `EventCommandHandlerProvider`: `Updated constructor` of `CopyEventHandler`
- `ProductionCommandHandler`: added handle for `RemoveEventsFromProduction`
- `ProductionRepository` & implementations `DBALProductionRepository` & `BroadcastingProductionRepository` added `removeEvents()`

### Fixed

- When a production has many incorrect events, we can now bulk remove those events without overloading the queues.
Also to see if it investigate if it will diminish the incorrect events, adding a copied Event to the same production as the original has been made optional.
- Improved Codestyle for several files

### Related PR

- https://github.com/cultuurnet/appconfig/pull/719

---
Ticket: https://jira.publiq.be/browse/III-5990

### DIFF
--- a/src/Event/CommandHandlers/CopyEventHandler.php
+++ b/src/Event/CommandHandlers/CopyEventHandler.php
@@ -17,12 +17,16 @@ final class CopyEventHandler implements CommandHandler
     private EventRepository $eventRepository;
     private ProductionRepository $productionRepository;
 
+    private bool $copyProduction;
+
     public function __construct(
         EventRepository $eventRepository,
-        ProductionRepository $productionRepository
+        ProductionRepository $productionRepository,
+        bool $copyProduction
     ) {
         $this->eventRepository = $eventRepository;
         $this->productionRepository = $productionRepository;
+        $this->copyProduction = $copyProduction;
     }
 
     public function handle($command): void
@@ -42,11 +46,13 @@ final class CopyEventHandler implements CommandHandler
 
         // Add the new event to the same production as the original one.
         // This is not done as part of Event::copy() because productions are not event-sourced.
-        try {
-            $production = $this->productionRepository->findProductionForEventId($originalEventId);
-            $this->productionRepository->addEvent($newEventId, $production);
-        } catch (EntityNotFoundException $e) {
-            return;
+        if ($this->copyProduction) {
+            try {
+                $production = $this->productionRepository->findProductionForEventId($originalEventId);
+                $this->productionRepository->addEvent($newEventId, $production);
+            } catch (EntityNotFoundException $e) {
+                return;
+            }
         }
     }
 }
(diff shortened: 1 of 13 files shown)


LucWollants (changes requested):
As discussed through Slack it's better to create a new command `RemoveEventsFromProduction` to avoid triggering a dispatch twice because of the broadcasting production repository

src/Event/Productions/DBALProductionRepository.php
```diff
@@ -77,6 +77,14 @@ public function removeEvent(string $eventId, ProductionId $productionId): void
         );
     }
 
+    /** @param string[] $eventIds */
+    public function removeEvents(array $eventIds, ProductionId $productionId): void
+    {
+        foreach ($eventIds as $eventId) {
+            $this->removeEvent($eventId, $productionId);
```
LucWollants, line 84: Would it be safer to create a query builder and remove all the rows at once so they are wrapped in a transaction and multiple DB calls are avoided?

src/Event/Productions/DBALProductionRepository.php
```diff
@@ -77,6 +77,20 @@ public function removeEvent(string $eventId, ProductionId $productionId): void
         );
     }
 
+    /** @param string[] $eventIds */
+    public function removeEvents(array $eventIds, ProductionId $productionId): void
+    {
+        (
```
LucWollants, line 83: Seems weird to have extra parenthesis here?

LucWollants (approved):
Looks good, thanks for the refactor. Only one minor remark, so I already approved.

src/Event/CommandHandlers/CopyEventHandler.php
```diff
@@ -42,11 +46,13 @@ public function handle($command): void
 
         // Add the new event to the same production as the original one.
         // This is not done as part of Event::copy() because productions are not event-sourced.
-        try {
-            $production = $this->productionRepository->findProductionForEventId($originalEventId);
-            $this->productionRepository->addEvent($newEventId, $production);
-        } catch (EntityNotFoundException $e) {
-            return;
+        if ($this->copyProduction) {
```
LucWollants, line 49: I missed this feature completely with the first review. This would have been something for a separate pull request. Also, the name is very confusing, also see https://github.com/cultuurnet/appconfig/pull/719/files#r1686143312

PR #1754, opened 2024-08-26
III-6255 sanitize trailer

JonasVHG:
### Added

- `YoutubeTrailerRepositoryTest`: Unit tests for `YoutubeTrailerRepository`.
- `YoutubeSearchResult.json` & `YoutubeNoResults.json`: Sample file of a `Youtube`-search Result to be used in tests.

### Changed

- `TrailerRepository`: rename `search()` to `findMatchingTrailer()` for clarity
- `YoutubeTrailerRepository`: it will sanitize a movie title, before searching for it.
- `YoutubeTrailerRepository`: Changed constructor, to directly inject a `Google_Service_YouTube` instead of a `developer_key` and construct one inside the `constructor`, to facilitate mocking.

### Fixed

- Movie Titles with exotic characters can now be searched for a trailer.

---
Ticket: https://jira.publiq.be/browse/III-6255


tests/Kinepolis/Trailer/YoutubeTrailerRepositoryTest.php
```diff
@@ -0,0 +1,119 @@
...
+            'maxResults' => 1,
+        ])->willReturn(
+            Json::decodeAssociatively(
+                SampleFiles::read(__DIR__ . '/../samples/YoutubeNoResults.json')
+            )
+        );
+        $this->uuidGenerator->expects($this->never())->method('generate');
+
+        $video = $this->trailerRepository->search('NotFound');
+        $this->assertNull($video);
```
simon-debruijn, line 89: To me it makes more sense, for this to always return a list of items.
If none are found, it returns an empty list

simon-debruijn (approved):

tests/Kinepolis/Trailer/YoutubeTrailerRepositoryTest.php
JonasVHG: The function name `search()` is indeed misleading, but the problem is that Kinepolis almost exclusively only has 1 trailer per movie, so we are never interested in the elements after the first result, because they are the wrong trailers. It either has the correct Trailer as a first searchResult or it doesn't.
Maybe renaming the function to findMatchingTrailer() would be more clear?

tests/Kinepolis/Trailer/YoutubeTrailerRepositoryTest.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @test
+     */
+    public function it_will_only_search_when_enabled(): void
+    {
+        $disabledTrailerRepository = new YoutubeTrailerRepository(
+            $this->youtubeClient,
+            $this->channelId,
+            $this->uuidGenerator,
+            false
```
LucWollants, line 64: Now that I see this code it's a bit weird that the repo can be disabled "internally". I think it makes more sense to put the disabling on another higher level and never call the any methods of the repo. But no need to change it in this pull request.

tests/Kinepolis/Trailer/YoutubeTrailerRepositoryTest.php
LucWollants: Another approach could be to throw something like `NoTrailerFound` and have a non-nullable return type.

LucWollants (approved):

tests/Kinepolis/Trailer/YoutubeTrailerRepositoryTest.php
JonasVHG: Yeah, but the original idea for this to be a temporary flag, to quickly disable the feature if we got weird results. Now that it works, I think it can safely removed during: https://jira.publiq.be/browse/III-6184 but I'll have to wait till Kristof gets back before I pickup that ticket, because I do not know how the deployment pipeline will handle removing the files in appconfig.

LucWollants: I guess we could already do the code part without the help of Kristof? The other parts of the ticket can be discussed during a backend meeting?

PR #1755, opened 2024-08-27
III-6258 - Remove service definitions that are only used once. Move those inline to where they are used.

simon-debruijn:
### Fixed

---
Ticket: https://jira.uitdatabank.be/browse/III-6258


simon-debruijn:
I could only find 1 or 2 services that only have one get call to the container.
Moving the creation inside the classes themselves (inline) pollutes the code.
Introducing factories instead has no added benefit over the current implementation.

PR #1758, opened 2024-08-27
III-6202 AddLabel for UiTPAS-Middenkust

JonasVHG:
### Added

- `AddLabelToItems`: new CLI-Command to add Labels to items via a query

### Changed

- `ConsoleServiceProvider`: Registered new CLI-Command

### Fixed

- Labels can now be edited via a CLI-Command to multiple items based on a Query

---
Ticket: https://jira.publiq.be/browse/III-6202


app/Console/Command/AddLabelToItems.php
```diff
@@ -0,0 +1,170 @@
...
+
+    private ResultsGenerator $organizersSearchResultsGenerator;
+
+    public function __construct(
+        CommandBus $commandBus,
+        SearchServiceInterface $eventsSearchService,
+        SearchServiceInterface $placesSearchService,
+        SearchServiceInterface $organizersSearchService
+    ) {
+        $this->eventsSearchResultsGenerator = new ResultsGenerator(
```
LucWollants, line 38: I think you can make an internal method that creates the generator based on the item type. Once you have this method you can use it inside the execute and avoid a lot of duplicated code.

LucWollants: The only thing that is different seems to be the injected search service. So a lot of duplication can be avoided, even the sorting and page size is the same.

app/Console/Command/AddLabelToItems.php
```diff
@@ -0,0 +1,170 @@
...
+                if ($itemType->sameAs(ItemType::organizer())) {
+                    $labelCommand = new OrganizerAddLabel($id, $label);
+                } else {
+                    $labelCommand = new OfferAddLabel($id, $label);
+                }
+                $this->commandBus->dispatch(
+                    $labelCommand
+                );
+            } catch (Exception $exception) {
+                $exceptions[$id] = 'Item with id: ' . $id . ' caused an exception: ' . $exception->getMessage();
```
LucWollants, line 160: I wonder if it's better to directly show the error, if we run the command in attended mode you can force quit it after the first failure. Otherwise, it could fail for all items but run for a very long time before showing any errors.

app/Console/Command/AddLabelToItems.php
```diff
@@ -0,0 +1,141 @@
...
+                $input,
+                $output,
+                new ConfirmationQuestion(
+                    sprintf('This action will add the label %s to %u %ss, continue? [y/N] ', $label->getName()->toString(), $count, $itemType->toString()),
+                    true
+                )
+            );
+    }
+
+    private function getSearchGenerator(ItemType $itemType): ResultsGenerator
```
LucWollants, line 104:
```suggestion
    private function getResultsGenerator(ItemType $itemType): ResultsGenerator
```

LucWollants (approved):

PR #1761, opened 2024-08-29
Added PHP 8.1 to test matrix

Anahkiasen:
### Added
- Added PHP 8.1 to test matrix

---
Ticket: https://jira.publiq.be/browse/III-5031


LucWollants (approved):
Looks good. Maybe we should run all RDF acceptance tests for this one?

Anahkiasen:
@LucWollants Is that something I can do before it's merged?

LucWollants:
> @LucWollants Is that something I can do before it's merged?

Yes, that is something we can do with the `make feature` command or with `make feature-filter path=features/event/rdf.feature`. There are also tests in the `place` and `organizer` folder.

PR #1766, opened 2024-09-02
III-6252 Change cluster id to string

grubolsch:
### Added
In preparation for the new delta logic of Duplicate places, cluster ids will be hashes (sha1) instead of ints.

---
Ticket: https://jira.uitdatabank.be/browse/III-6252


tests/Place/Canonical/CanonicalServiceTest.php
```diff
@@ -203,7 +203,7 @@ public function setUp(): void
      */
     public function it_will_return_the_MPM_place(): void
     {
-        $canonicalId = $this->canonicalService->getCanonical(1);
+        $canonicalId = $this->canonicalService->getCanonical('1');
```
JonasVHG, line 206: To avoid confusion I would sure, that everyone sees it's a `string`, so maybe also `cluster_1` like you did in `DBALDuplicatePlaceRepositoryTest.php` below.

JonasVHG (approved):

PR #1767, opened 2024-09-02
III-6266 [Duplicate places] Delete old clusters

grubolsch:
### Changed
This script compares 2 tables (duplicate_places vs duplicate_places_import).
if a cluster exists in duplicate_places but no longer in duplicate_places_import it should be removed.
We will then the place it in the duplicate_places_removed_from_cluster table  to queue up reindexing of the cluster.

---
Ticket: https://jira.uitdatabank.be/browse/[III-6266](https://jira.publiq.be/browse/III-6266


grubolsch:
Processed feedback discussed with Luc in slack.

app/Migrations/Version20240830071940.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/*
+ * Adding primary keys
```
LucWollants, line 11: No need for this comment because that is clear from the code.

app/Migrations/Version20240830071940.php
```diff
@@ -0,0 +1,32 @@
...
+ * */
+class Version20240830071940 extends AbstractMigration
```
LucWollants, line 13: I would advise to change the date, so it comes after the last migration to make sure it gets run.

src/Place/DuplicatePlace/Dto/ClusterRecord.php
```diff
@@ -0,0 +1,36 @@
...
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+
+class ClusterRecord
+{
+    private string $clusterId;
+
+    private UuidInterface $placeUuid;
+
+    public function __construct(string $clusterId, UuidInterface $placeUuid)
```
LucWollants, line 16: If this is a wrapper around a cluster I would expect it to have a list of places instead of just one place. This feels more like a `PlaceInCluster` row

src/Place/Canonical/DBALDuplicatePlaceRepository.php
```diff
@@ -78,4 +78,57 @@ public function getDuplicatesOfPlace(string $placeId): ?array
...
+            ->where('dpi.cluster_id IS NULL')
+            ->orderBy('dp.cluster_id', 'asc')
+            ->addOrderBy('dp.place_uuid', 'asc');
+
+        $statement = $qb->execute();
+
+        return $this->processRawToClusterRecord($statement->fetchAllAssociative());
+    }
+
+    public function countPlacesInDuplicatePlacesImport(string $placeId): int
```
LucWollants, line 98:
```suggestion
    public function countPlaceInDuplicatePlacesImport(string $placeId): int
```

src/Place/Canonical/DBALDuplicatePlaceRepository.php
```diff
@@ -78,4 +78,57 @@ public function getDuplicatesOfPlace(string $placeId): ?array
...
+            ->setParameter(':place_id', $placeId)
+            ->execute();
+
+        return $statement->fetchAssociative()['total'];
+    }
+
+    public function addToDuplicatePlacesRemovedFromCluster(string $placeId): void
+    {
+        $this->connection->createQueryBuilder()
+            ->insert('duplicate_places_removed_from_cluster')
```
LucWollants, line 113: I would create a new repo for this. There is also one query in `ProcessDuplicatePlaces` in the method `getDuplicatePlacesRemovedFromCluster` which can be moved inside this new repo.

src/Place/Canonical/DBALDuplicatePlaceRepository.php
```diff
@@ -78,4 +78,57 @@ public function getDuplicatesOfPlace(string $placeId): ?array
...
+            ->setValue('place_uuid', ':place_uuid')
+            ->setParameter(':place_uuid', $placeId)
+            ->execute();
+    }
+
+    private function processRawToClusterRecord(array $data): array
```
LucWollants, line 119:
```suggestion
    private function convertRawToClusterRecord(array $data): array
```

src/Place/Canonical/DBALDuplicatePlaceRepository.php
```diff
@@ -78,4 +78,57 @@ public function getDuplicatesOfPlace(string $placeId): ?array
 
         return count($duplicates) > 0 ? $duplicates : null;
     }
+
+    /** @return ClusterRecord[] */
+    public function calculateNoLongerInCluster(): array
```
LucWollants, line 83:
```suggestion
    public function getPlacesRemovedFromCluster(): array
```

src/Place/DuplicatePlace/ImportDuplicatePlacesProcessor.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->connection = $connection;
+    }
+
+    public function sync(): void
+    {
+        $noLongerInClusters = $this->duplicatePlaceRepository->calculateNoLongerInCluster();
+
+        $this->connection->beginTransaction();
+        foreach ($noLongerInClusters as $noLongerInCluster) {
+            $this->duplicatePlaceRepository->deleteCluster($noLongerInCluster->getClusterId());
```
LucWollants, line 29: This needs to be done only once I think, because you do this place by place and it only needs to happen once for a cluster

src/Place/DuplicatePlace/ImportDuplicatePlacesProcessor.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\DuplicatePlace;
+
+use CultuurNet\UDB3\Place\Canonical\DuplicatePlaceRepository;
+use Doctrine\DBAL\Connection;
+
+class ImportDuplicatePlacesProcessor
```
LucWollants, line 10: This class needs a test

src/Place/DuplicatePlace/ImportDuplicatePlacesProcessor.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Place\DuplicatePlace;
```
LucWollants, line 5: I would avoid having 2 namespaces for the same thing. We have a `Canonical` namespace and now also a `DuplicatePlace`. I would put everything in the `Canonical` namespace for now and in a next pull request you can rename that to `Duplicates`

LucWollants:
> Processed feedback discussed with Luc in slack.

Thanks for this 🙏!

I think the code is much easier to understand now. Can you also update the description and title so it matches the changes better?

src/Place/Canonical/DBALDuplicatePlaceRepository.php
grubolsch: It counts how many placeS are in the import table so is the current name not correct?

LucWollants: So that makes me wonder if `isPlaceInDuplicatePlacesImport(string $placeId): bool` is not the correct name?

src/Place/DuplicatePlace/Dto/ClusterRecord.php
grubolsch: This class has been deleted.

grubolsch:
All feedback has been processed

app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -158,9 +158,11 @@ public function execute(InputInterface $input, OutputInterface $output): int
             }
         }
 
-        if (! $onlySetCanonical) {
+        if (!$onlySetCanonical) {
             // 4. Trigger a SAPI3 reindex on the places removed from duplicate_places
-            $this->reindexPlaces($this->getDuplicatePlacesRemovedFromCluster(), $output, $dryRun);
+
+            $this->reindexPlaces($this->duplicatePlaceRemovedFromClusterRepository->getDuplicatePlacesRemovedFromCluster(), $output, $dryRun);
+            $this->duplicatePlaceRemovedFromClusterRepository->truncateTable();
```
LucWollants, line 165: Maybe it is better to truncate `duplicate_places_removed_from_cluster` before adding new ones? So we have a longer "history" of the previous run.

src/Place/Canonical/DBALDuplicatePlacesRemovedFromClusterRepository.php
```diff
@@ -0,0 +1,40 @@
...
+    public function addPlace(string $placeId): void
+    {
+        $this->connection->createQueryBuilder()
+            ->insert('duplicate_places_removed_from_cluster')
+            ->setValue('place_uuid', ':place_uuid')
+            ->setParameter(':place_uuid', $placeId)
+            ->execute();
+    }
+
+    public function getDuplicatePlacesRemovedFromCluster(): array
```
LucWollants, line 27: This can become `getAllPlaces`

src/Place/Canonical/ImportDuplicatePlacesProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+        DuplicatePlaceRepository $duplicatePlaceRepository,
+        DuplicatePlaceRemovedFromClusterRepository $duplicatePlacesRemovedFromClusterRepository
+    ) {
+        $this->duplicatePlaceRepository = $duplicatePlaceRepository;
+        $this->duplicatePlacesRemovedFromClusterRepository = $duplicatePlacesRemovedFromClusterRepository;
+    }
+
+    public function sync(): void
+    {
+        $this->deleteOldClusters();
```
LucWollants, line 22: I guess this is weird, but will get more method calls in the next pull requests?

tests/Place/Canonical/DBALDuplicatePlaceRepositoryTest.php
```diff
@@ -173,4 +173,52 @@ public function it_can_get_the_duplicates_of_a_place(): void
...
+            $this->duplicatePlaceRepository->getClustersToBeRemoved()
+        );
+    }
+
+
+    public function test_delete_cluster(): void
+    {
+        $this->duplicatePlaceRepository->deleteCluster('cluster_1');
+
+        $raw = $this->connection->fetchAssociative('select count(*) as total from duplicate_places where cluster_id = :cluster_id', ['cluster_id' => 'cluster_1']);
```
LucWollants, line 221: I would put in the extra effort to also use the query builder in tests, the query builder fluent syntax is so easy to read.

src/Place/Canonical/ImportDuplicatePlacesProcessor.php
grubolsch: Yes :-)

LucWollants (approved):
Thanks for all the extra improvements!

PR #1777, opened 2024-09-05
III-6266 Add script to convert cluster ids from numbers to sha1 hashes

grubolsch:
### Added
In order to run delta's of duplicate places we need consistent cluster ids.
In order to do this we agreed with the data team that all cluster ids will change to hashes of all place uuids combined in a sha1.
The idea is that when a place changes/added/removed it changes the hash.

For our current data we will also need to update these hashes, this is what this script is for.

This is a "run once and delete" script and with some good luck should not stay long in our code base.

---

Ticket: https://jira.uitdatabank.be/browse/III-6266


grubolsch:
I did not write a test for this because we only need to run this script once, but of course I tested it manually.

grubolsch:
Even in draft gezet omdat dit een groot verschil geeft met de nieuwe export van het datateam. Ik wil even nakijken of zij op een andere manier de hash bekijken voordat ik dit merge.

grubolsch:
Script zelf is oké, er zijn andere redenen gevonden voor het grote aantal nieuwe clusters.

app/Console/Command/FixClusterIdsDuplicatePlaces.php
```diff
@@ -0,0 +1,71 @@
...
+    ) {
+        parent::__construct();
+
+        $this->connection = $connection;
+    }
+
+    public function configure(): void
+    {
+        $this
+            ->setName('place:duplicate-places:fix-cluster-ids')
```
LucWollants, line 30: It is a convert of cluster ids to hash. Also the file name is wrong

app/Console/Command/ConvertClusterIdsToHashes.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command as BaseCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/*
+ * @todo This is a temporary script and can probably be removed somewhere in 2025
```
LucWollants, line 13: I would make a ticket for this; this can be removed the first time this has run.

LucWollants (approved):

PR #1778, opened 2024-09-06
III-6271 InvalidArgumentException on Empty Category

JonasVHG:
### Added

- `EmptyCategoryId`: Create new Exception for when `CategoryId` is empty.
- `event-with-empty-term-id.json`: new data file to be used in acceptance tests

### Changed

- `CategoryID`: throw new Exception `EmptyCategoryId`, when Category is Empty.
- `ImportTermRequestBodyParser`: catch new `EmptyCategoryId`
- `ImportTermRequestBodyParserTest`:  Add Unit Test with empty `Category`.
- `terms.feature`:  Add acceptance test with empty `Category`.

### Fixed

- Cleaner error when an empty Category is posted.

---

Ticket: https://jira.publiq.be/browse/III-6271


grubolsch (approved):

src/Http/Import/ImportTermRequestBodyParser.php
```diff
@@ -58,7 +59,11 @@ function ($term) use ($ignoreDomains) {
                 function ($term, int $index) {
                     if ($term instanceof stdClass && isset($term->id) && is_string($term->id)) {
                         $id = $term->id;
-                        $category = $this->categoryResolver->byId(new CategoryID($id));
+                        try {
+                            $category = $this->categoryResolver->byId(new CategoryID($id));
+                        } catch (EmptyCategoryId $exception) {
+                            $category = null;
```
LucWollants, line 65: Should we catch this? We once discussed that we would not silently accept wrong values or modify them to correct values. We should always inform the API integrator with a problem report.

JonasVHG: They will get a 4xx Error Response, instead of a 5xx Server Error.
`The term  does not exist or is not supported` Same as one uses  a termid that does not exist(e.g., a term with id 123)

tests/Http/Import/ImportTermRequestBodyParserTest.php
```diff
@@ -214,4 +214,37 @@ public function it_throws_on_unsupported_term(): void
...
+                    'domain' => 'eventtype',
+                ],
+            ],
+        ];
+
+        $request = (new Psr7RequestBuilder())
+            ->withParsedBody($terms)
+            ->build('POST');
+
+        $this->assertCallableThrowsApiProblem(
```
LucWollants, line 240: This is not what we see inside the Sentry issue

src/Http/Import/ImportTermRequestBodyParser.php
LucWollants: As discussed we now report a more specific error about an empty category ID

LucWollants (approved):

PR #1779, opened 2024-09-06
III-6246 remove duplicate DateTimeFactory

JonasVHG:
### Changed

- `DateTimeFactory`: Added new function `fromCdbFormat()` to handle dateTimes from legacy `CdbXml`
- `CalendarFactory`: Replace all occurrences of the `LegacyDateTimeFactory` with the new one.
- `CdbXMLItemBaseImporter`: Replace all occurrences of the `LegacyDateTimeFactory` with the new one.

- `DateTimeFactoryTest`: Add test for `fromCdbFormat()`

### Removed

- `src/Cdb/DateTimeFactory`: Legacy `DateTimeFactory` that is no longer in use.

---

Ticket: https://jira.publiq.be/browse/III-6246


src/Calendar/CalendarFactory.php
```diff
@@ -32,7 +33,12 @@ public function createFromCdbCalendar(\CultureFeed_Cdb_Data_Calendar $cdbCalenda
                 $startDateString = $timestamp->getDate() . 'T00:00:00';
             }
         }
-        $startDate = !empty($startDateString) ? DateTimeFactory::dateTimeFromDateString($startDateString) : null;
+        $startDate = !empty($startDateString) ?
+            DateTimeFactory::fromFormat(
```
LucWollants, line 37: Maybe we can hide some parameters and create a method called `DateTimeFactory::fromCdbFormat($startDateString)` and the format `'Y-m-d?H:i:s'` and timezone `new DateTimeZone('Europe/Brussels')` are reused inside that new method.

LucWollants: This format `'Y-m-d?H:i:s'` only seems to make sense inside the Cdb namespace and it doesn't represent any PHP standard format.

LucWollants (changes requested):
Good start, but maybe we can avoid some duplication.

tests/DateTimeFactoryTest.php
```diff
@@ -81,6 +81,21 @@ public function it_creates_a_date_time_object_from_a_valid_atom_string(string $g
...
 
+    /**
+     * @test
+     */
+    public function it_creates_a_date_time_object_from_a_cdb_format(): void
+    {
+        $dateTime = DateTimeFactory::fromCdbFormat('2014-06-30T11:48:26');
+        $this->assertEquals(
+            new DateTimeImmutable(
+                '2014-06-30T11:48:26',
```
LucWollants, line 92: Maybe we also need to check if it works without the `T` because I think that is why `?` is used.

JonasVHG: It's kinda hard, to still find all the formats that were used in `CdbXml` 🫣 I've only found ones with a `T` but all add some more testData.

LucWollants (approved):

PR #1795, opened 2024-09-17
III-6316 quota error

JonasVHG:
### Changed

- `KinepolisService`: Add bool to enable/disable searching for trailers.
- `YoutubeTrailerRepository`:  catch All Google Exceptions from YouTube API.

### Fixed

- By only searching for trailers on production, we have a lower change of hitting the Quota Limit.

### Related PR

- https://github.com/cultuurnet/appconfig/pull/801
---

Ticket: https://jira.publiq.be/browse/III-6316


simon-debruijn (approved):

app/Console/ConsoleServiceProvider.php
```diff
@@ -514,7 +514,8 @@ function () use ($container) {
                     LoggerFactory::create(
                         $container,
                         LoggerName::forService('fetching-movies', 'kinepolis')
-                    )
+                    ),
+                    $container->get('config')['kinepolis']['trailers']['enabled'],
```
LucWollants, line 518: The list of parameters is getting rather big. This makes me wonder if it should become part of `YoutubeTrailerRepository`

src/Kinepolis/KinepolisService.php
```diff
@@ -72,7 +74,8 @@ public function __construct(
         UuidGeneratorInterface $uuidGenerator,
         TrailerRepository $trailerRepository,
         ProductionRepository $productionRepository,
-        LoggerInterface $logger
+        LoggerInterface $logger,
+        bool $trailersEnabled = true
```
LucWollants, line 78: I'm not a fan of default parameters. It's better to put the default behavior at the place where the constructor is called.

src/Kinepolis/Trailer/YoutubeTrailerRepository.php
```diff
@@ -27,11 +28,15 @@ public function __construct(Google_Service_YouTube $youTubeClient, string $chann
...
-            'q' => urlencode($title),
-            'maxResults' => 1,
-        ]);
+        try {
+            $response = $this->youTubeClient->search->listSearch('id,snippet', [
+                'channelId' => $this->channelId,
+                'q' => urlencode($title),
+                'maxResults' => 1,
+            ]);
+        } catch (GoogleException $exception) {
```
LucWollants, line 37: Is every exception a quota exception? What about bad requests?

LucWollants (approved):

PR #1816, opened 2024-09-26
III-6310 Handle update with empty description

LucWollants:
### Added
- Added new exception `DescriptionShouldNotBeEmpty`

### Changed
- Handle an empty description and dispatch a `DeleteDescription` for this

---

### DIFF
--- a/src/Http/Offer/UpdateDescriptionRequestHandler.php
+++ b/src/Http/Offer/UpdateDescriptionRequestHandler.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace CultuurNet\UDB3\Http\Offer;
 
 use Broadway\CommandHandling\CommandBus;
-use CultuurNet\UDB3\Description;
+use CultuurNet\UDB3\Description as LegacyDescription;
 use CultuurNet\UDB3\Event\Commands\UpdateDescription as EventUpdateDescription;
 use CultuurNet\UDB3\Http\Request\Body\DenormalizingRequestBodyParser;
 use CultuurNet\UDB3\Http\Request\Body\JsonSchemaLocator;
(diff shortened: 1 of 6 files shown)


src/Http/Offer/UpdateDescriptionRequestHandler.php
```diff
@@ -47,26 +50,35 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
 
-        if ($offerType->sameAs(OfferType::event())) {
-            $updateDescription = new EventUpdateDescription(
-                $offerId,
-                $language,
-                $description
-            );
-        } else {
-            $updateDescription = new PlaceUpdateDescription(
+            if ($offerType->sameAs(OfferType::event())) {
```
grubolsch, line 59: Can we make either a seperate function or class helper that expects any type of UpdateDescription and just returns the correct type of Command?

grubolsch: Would be useful for example to have an OfferUpdateDescription class with a static method "buildCommand()" that returns the correct object? 
If you think that is overkill I am also happy with just a function in this handler.

grubolsch (changes requested):

src/Http/Offer/UpdateDescriptionRequestHandler.php
LucWollants: The approach described here should be used https://jira.publiq.be/browse/III-3834. This is not inside the scope of this refactor of value objects and that is why I didn't change it.

grubolsch (approved):

src/Http/Offer/UpdateDescriptionRequestHandler.php
grubolsch: I will approve then

PR #1818, opened 2024-09-26
III-6310 Refactor `AbstractDescriptionTranslated` to use new `Description`

LucWollants:
### Changed
- Refactored `AbstractDescriptionTranslated` to use new `Description`

---


src/Description.php
```diff
@@ -27,4 +27,9 @@ public static function fromUdb3ModelDescription(Udb3ModelDescription $udb3ModelD
     {
         return new self($udb3ModelDescription->toString());
     }
+
+    public function toUdb3ModelDescription(): Udb3ModelDescription
```
grubolsch, line 31: Can we use return self?

```suggestion
    public function toUdb3ModelDescription(): self
    {
        return new self($this->toString());
    }
```

grubolsch (approved):
1 suggestion but I already approved

src/Description.php
LucWollants: This class will be deleted https://github.com/cultuurnet/udb3-backend/pull/1821

PR #1822, opened 2024-09-30
III-6339: bugfix PDF export

JonasVHG:
### Changed

- `TransformingIteratorIterator`: remove returnType from `current()`
- `phpstan.neon` : ignore TransformingIteratorIterator for the moment, because we can not get phpstan to check it for both PHP 7.4 & PHP 8.1

### Fixed

- PDF exports should work again.

### Follow-up ticket to be done in the future

- https://jira.publiq.be/browse/III-6342

---

Ticket: https://jira.publiq.be/browse/III-6339

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -14,7 +14,7 @@ parameters:
 	scanFiles:
 		- vendor/gridonic/princexml-php/lib/prince.php
 	excludePaths:
-	    - src/EventExport/Format/HTML/TransformingIteratorIterator.php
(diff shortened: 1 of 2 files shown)


simon-debruijn (approved):

LucWollants:
@JonasVHG I wonder if we could write a unit test to cover this? We know that the acceptance tests are more tricky to fix because of the asynchronous behaviour. But we could write a test for `WebArchiveFileWriter::writeHtml` which gets input that makes the `current` method crash on the return type.

PR #1827, opened 2024-10-03
III-6347 BUGFIX Keycloak id token was wrongly classified as access token

grubolsch:
### Fixed
Keycloak tokens always have an AZP field, auth0 does not.
This tripped up our code to identify if it is a provider token (a token with an api key).
We added an extra check to typ, a field that only exists in Keycloak.

---

Ticket: https://jira.uitdatabank.be/browse/III-6347

### DIFF
--- a/src/Http/Auth/Jwt/JsonWebToken.php
+++ b/src/Http/Auth/Jwt/JsonWebToken.php
@@ -63,7 +63,7 @@ final class JsonWebToken
             return self::UIT_ID_V2_JWT_PROVIDER_TOKEN;
         }
 
-        // This is a check for Keycloak tokens, auth0 does not have this field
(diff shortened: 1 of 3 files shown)


src/Http/Auth/Jwt/JsonWebToken.php
```diff
@@ -63,6 +63,11 @@ public function getType(): string
             return self::UIT_ID_V2_JWT_PROVIDER_TOKEN;
         }
 
+        // This is a check for Keycloak tokens, auth0 does not have this field
```
JonasVHG, line 66:
```suggestion
        // Keycloak ID Tokens have a claim `typ` with a value of ID
```
I would give more info in the comment, because  Keycloak `accessTokens` also have a claim `typ` but with a different value.

PR #1833, opened 2024-10-10
III-6342: Install prince in Dockerfile

JonasVHG:
### Changed

- `Makefile`: Ignore `@pdf`-tagged tests on `Jenkins`
- `Dockerfile`: install `Prince` in the container, if running on `arm64`
- `export.feature`: add `@pdf`-tag to related tests.

### Related PR

- https://github.com/cultuurnet/appconfig/pull/845

---

https://jira.publiq.be/browse/III-6342


LucWollants (approved):

docker/php/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM php:7.4-apache
 
-RUN apt-get update && apt-get install -y --no-install-recommends git zip unzip libicu-dev libtidy-dev zlib1g-dev libpng-dev libzip-dev supervisor
+RUN apt-get update && apt-get install -y --no-install-recommends git zip unzip libicu-dev libtidy-dev zlib1g-dev libpng-dev libzip-dev supervisor libfontconfig1
```
LucWollants, line 3: Do we need this `libfontconfig1`? It is not needed on macOS and was a test on Jenkins.

Makefile
```diff
@@ -48,7 +48,7 @@ feature-tag:
 	docker-compose exec php composer feature -- --tags $(tag)
 
 feature-ci:
-	docker-compose exec php composer feature -- --tags "~@init&&~@external" -f pretty -o std -f junit -o output/junit
+	docker-compose exec php composer feature -- --tags "~@init&&~@external&&~@pdf" -f pretty -o std -f junit -o output/junit
```
LucWollants, line 51: Looks good. But I wonder, why the mailing list test was excluded?

LucWollants (approved):

Makefile
JonasVHG: Probably because we don't want to get into problems with `Mailjet` by subscribing non-existing mailAddresses.
Still since we're getting a new testMail-account for PPF, It might be a good idea to also it here.

LucWollants: I see, this is also done locally. Thanks for the explanation.

PR #1837, opened 2024-10-14
III-6275 Make Labels findable by using Levenshtein distance

JonasVHG:
### Changed

- `ReadRepositoryInterface`, `DBALReadRepository`, `AppConfigReadRepositoryDecorator` & `GodUserReadRepositoryDecorator`: Add `searchByLevenshtein()` functionality.
- `SearchLabelsRequestHandler`: Use `searchByLevenshtein()` instead of `search()`
- `tests/SearchLabelsRequestHandlerTest`: Update test with new functionality
- `tests/DBALReadRepositoryTest`: Add unit tests for alphabetically & Levenshtein search.
- `features/label/get.feature`: Update `acceptance test`

### Fixed

- Searching for labels `mantelzorgers` will return a more exact list instead of alphabetically sorted labels.

---

Ticket: https://jira.publiq.be/browse/III-6275


LucWollants:
As discussed through Slack this solution will not return the exact match if this was not returned as one of the first n results of the default query.

JonasVHG:
This is not the correct approach.
A better solution would be something like:
`SELECT uuid_col, name, visible, private, excluded, rank
FROM (
    SELECT uuid_col, name, visible, private, excluded, 
           (CASE 
               WHEN name = 'abc' THEN 1
               WHEN name LIKE 'abc%' THEN 2
               WHEN name LIKE '%abc%' THEN 3
               ELSE 4
           END)
    FROM labels_json
    WHERE name LIKE '%abc%'
) AS subquery
ORDER BY rank ASC;` which should be done in a different PR.

PR #1848, opened 2024-10-21
III-6346 Use organizer name for role name

LucWollants:
### Changed
- Used organizer name for role name. Currently, this only works for organizers, but it can be easily extended for places and events.

---

Ticket: https://jira.uitdatabank.be/browse/III-6346

### DIFF
--- a/src/Ownership/Readmodels/OwnershipPermissionProjector.php
+++ b/src/Ownership/Readmodels/OwnershipPermissionProjector.php
@@ -151,7 +151,7 @@ final class OwnershipPermissionProjector implements EventListener
         $this->commandBus->dispatch(
             new CreateRole(
                 $roleId,
-                'Beheerders ' . $this->itemNameResolver->resolve($ownershipItem->getItemId())
+                'Beheerders organisatie ' . $this->itemNameResolver->resolve($ownershipItem->getItemId())
             )
         );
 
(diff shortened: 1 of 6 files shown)


src/Ownership/Readmodels/OwnershipPermissionProjector.php
```diff
@@ -147,7 +151,7 @@ private function createRole(OwnershipItem $ownershipItem): UUID
         $this->commandBus->dispatch(
             new CreateRole(
                 $roleId,
-                $this->createRoleName($ownershipItem->getId())
+                'Beheerders ' . $this->itemNameResolver->resolve($ownershipItem->getItemId())
```
grubolsch, line 154: Is this a public visible name?
Doesn't UDB have multiple languages? This will never be translated?
Shouldn't we add this label to the frontend, instead of hardcoding it in one language?

grubolsch (changes requested):

src/Ownership/Readmodels/OwnershipPermissionProjector.php
LucWollants: The role name is always in one language, there are no translations for role names.

LucWollants: That being said, the role name itself, is never exposed to a normal user. This role, in fact, all roles are only visible in an admin view for users who have this very limited permission.

grubolsch (approved):

PR #1859, opened 2024-11-05
III-6310 Remove `AbstractCollection` and use `Collection`

LucWollants:
### Changed
- Refactored `SpecificationCollection` to use `Collection`
- Refactored `OfferIdentifierCollection` to use `Collection`
- Refactored `ImageCollection` to use `Collection`

### Removed
- Removed `AbstractCollection`

---


src/Model/ValueObject/Collection/Collection.php
```diff
@@ -41,6 +41,21 @@ public function with($value) // @phpstan-ignore-line III-5812 Can be fixed once
         return new static(...$values);
     }
 
+    /**
+     * @return static
+     */
+    public function without($value) // @phpstan-ignore-line
```
simon-debruijn, line 47:
```suggestion
    public function without(Image $value) // @phpstan-ignore-line
```
Can this be Image type?

simon-debruijn (approved):

src/Model/ValueObject/Collection/Collection.php
LucWollants: No, because this `Collection` is generic and unfortunately there are no generics in PHP.

simon-debruijn: never mind, I thought I was looking in the specific ImageColection.

JonasVHG (approved):

PR #1861, opened 2024-11-05
III-6389 Prevent UiTpas locations to be deleted by the delete place cli script

grubolsch:
### Added
- Added unit tests for DeletePlace command

### Changed
- Prevent UiTpas locations to be deleted by the delete place cli script
- Check if canonical place actually exists before moving events to it

---

Ticket: https://jira.uitdatabank.be/browse/III-6389


LucWollants (commented):
I wonder if we should prevent it inside the command handler instead of the CLI command?

JonasVHG:
> I wonder if we should prevent it inside the command handler instead of the CLI command?

That would make impossible to manually delete a place, which might not be desired.
On the other hand, an admin can also unlink an UiTPAS label from a place, if he/she wants to delete the place manually. Which might be a good added safety mechanism.

LucWollants:
> > I wonder if we should prevent it inside the command handler instead of the CLI command?
> 
> That would make impossible to manually delete a place, which might not be desired. On the other hand, an admin can also unlink an UiTPAS label from a place, if he/she wants to delete the place manually. Which might be a good added safety mechanism.

I have the feeling that preventing deleting UiTPAS places is the most important reason for this change, so we should prevent this inside the command handler

grubolsch:
I made it fully blocking.

LucWollants (changes requested):
As discussed through Slack, together with @JonasVHG, deleting a place with an UiTPAS label should never be possible. This hard restriction makes it better to prevent this in the `DeleteOfferHandler` and write some extra unit and acceptance tests for this on the command and request handlers, no test is needed on the CLI command

grubolsch:
I revert the check in the console command itself, but put a check in DeleteOfferHandler.
I choose a more generic approach with a new Validator interface, with the idea that in the future we might introduce more validators, looping over them to check validity.
I did not want to introduce this loop right now, because it is not needed.

I wanted to avoid code in the Handler itself.

LucWollants (changes requested):
As commented, it seems better to reuse the `PermissionVoter` interface

grubolsch:
Processed

src/Offer/Validator/OfferCommandValidator.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Offer\Validator;
+
+use CultuurNet\UDB3\Http\ApiProblem\ApiProblem;
+
+interface OfferCommandValidator
```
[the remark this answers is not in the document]
grubolsch: I processed this feedback

src/Http/Offer/DeleteRequestHandler.php
```diff
@@ -26,7 +27,11 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         $routeParameters = new RouteParameters($request);
         $offerId = $routeParameters->getOfferId();
 
-        $this->commandBus->dispatch(new DeleteOffer($offerId));
+        try {
+            $this->commandBus->dispatch(new DeleteOffer($offerId));
+        } catch (CannotDeleteUiTPASPlace $e) {
+            throw $e->toApiProblem();
```
LucWollants, line 33: There is no need to call this, I think the default ErrorLogger will handle this

src/Http/ApiProblem/ApiProblem.php
```diff
@@ -562,4 +562,14 @@ public static function failedToSubscribeToNewsletter(string $detail): self
...
         );
     }
+
+    public static function cannotDeleteUitpasPlace(): self
+    {
+        return self::create(
+            'https://api.publiq.be/probs/uitdatabank/cannot-delete-uitpas-place',
+            'Cannot delete UiTPAS place',
+            StatusCodeInterface::STATUS_FORBIDDEN,
+            'Place is an UiTPAS balie. UiTPAS locations cannot be deleted.'
```
LucWollants, line 572: There is a naming mix of place and location

src/Offer/Commands/DeleteOffer.php
```diff
@@ -4,6 +4,12 @@
 
 namespace CultuurNet\UDB3\Offer\Commands;
 
+use CultuurNet\UDB3\Role\ValueObjects\Permission;
+
 final class DeleteOffer extends AbstractCommand
 {
+    public function getPermission(): Permission
+    {
+        return Permission::aanbodVerwijderen();
```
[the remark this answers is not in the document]
LucWollants: I wonder if the existing acceptance tests still work when this is changed. Maybe there is no acceptance test for deleting a place? If there is no acceptance test to delete a place it should be added.

tests/Offer/CommandHandlers/DeleteOfferHandlerTest.php
```diff
@@ -124,6 +167,30 @@ public function it_does_not_delete_a_deleted_offer_again(): void
             ->then([]);
     }
 
+    /**
+     * @test
+     */
+    public function it_prevents_deletion_when_validator_fails(): void
```
LucWollants, line 173:
```suggestion
    public function it_prevents_deletion_when_voter_fails(): void
```

tests/Security/Permission/DeleteUiTPASPlaceVoterTest.php
```diff
@@ -0,0 +1,79 @@
...
+    private string $userId;
+
+    protected function setUp(): void
+    {
+        $this->documentRepository = $this->createMock(DocumentRepository::class);
+        $this->uitpasLabels = ['UiTPAS', 'Paspartoe'];
+        $this->userId = Uuid::uuid4()->toString();
+    }
+
+    public function test_is_valid_success(): void
```
LucWollants, line 29: The names of the test methods are rather cryptic, it should me something like "deleting a place without UiTPAS labels is allowed"

LucWollants: We still mark all the tests with a header doc instead of the `test` prefix, maybe best to stay consistent?

tests/Security/Permission/DeleteUiTPASPlaceVoterTest.php
```diff
@@ -0,0 +1,79 @@
...
+            ], JSON_THROW_ON_ERROR)));
+
+        $validator = new DeleteUiTPASPlaceVoter($this->documentRepository, $this->uitpasLabels);
+
+        $result = $validator->isAllowed(Permission::aanbodVerwijderen(), $offerId, $this->userId);
+
+        $this->assertTrue($result);
+    }
+
+    public function test_is_valid_failure(): void
```
LucWollants, line 46: "deleting a place with an UiTPAS label is not allowed"

features/place/delete.feature
```diff
@@ -5,9 +5,24 @@ Feature: Test deleting places
...
 
    Scenario: Delete place
-    When I delete the place at "%{placeUrl}"
+    When I create a minimal place and save the "url" as "placeUrl"
+    And I delete the place at "%{placeUrl}"
     And I get the place at "%{placeUrl}"
     Then the JSON response at "workflowStatus" should be "DELETED"
+
+   Scenario: Prevent deletion of UiTPAS place
+    Given I set the name at value "UiTPAS"
```
LucWollants, line 16: I set the value of name to "UiTPAS

features/Steps/UtilitySteps.php
```diff
@@ -24,6 +24,14 @@ public function iCreateARandomNameOfCharacters(int $nrOfCharacters): void
         $this->variableState->setRandomVariable('name', $nrOfCharacters);
     }
 
+    /**
+     * @Given I set the name at value :value
```
LucWollants, line 28: `I set the value of name to :value`

features/Steps/PlaceSteps.php
```diff
@@ -152,6 +152,18 @@ public function iDeleteThePlaceAt(string $url): void
         $this->theResponseStatusShouldBe(204);
     }
 
+    /**
+     * @When I delete the place at :url then the response status should be :responseStatus
+     */
+    public function iDeleteThePlaceAtThenTheResponseStatusShouldBe(string $url, int $responseStatus): void
```
LucWollants, line 158: The approach for exceptions or checking bad requests is just to call `And I send a POST request to `, see https://github.com/cultuurnet/udb3-backend/blob/master/features/auth/user-access-token.feature#L47

Otherwise, a method needs to be called for every specific endpoint.

tests/Console/Command/DeletePlaceTest.php
```diff
@@ -0,0 +1,144 @@
...
+        $this->placeDocumentRepository
+            ->method('fetch')
+            ->willReturn(new JsonDocument(Uuid::uuid4()->toString(), json_encode([])));
+
+        $this->eventRelationsRepository
+            ->method('getEventsLocatedAtPlace')
+            ->willReturn([$eventId1, $eventId2]);
+
+        $dispatchedCommands = [];
+        $this->commandBus
```
LucWollants, line 122: There is a special class to handle this, the class is called `TraceableCommandBus`

LucWollants (commented):
Nice work! Only a few naming improvements and convenience classes to use.

src/Http/Offer/DeleteRequestHandler.php
grubolsch: Ah, oké, did not know this. Changed it, and tested it again - it works.

src/Http/ApiProblem/ApiProblem.php
grubolsch: Fixed it. Also replaced the word "balie" which is dutch.

src/Offer/Commands/DeleteOffer.php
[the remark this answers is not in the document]
grubolsch: I have reverted it for now, to get this PR passed, but I think we should investigate this future.
This seems to be a bug.

features/Steps/UtilitySteps.php
```diff
@@ -24,6 +24,14 @@ public function iCreateARandomNameOfCharacters(int $nrOfCharacters): void
         $this->variableState->setRandomVariable('name', $nrOfCharacters);
     }
 
+    /**
+     * @Given I set the value of name to value :value
```
LucWollants, line 28: No need for the last "value" in the sentence, it can just be "I set the value of name to"

src/Offer/Commands/DeleteOffer.php
```diff
@@ -4,6 +4,12 @@
 
 namespace CultuurNet\UDB3\Offer\Commands;
 
+use CultuurNet\UDB3\Role\ValueObjects\Permission;
+
 final class DeleteOffer extends AbstractCommand
 {
+    public function getPermission(): Permission
```
LucWollants, line 11: No need to add this, it's part of `AbstractCommand`

tests/Security/Permission/DeleteUiTPASPlaceVoterTest.php
LucWollants: @grubolsch what do you think about this remark

> We still mark all the tests with a header doc instead of the test prefix, maybe best to stay consistent?

src/Offer/Commands/DeleteOffer.php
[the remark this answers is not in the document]
LucWollants: I think adding the extra acceptance test to delete a place as a normal user can be useful. So the scenarios to add are:
- An admin creates a place, a normal users tries to delete this place, this should fail
- A normal user creates a place, this users tries to delete this place,  this should work

features/Steps/UtilitySteps.php
LucWollants: The sentence is still confusing with the extra "value"

tests/Security/Permission/DeleteUiTPASPlaceVoterTest.php
grubolsch: I know I am not doing it consistent, I have not being this for a while.
I really prefer not to use docblocks when i can, this seems to be a more elegant solution to me.

So I prefer to keep it this way, but if really pressed I will change it.

src/Offer/Commands/DeleteOffer.php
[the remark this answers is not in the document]
grubolsch: Done

features/place/delete.feature
```diff
@@ -27,3 +27,17 @@ Feature: Test deleting places
         "detail": "Place is an UiTPAS counter. UiTPAS places cannot be deleted."
     }
     """
+
+    Scenario: An admin creates a place, a normal users tries to delete this place, this should fail
+        Given I am authorized as JWT provider v1 user "centraal_beheerder"
```
LucWollants, line 32: This is already part of the background

PR #1875, opened 2024-11-08
III-6310 Remove legacy `BookingAvailability` from `Timestamp`

LucWollants:
### Removed
- Refactored `Timestamp` to use new `BookingAvailability` model

---


src/Model/Serializer/ValueObject/Calendar/BookingAvailabilityNormalizer.php
```diff
@@ -0,0 +1,26 @@
...
+
+use CultuurNet\UDB3\Model\ValueObject\Calendar\BookingAvailability;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+final class BookingAvailabilityNormalizer implements NormalizerInterface
+{
+    /**
+     * @param BookingAvailability $bookingAvailability
+     */
+    public function normalize($bookingAvailability, $format = null, array $context = []): array
```
grubolsch, line 15: Optional nitpick.
Instead of a weak param doc, you could have an instanceof check in the code to enforce the type.
I prefer to avoid docblocks at all costs.

I am also thinking for future Rector runs to upgrade the code to PHP8, it might trip over this piece of code.

grubolsch (approved):

PR #1882, opened 2024-11-12
III-6310 Remove legacy `OpeningTime`

LucWollants:
### Removed
- Removed legacy `OpeningTime`

---


src/Calendar/CalendarFactory.php
```diff
@@ -209,8 +210,8 @@ private function createOpeningHoursFromWeekScheme(
                     );
 
                     $openingHour = new OpeningHour(
-                        OpeningTime::fromNativeDateTime($opens),
-                        $closes ? OpeningTime::fromNativeDateTime($closes) : OpeningTime::fromNativeDateTime($opens),
+                        Time::fromDateTime($opens),
+                        $closes ? Time::fromDateTime($closes) : Time::fromDateTime($opens),
```
grubolsch, line 214:
```suggestion
                        Time::fromDateTime($closes ?? $opens),
```

src/Model/ValueObject/Calendar/OpeningHours/Time.php
```diff
@@ -4,6 +4,10 @@
 
 namespace CultuurNet\UDB3\Model\ValueObject\Calendar\OpeningHours;
 
+use CultuurNet\UDB3\DateTimeFactory;
+use DateTime;
+use DateTimeInterface;
+
 class Time
```
grubolsch, line 11: I wonder if it would be a good idea to not put this in the OpeningHours namespace, but in the general ValueObject namespace, because this is such a broad and useful DTO.
I can imagine in the future using this class for other uses than the Calender.

grubolsch (approved):
Some opinionated suggestions but will already approve.

src/Model/ValueObject/Calendar/OpeningHours/Time.php
LucWollants: That is possible, but for now, the only supported format is `'H:i'` which is rather very specific in the context of opening hours and not in the sense of a time of the day.

src/Calendar/CalendarFactory.php
LucWollants: Ah, but this will not work, according to PHPStan "Variable $closes on left side of ?? always exists and is not nullable."

PR #1887, opened 2024-11-15
III-6396 BUGFIX description should not be empty

grubolsch:
### Fixed

- BUGFIX description should not be empty
---

Ticket: https://jira.uitdatabank.be/browse/III-6396


src/BackwardsCompatiblePayloadSerializerFactory.php
```diff
@@ -420,6 +418,9 @@ function (array $serializedObject) use ($contactPointUpdatedEvent) {
             );
         }
 
+        self::fillDescriptions($payloadManipulatingSerializer);
```
LucWollants, line 421: If you follow the existing approach then the code will look like

```
$updateDescriptionEvents = [
  EventDescriptionUpdated::class,
  PlaceDescriptionUpdated::class,
];
foreach ($updateDescriptionEvents as $updateDescriptionEvents) {
   ...
   self::fillDescriptions(serializedObject)
   self::replacePlaceIdWithItemId($serializedObject)
   ...
}
```

The reasoning behind this was that:
- This makes it more clear which events are manipulated
- The update methods are very granular and are building blocks

LucWollants (approved):
I already approved because the bug seems to be fixed. Up to you to make the code consistent or not.

PR #1888, opened 2024-11-15
III-6374 - get list suggested organizers

simon-debruijn:
### Added
- Added endpoint `Get /ownership/suggestions` to get a list of organizers. This list is based on the events and places the current user is the owner of.

### Changed
- Updated the ownership search repository to handle multiple states in a search query

---

Ticket: https://jira.uitdatabank.be/browse/III-6374


src/Http/Ownership/SuggestOwnershipsRequestHandler.php
```diff
@@ -0,0 +1,148 @@
...
+
+    private static function activeOwnerships(): callable
+    {
+        return function (OwnershipItem $item): bool {
+            $state = new OwnershipState($item->getState());
+            return OwnershipState::requested()->sameAs($state) || OwnershipState::approved()->sameAs($state);
+        };
+    }
+
+    private function guardAccess(): void
```
LucWollants, line 133: I think this should use `organizer_permission_voter` like `OwnershipStatusGuard` does because the organizer ID is an important parameter to decide if the current user has access to the current organisation.

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
```diff
@@ -0,0 +1,148 @@
...
+
+            $organizer = $this->organizerRepository->fetch($organizerId)->getAssocBody();
+
+            $idToOrganizerMap[$organizerId] = $organizer;
+        }
+
+        return new JsonLdResponse(array_values($idToOrganizerMap));
+    }
+
+    private static function activeOwnerships(): callable
```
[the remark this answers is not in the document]
simon-debruijn: Yes that would be very helpful if we could search for multiple states instead of filtering after request

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
```diff
@@ -0,0 +1,148 @@
...
+         * A map to deduplicate returned organizers
+         * @var array<string, array> $idToOrganizerMap
+         */
+        $idToOrganizerMap = [];
+
+        /**
+         * @var ItemIdentifier $result
+         */
+        foreach ($this->resultsGenerator->search($query) as $result) {
+            $offerType = OfferType::fromCaseInsensitiveValue($result->getItemType()->toString());
```
LucWollants, line 98: I wonder if we even need to know the type. It is possible to write the query so that only events are returned and then you only need to access the JSON projections of events.

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
```diff
@@ -0,0 +1,148 @@
...
+            if (!$ownershipsForOrganizer->isEmpty()) {
+                continue;
+            }
+
+            // don't suggest duplicate items
+            if (isset($idToOrganizerMap[$organizerId])) {
+                continue;
+            }
+
+            $organizer = $this->organizerRepository->fetch($organizerId)->getAssocBody();
```
LucWollants, line 117: Why do we get the full organizer JSON LD body? Is it not enough to just get the IDs?

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
```diff
@@ -0,0 +1,148 @@
...
+         */
+        foreach ($this->resultsGenerator->search($query) as $result) {
+            $offerType = OfferType::fromCaseInsensitiveValue($result->getItemType()->toString());
+            $id = $result->getId();
+
+            $offer = $this->offerRepository->fetch($offerType, $id)->getAssocBody();
+
+            $organizerId = $this->organizerIDParser->fromUrl(new Url($offer['organizer']['@id']))->toString();
+
+            $ownershipsForOrganizer = $activeOwnerships->filter(fn (OwnershipItem $item) => $item->getItemId() === $organizerId);
```
LucWollants, line 105: Is it not more efficient to just get all the organizer IDs first and then filter out the once with already an ownership? That would limit the number of filter calls.

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
LucWollants: I'm wrong, I don't even think there is a check needed.

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
simon-debruijn: Do we want to only return organizers linked to events? The ticket mentioned fetching from offers endpoint.
The `OfferJsonDocumentReadRepository` requires the `OfferType` as the first argument

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
simon-debruijn: The frontend will have to display the title, score, status, image ...

Otherwise the frontend will have to do a fetch for each suggestion, which is also possible

Screenshot 2024-11-18 at 09 35 20

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
LucWollants: You are right, we need to take into account events and places.

src/Http/Ownership/SuggestOwnershipsRequestHandler.php
LucWollants: I always have a strong opinion that an API should just provide the building blocks and should not be designed as a convenience for a representation layer. In this case, the UI is the representation layer and in this specific case, it needs information about the organizers. So it's up to the UI to get that extra information.

simon-debruijn: Ok
[Return organizers without assoc body](https://github.com/cultuurnet/udb3-backend/pull/1888/commits/9fda91b11ec413e6c9b828f0469c24c8838ffc93)

LucWollants (approved):

features/ownership/suggestion.feature
```diff
@@ -0,0 +1,50 @@
...
+  Scenario: Suggest the ownership on an organizer of a created place
+    And I am authorized as JWT provider v1 user "centraal_beheerder"
+    And I create a minimal organizer and save the "id" as "organizerId"
+    And I keep the value of the JSON response at "url" as "organizerUrl"
+    And I am authorized as JWT provider v2 user "invoerder"
+    And I create a place from "places/place-with-organizer.json" and save the "id" as "placeId"
+    And I wait for the place with url "places/%{placeId}" to be indexed
+    When I send a GET request to "ownerships/suggestions/?itemType=organizer"
+    Then the response status should be 200
+    And the JSON response should include:
```
simon-debruijn, line 16: Wouldn't it be better to be a bit more strict in how the response should look?

features/ownership/suggestion.feature
```diff
@@ -0,0 +1,50 @@
...
+    And I create a minimal organizer and save the "id" as "organizerId"
+    And I keep the value of the JSON response at "url" as "organizerUrl"
+    And I create a minimal place and save the "id" as "placeId"
+    And I am authorized as JWT provider v2 user "invoerder"
+    And I create an event from "events/event-with-organizer.json" and save the "id" as "eventId"
+    And I wait for the event with url "events/%{eventId}" to be indexed
+    When I send a GET request to "ownerships/suggestions/?itemType=organizer"
+    Then the response status should be 200
+    And the JSON response should include:
+    """
```
simon-debruijn, line 32: Same remark

features/ownership/suggestion.feature
LucWollants: That was something I first tried. But when running the acceptance tests locally for this user, the list of suggestions only grows because of previous runs. So we can only check if the new place or event is part of the query.

PR #1892, opened 2024-11-19
III-6395 Ignore rejected  & deleted ownership requests

LucWollants:
### Added
- Added editor setting for feature files to 2 spaces

### Changed
-  Ignored rejected  & deleted ownership requests when requesting a new ownership

### Removed
- Removed unique constraint on `ownership_search`

---

Ticket: https://jira.uitdatabank.be/browse/III-6395


app/Migrations/Version20240306114323.php
```diff
@@ -22,7 +22,6 @@ public function up(Schema $schema): void
         $table->setPrimaryKey(['id']);
         $table->addIndex(['item_id']);
         $table->addIndex(['owner_id']);
-        $table->addUniqueIndex(['item_id', 'owner_id']);
```
simon-debruijn, line 25: This will require some manual work then to remove existing unique indexes?

Otherwise a new migration removing if exists would be better no?

src/Http/Ownership/RequestOwnershipRequestHandler.php
```diff
@@ -93,12 +83,31 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+                new SearchParameter('ownerId', $requestOwnership->getOwnerId()->toString()),
+            ])
+        );
+
+        /** @var OwnershipItem $ownershipItem */
+        foreach ($existingOwnershipItems as $ownershipItem) {
+            if ($ownershipItem->getState() === OwnershipState::requested()->toString() ||
+                $ownershipItem->getState() === OwnershipState::approved()->toString()) {
+                throw ApiProblem::ownerShipAlreadyExists(
+                    'An ownership request for this item and owner already exists with id ' . $ownershipItem->getId()
```
simon-debruijn, line 106: I see no unit test for this case, there is a feature test tho

simon-debruijn (approved):

JonasVHG (approved):

app/Migrations/Version20240306114323.php
LucWollants: True!

But the issue is that the unique index was created without a name, this name will be random so it can't be deleted with a migration. I already discussed this with @willaerk on Slack and already did the manual removal of these indexes. The tables on TEST and PROD were still empty.

Also, the acceptance tests start from scratch and they still work.

src/Http/Ownership/RequestOwnershipRequestHandler.php
LucWollants: If you check `RequestOwnershipRequestHandlerTest` there are in total 4 tests that take into account the state. 

There are 2 tests that work because the state is either rejected or deleted:
- it_allows_requesting_same_ownership_when_rejected
- it_allows_requesting_same_ownership_when_deleted

There are 2 tests that test the requested and approved states:
- it_prevents_requesting_same_ownership_when_approved
- it_prevents_requesting_same_ownership (maybe less explicit about the state)

PR #1902, opened 2024-11-25
III-6310 Use `TelephoneNumber` in legacy `BookingInfo`

LucWollants:
### Changed
- Used `TelephoneNumber` in legacy `BookingInfo`

### Note
- This approach will make it easier to switch from the legacy `BookingInfo` to the new `BookingInfo` because their interface will be the same.

---

### DIFF
--- a/src/BookingInfo.php
+++ b/src/BookingInfo.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace CultuurNet\UDB3;
 
 use CultuurNet\UDB3\Model\ValueObject\Contact\BookingInfo as Udb3ModelBookingInfo;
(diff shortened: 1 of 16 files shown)


grubolsch:
In principle it looks good, but I do hope we are not getting a lot of problems with the eventStore again for changed validation in the phone number.

grubolsch (approved):

LucWollants:
> In principle it looks good, but I do hope we are not getting a lot of problems with the eventStore again for changed validation in the phone number.

Same remark as here https://github.com/cultuurnet/udb3-backend/pull/1904#issuecomment-2505738485

PR #1904, opened 2024-11-25
III-6310 Use `EmailAddress` in legacy `BookingInfo`

LucWollants:
### Changed
- Used `EmailAddress` in legacy `BookingInfo`

---


JonasVHG (approved):

grubolsch:
Before we merge this, I wonder if once again this is going to introduce issues with a stricter validation of email addreses in `EmailAddress`? Can we do something to mitigate this risk?

LucWollants:
> Before we merge this, I wonder if once again this is going to introduce issues with a stricter validation of email addreses in `EmailAddress`? Can we do something to mitigate this risk?

Valid remark. But the command layer is already strict because the `UpdateBookingInfoRequestHandler` was already using the schema validation. The only thing that can happen are older events with again some events that don't have a valid email. But when this happens we can look into the `BackwardsCompatiblePayloadSerializerFactory`

PR #1906, opened 2024-11-25
III-6310 use booking availability inside legacy booking info

LucWollants:
### Changed
- Used `BookingAvailability` inside legacy `BookingInfo`

---


src/BookingInfo.php
```diff
@@ -151,35 +131,28 @@ public function sameAs(BookingInfo $otherBookingInfo): bool
...
-        $website = null;
-        $phone = null;
-        $email = null;
-        $availabilityStarts = null;
-        $availabilityEnds = null;
-
-        if ($udb3ModelWebsite = $udb3ModelBookingInfo->getWebsite()) {
-            $website = $udb3ModelWebsite;
-        }
+        return new self(
```
grubolsch, line 134: this is so much nicer!

LucWollants: Finally this class gets removed because it's converting the new model to the legacy model which now has the same type of properties 😏

LucWollants: See https://github.com/cultuurnet/udb3-backend/pull/1907

BTW, there is a story in these pull requests, from bottom to top  😊. Small steps of making the old legacy model look and behave like the new model.

grubolsch (approved):

PR #1908, opened 2024-11-27
III-6310 Refactor `EventCreated` to use `EventType`

LucWollants:
### Changed
- Refactored `EventCreated` to use `EventType`

---


src/Model/Serializer/ValueObject/Taxonomy/Category/CategoryDenormalizer.php
```diff
@@ -0,0 +1,53 @@
...
+        return $type === Category::class;
+    }
+
+    private function createCategoryDomain(array $data): ?CategoryDomain
+    {
+        if ($this->defaultDomain) {
+            return $this->defaultDomain;
+        }
+
+        return isset($data['domain']) ? new CategoryDomain($data['domain']) : null;
```
grubolsch, line 51: I find it strange a method called "createCategoryDomain" can return NULL.
I think it should throw an exception when it cannot create a category domain.

I digged a little bit deeper and I see that this object normally goes into the constructor of `CultuurNet\UDB3\Model\ValueObject\Taxonomy\Category`. It does expect a NULL there.
But if you look further, none of the `getDomain()` functions actually check for a NULL, leading to possible NULL errors.

One random example:
```
    private function setTerms(Resource $resource, Categories $terms): void
    {
        /** @var Category $term */
        foreach ($terms as $term) {
            if ($term->getDomain()->sameAs(new CategoryDomain('eventtype'))) {
                $terms = $this->termsIriGenerator->iri($term->getId()->toString());
                $resource->set(self::PROPERTY_ACTIVITEIT_TYPE, new Resource($terms));
            }

            if ($term->getDomain()->sameAs(new CategoryDomain('theme'))) {
                $terms = $this->termsIriGenerator->iri($term->getId()->toString());
                $resource->set(self::PROPERTY_ACTIVITEIT_THEMA, new Resource($terms));
            }
        }
    }
```

Not saying you introduced this bug btw, it is already there but might be a good moment to look critical at it. :-)

src/Model/ValueObject/Taxonomy/Category/CategoryDomain.php
```diff
@@ -20,4 +20,9 @@ public function __construct(string $value)
         $this->guardNotEmpty($value);
         $this->setValue($value);
     }
+
+    public static function eventType(): CategoryDomain
+    {
+        return new CategoryDomain('eventtype');
```
grubolsch, line 26: You could use a constant for this

grubolsch (changes requested):
Huge refactor, clearly a ton of effort went into it! 
I just spotted one potential null bug to tweak, but everything else looks solid.

src/Event/Events/EventCreated.php
```diff
@@ -132,7 +136,7 @@ public static function deserialize(array $data): EventCreated
             $data['event_id'],
             new Language($data['main_language']),
             $data['title'],
-            EventType::deserialize($data['event_type']),
+            (new CategoryDenormalizer(CategoryDomain::eventType()))->denormalize($data['event_type'], Category::class),
```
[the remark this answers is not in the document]
LucWollants: The `CategoryDenormalizer` derives from `DenormalizerInterface` and the `denormalize` method requires the array and the class. Not much we can do about this I think.

We can add factory methods if you want but that is not in the scope of this cleanup.

src/Model/Serializer/ValueObject/Taxonomy/Category/CategoryDenormalizer.php
```diff
@@ -0,0 +1,53 @@
...
+use Symfony\Component\Serializer\Exception\UnsupportedException;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+final class CategoryDenormalizer implements DenormalizerInterface
+{
+    private ?CategoryDomain $defaultDomain;
+
+    public function __construct(?CategoryDomain $defaultDomain = null)
+    {
+        $this->defaultDomain = $defaultDomain;
```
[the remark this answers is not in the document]
LucWollants: There are several legacy classes that need to be removed, namely `EventType`, `Facility` and `Theme`. They all derive from `Category` and have a domain of `eventtype`, `facility` or `theme`. In the new model layer, we only have one `Category` and one denormalizer. But now this denormalizer also needs to be used for older data that was lacking the domain.

Once we are on a common interface this can be improved.

src/Model/ValueObject/Taxonomy/Category/CategoryDomain.php
LucWollants: But where would we reuse that constant?

src/Model/Serializer/ValueObject/Taxonomy/Category/CategoryDenormalizer.php
LucWollants: Good catch! But I'm hesitant to change it now. Maybe we need a follow up ticket for this?

grubolsch: I can live with that: https://jira.publiq.be/browse/III-6433

src/Model/ValueObject/Taxonomy/Category/CategoryDomain.php
grubolsch: For me that does not matter, I just try to avoid strings in my code anywhere (except tests).
Currently it is only used once, but code gets changed, and in the future somebody might need that value again, promoting code reuse if the const already is there.

grubolsch (approved):
CHanged it to approve because comments are either out of scope or moved to seperate ticket.

src/Event/Events/EventCreated.php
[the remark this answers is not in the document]
grubolsch: I understand that it is needed by the interface, but I would implement a wrapper around it.
But I agree it is out of scope for now.

src/Model/ValueObject/Taxonomy/Category/CategoryDomain.php
LucWollants: I can turn the question around. Why would this string literal ever be exposed, it is contained in this class?

If you need a `CategoryDomain` of the type `eventtype` you can use `CategoryDomain::eventType()` and if you need the string literal you can use `toString`. So I would say by adding the `const` and making it `public` you break the encapsulation and open it for usage as a `string` again instead of a typed class.

If you could start from scratch this would have been a simple enum, but because of the legacy data we need to take into account unexpected values.

src/Model/Serializer/ValueObject/Taxonomy/Category/CategoryDenormalizer.php
LucWollants: Great! Thanks for creating that ticket.

PR #1911, opened 2024-11-28
III-6434 Cleanup process duplicate places

grubolsch:
Improve the Script ProcessDuplicatePlaces .

### Removed

- Remove "start-cluster-id" from the option. It no longer makes sense to have this because the cluster ids are no longer auto increment ids, but hashes with no ordering.

---

Ticket: https://jira.uitdatabank.be/browse/III-6434


app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -45,14 +46,16 @@ public function __construct(
...
+        EventRelationsRepository $eventRelationsRepository,
+        LoggerInterface $logger
     ) {
         $this->duplicatePlaceRepository = $duplicatePlaceRepository;
         $this->duplicatePlaceRemovedFromClusterRepository = $duplicatePlaceRemovedFromClusterRepository;
         $this->canonicalService = $canonicalService;
         $this->eventBus = $eventBus;
         $this->placeEventFactory = $placeEventFactory;
         $this->eventRelationsRepository = $eventRelationsRepository;
+        $this->logger = $logger;
```
LucWollants, line 58: Why do we need a logger in the context of a CLI script?

app/Console/Command/ProcessDuplicatePlaces.php
```diff
@@ -152,7 +145,12 @@ public function execute(InputInterface $input, OutputInterface $output): int
                 foreach ($commands as $command) {
                     $output->writeln('Dispatching UpdateLocation for event with id ' . $command->getItemId());
                     if (!$dryRun) {
-                        $this->commandBus->dispatch($command);
+                        try {
+                            $this->commandBus->dispatch($command);
+                        } catch (MuseumPassNotUniqueInCluster $e) {
```
LucWollants, line 150: It's not the dispatch that will fail, but `getCanonical`

app/Console/Command/ProcessDuplicatePlaces.php
LucWollants: We normally use the standard output and use `$output` from the `execute` method. I don't think there is a specific file to log CLI command output. This would be the first CLI command that uses `LoggerInterface`.

LucWollants: Well there is one, `LoggerName::forService('fetching-movies', 'kinepolis')` but that is setup as a scheduled task with a dedicated log file.

LucWollants (approved):

PR #1922, opened 2024-12-04
III-6310 Remove legacy `EventType`

LucWollants:
### Removed
- Removed legacy `EventType`

---


src/EventExport/Format/HTML/HTMLEventFormatter.php
```diff
@@ -282,4 +285,17 @@ private function getAddressField(stdClass $event, string $addressField): string
 
         return $event->location->address->{$mainLanguage}->{$addressField} ?? '';
     }
+
+    private function eventTypeFromJSONLDEvent(string $eventString): ?Category
```
grubolsch, line 289: Personally I would import this as an assoc array and then check if the property "terms" and "domain" actually exist, if not return null. 
This very trusting that the json you will decode will always be a format you expect.

grubolsch (approved):

src/EventExport/Format/HTML/HTMLEventFormatter.php
LucWollants: If the "terms" or "domain" are missing it is caused by the JSON projection of that event. So I would say let it crash hard and then fix it instead of just ignoring it and keeping invalid data around.

PR #1926, opened 2024-12-05
III-6435 Look at date when the event happend

grubolsch:
### Changed

I made a small mistake withthe feature that records how an events gets moved to different places.
Instead of using the now() function for saving the date, I need to look at the date in the event store in case of a replay.

---

Ticket: https://jira.uitdatabank.be/browse/III-6435


src/Event/ReadModel/Relations/Doctrine/DBALEventPlaceHistoryRepository.php
```diff
@@ -18,27 +18,25 @@ public function __construct(Connection $connection)
...
-        $currentTimestamp = new DateTimeImmutable();
-
         $this->connection->insert(
             'event_place_history',
             [
                 'event' => $eventId->toString(),
                 'old_place' => $oldPlaceId ? $oldPlaceId->toString() : null,
                 'new_place' => $newPlaceId->toString(),
-                'date' => $currentTimestamp->format('Y-m-d H:i:s'),
+                'date' => $date->format('Y-m-d H:i:s'),
```
LucWollants, line 39: I would use the exact same format as `recordedOn` and just call `toString` on it to store exactly the same value as inside the metadata.

LucWollants (changes requested):
Looks good, but I wouldn't change the type of `recordedOn` and also store the same values as the metadata. The event store is very time sensitive so when only using seconds we can loose a lot of valuable information.

src/Event/ReadModel/Relations/Doctrine/DBALEventPlaceHistoryRepository.php
grubolsch: I am not using to String, I still use DateTimeInterface, but I did change all the formats to Atom. Is this oke?

LucWollants (approved):

PR #1935, opened 2024-12-10
III-6438 UUID refactor

grubolsch:
First in a sequence of prs to remove the references to Ramsey and hide them behind our Uuid DTO.

### Changed
- Add new method in CultuurNet\UDB3\Model\ValueObject\Identity\UUID.
- Added new constant NIL from old Uuid class to new Identity\UUID class
- Change all references from Ramsey to our internal DTO UUID class
---


src/Model/ValueObject/Identity/UUID.php
```diff
@@ -6,25 +6,33 @@
...
     use IsString;
     use MatchesRegexPattern;
 
+    public const NIL = '00000000-0000-0000-0000-000000000000';
+
     /**
      * Ensures backwards compatibility with older, malformed, uuids present in UDB.
      */
-    public const BC_REGEX = '\\A[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{12}\\z';
+    public const BC_REGEX = '/\\A[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{12}\\z/';
```
LucWollants, line 21: Maybe we need to make this private because it is nowhere used?

src/Model/ValueObject/Identity/UUID.php
```diff
@@ -6,25 +6,33 @@
...
 use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\IsString;
 use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\MatchesRegexPattern;
+use Ramsey\Uuid\Uuid as RamseyUuid;
 
 class UUID
 {
     use IsString;
     use MatchesRegexPattern;
 
+    public const NIL = '00000000-0000-0000-0000-000000000000';
```
LucWollants, line 16: There is also `LocationId::NIL_LOCATION` and some hardcoded strings, maybe we need to make that consistent?

LucWollants (changes requested):
After doing a quick search on `Ramsey\Uuid\Uuid` I still find direct usages of `Uuid` from the `Ramsey` package, mainly in tests.

src/Model/ValueObject/Identity/UUID.php
grubolsch: I use it in the new test tests/Model/ValueObject/Identity/UUIDTest.php.

LucWollants: I don't think that is needed. You could just create a valid and invalid UUID?

src/Event/ValueObjects/LocationId.php
```diff
@@ -4,14 +4,13 @@
...
 namespace CultuurNet\UDB3\Event\ValueObjects;
 
+use CultuurNet\UDB3\Model\ValueObject\Identity\UUID;
 use CultuurNet\UDB3\Model\ValueObject\String\Behaviour\IsString;
 
 final class LocationId
 {
     use IsString;
 
-    public const NIL_LOCATION = '00000000-0000-0000-0000-000000000000';
```
JonasVHG, line 13: Apart from the systemUser I cannot find any uses for a Nil Uuid apart from Location. Since we do not have Nil-events, organizers, Media,... This kinda feels too permissive to me. But I'm open for discussion.

LucWollants (approved):

LucWollants (approved):

PR #1936, opened 2024-12-10
III-6438 Change UUID to Uuid

grubolsch:
### Changed

Changed capitalisation of one class `UUID` to `Uuid` to make it more consistent with other DTO's like `Id`.
This pr seems larger then it is. ;-)

---


grubolsch:
Will pick this up in a seperate PR

PR #1940, opened 2024-12-11
III-6438 Add new phpstan rule to prevent ramsey namespace

grubolsch:
### Added

To enforce the refactor the use our own Uuid implementation instead of the Ramsey library provided one, I added a custom phpstan rule to verify that Ramsey is never imported.

This is the final PR for this ticket.
---

### DIFF
--- a/app/Phpstan/ForbiddenRamseyNamespaceRule.php
+++ b/app/Phpstan/ForbiddenRamseyNamespaceRule.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Phpstan;
 
+use CultuurNet\UDB3\Model\ValueObject\Identity\Uuid;
 use PhpParser\Node;
 use PhpParser\Node\Stmt\Use_;
 use PHPStan\Analyser\Scope;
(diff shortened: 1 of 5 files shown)


phpstan/ForbiddenRamseyNamespaceRule.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        foreach ($node->uses as $use) {
+            if (! str_contains($use->name->toString(), 'Ramsey')) {
+                continue;
+            }
+
+            $filePath = $scope->getFile();
+            if (str_ends_with($filePath, 'src/Model/ValueObject/Identity/Uuid.php')) {
```
LucWollants, line 27: If the Uuid file is moved this will break. Is it not possible to add and ignore comment?

grubolsch: good idea, lets see...

grubolsch: it works!

app/Phpstan/ForbiddenRamseyNamespaceRule.php
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        foreach ($node->uses as $use) {
+            if (! str_contains($use->name->toString(), 'Ramsey')) {
+                continue;
+            }
+
+            return [
+                sprintf(
+                    'The "Ramsey" namespace is not allowed in file: %s, please us CultuurNet\UDB3\Model\ValueObject\Identity\Uuid',
```
LucWollants, line 32: Nit: use `Uuid::class`?

PR #1942, opened 2024-12-13
III-6213 Update history projector fields

grubolsch:
### Changed
- Add `clientId` and `clientName` fields
- Add new keys in case old ones are missing because of postponed replay
- Cleanup some other leftover references of auth0, all internal variables, should not result in behavior changes.

---

Ticket: https://jira.uitdatabank.be/browse/III-6213


grubolsch:
This can only be merged after the field is also changed in the frontend.

LucWollants:
> This can only be merged after the field is also changed in the frontend. See ticket [III-6443](https://jira.publiq.be/browse/III-6443)

The ticket contained an approach to make the changes independent between backend and frontend. First, add the new field and keep the old field in the backend. Then, modify the frontend to use the new field.

Also, take into account that a replay will be needed to fix this unless we also work with a class to convert the old name to the new name for old projections. Like we do for the JSON projections with the polyfil.

LucWollants:
> Main change is the metadata log of the event store will no longer have references to auth0.

This will not modify the metadata from the event store, that data was already independent from Auth0. What will change is the history projector which uses the existing metadata to create JSON fields. What is changed here are just the names of those fields.

PR #1951, opened 2024-12-20
III-6453 Provide fallback when calendar type does not match sub event count

LucWollants:
### Changed
- Provided fallback when calendar type does not match sub event count

---
Ticket: https://jira.uitdatabank.be/browse/III-6453


src/Model/Serializer/ValueObject/Calendar/CalendarSerializer.php
```diff
@@ -87,6 +87,15 @@ function ($openingHourData) {
             );
         }
 
+        // There are cases where the calendar type does not match the number of sub events.
+        // For those cases the amount of sub events dictates the calendar type.
+        if ($calendarType->sameAs(CalendarType::single()) && count($subEvents) > 1) {
```
simon-debruijn, line 92: Wouldn't a better solution be creating new events in the event store to fix these events?

simon-debruijn (changes requested):

src/Model/Serializer/ValueObject/Calendar/CalendarSerializer.php
LucWollants: This code is used to read the JSON payload from the existing and new events inside the event store. The new events which are created with commands with better validation will have the correct structure.

simon-debruijn (approved):

PR #1953, opened 2025-01-08
III-6456 cli search trailer

JonasVHG:
### Added

- `FindMatchingTrailersOnYouTube`: CLI-script to add matching trailers from YouTube to Kinepolis movies.

### Changed

- `ConsoleServiceProvider`: Register `FindMatchingTrailersOnYouTube`

### Note

- In a follow-up PR the trailerSearch within the KinepolisService will be removed, to have a clear separation of responsibilities.

---

Ticket: https://jira.publiq.be/browse/III-6456

### DIFF
--- a/app/Console/Command/AddMatchingTrailer.php
+++ b/app/Console/Command/AddMatchingTrailer.php
@@ -16,6 +16,7 @@ use CultuurNet\UDB3\Search\ResultsGenerator;
 use CultuurNet\UDB3\Search\ResultsGeneratorInterface;
 use CultuurNet\UDB3\Search\SearchServiceInterface;
 use CultuurNet\UDB3\Search\Sorting;
+use Google\Service\Exception as GoogleException;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -74,7 +75,12 @@ final class AddMatchingTrailer extends Command
 
         $results = $this->searchResultsGenerator->search($query);
         foreach ($results as $eventId => $result) {
-            $this->dispatchAddVideoCommand($eventId, $output);
+            try {
+                $this->dispatchAddVideoCommand($eventId, $output);
+            } catch (GoogleException $exception) {
+                $output->writeln($exception->getMessage());
+                break;
+            }
         }
 
         return 0;
--- a/app/Console/Command/AddMatchingTrailer.php
+++ b/app/Console/Command/AddMatchingTrailer.php
@@ -9,6 +9,7 @@ use Cake\Chronos\Chronos;
 use CultuurNet\UDB3\Json;
 use CultuurNet\UDB3\Kinepolis\Trailer\TrailerRepository;
 use CultuurNet\UDB3\Model\ValueObject\Identity\Uuid;
+use CultuurNet\UDB3\Model\ValueObject\MediaObject\Video;
 use CultuurNet\UDB3\Offer\Commands\Video\AddVideo;
 use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
 use CultuurNet\UDB3\ReadModel\DocumentRepository;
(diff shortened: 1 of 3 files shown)


app/Console/Command/FindMatchingTrailersOnYouTube.php
```diff
@@ -0,0 +1,116 @@
...
+use CultuurNet\UDB3\ReadModel\JsonDocument;
+use CultuurNet\UDB3\Search\ResultsGenerator;
+use CultuurNet\UDB3\Search\ResultsGeneratorInterface;
+use CultuurNet\UDB3\Search\SearchServiceInterface;
+use CultuurNet\UDB3\Search\Sorting;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class FindMatchingTrailersOnYouTube extends Command
```
LucWollants, line 23: It does more than just searching a movie on YouTube, it also adds a matching trailer. Maybe rename the class to something like `AddMatchingTrailer`?

Because an interface `TrailerRepository` is used it doesn't need to be bound to YouTube?

app/Console/Command/FindMatchingTrailersOnYouTube.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->commandBus = $commandBus;
+        $this->searchResultsGenerator = new ResultsGenerator(
+            $searchService,
+            new Sorting('created', 'desc'),
+            100
+        );
+        $this->documentRepository = $documentRepository;
+        $this->trailerRepository = $trailerRepository;
+    }
+    public function configure(): void
```
LucWollants, line 49: Can use a newline

app/Console/Command/FindMatchingTrailersOnYouTube.php
```diff
@@ -0,0 +1,116 @@
...
+        SearchServiceInterface $searchService,
+        DocumentRepository $documentRepository,
+        TrailerRepository $trailerRepository
+    ) {
+        parent::__construct();
+        $this->commandBus = $commandBus;
+        $this->searchResultsGenerator = new ResultsGenerator(
+            $searchService,
+            new Sorting('created', 'desc'),
+            100
```
LucWollants, line 44: I would search more and catch the exception from the quota limit and then just stop?

app/Console/Command/FindMatchingTrailersOnYouTube.php
```diff
@@ -0,0 +1,116 @@
...
+        foreach ($results as $eventId => $result) {
+            $this->dispatchAddVideoCommand($eventId, $output);
+        }
+
+        return 0;
+    }
+
+    private function getQueryForMoviesWithoutTrailer(): string
+    {
+        return 'terms.id:0.50.6.0.0 AND videosCount:0 AND creator:' . Uuid::NIL;
```
LucWollants, line 75: Should we go back 1 week?

app/Console/Command/FindMatchingTrailersOnYouTube.php
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    protected function getDocument(string $id): ?JsonDocument
```
LucWollants, line 78: I think this can be private?

app/Console/Command/FindMatchingTrailersOnYouTube.php
JonasVHG: I'm not in favor to just catch the Exception. If we know the limit before, I think it's best to keep it in mind.

LucWollants: Also fine, but maybe make it an appconfig then?

app/Console/Command/AddMatchingTrailer.php
```diff
@@ -0,0 +1,127 @@
...
+        if ($count === 0) {
+            $output->writeln('Could not find any movies without trailer.');
+            return 0;
+        }
+
+        $results = $this->searchResultsGenerator->search($query);
+        foreach ($results as $eventId => $result) {
+            try {
+                $this->dispatchAddVideoCommand($eventId, $output);
+            } catch (GoogleException $exception) {
```
LucWollants, line 80: This is confusing that a dispatch would also throw a `GoogleException`, I think that indicates that finding the video should be a separate method from dispatching the actual command. The method `dispatchAddVideoCommand` does both the find and the dispatch but that is not clear from the name.

LucWollants (approved):

PR #1956, opened 2025-01-14
III-6465 defaultQuery when using a ClientId

JonasVHG:
### Changed

- `features/search/default-queries-config-file.feature`: Renamed to `features/search/default-queries-config-file-apikey.feature` and add clarification that's is only for `apiKeys`. 
- `default-queries-config-file-clientid.feature`: Acceptance tests for searching with a clientId that has a `defaultQuery`.

### Related PR

- https://github.com/cultuurnet/udb3-search-service/pull/330
- https://github.com/cultuurnet/appconfig/pull/924

---

Ticket: https://jira.publiq.be/browse/III-6465

### DIFF
--- a/features/search/default-queries-config-file-clientid.feature
+++ b/features/search/default-queries-config-file-clientid.feature
@@ -1,7 +1,7 @@
 @sapi3
 Feature: Test the Search API v3 default queries from the config file when using an clientId
 
-  # sapi3KeyWithConfigFileFilterForScherpenheuvel has a default filter with 'regions:nis-24134' in the config file
+  # test_client_with_default_search_query has a default filter with 'regions:nis-24134' in the config file
   Background:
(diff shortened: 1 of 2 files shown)


LucWollants (approved):

features/search/default-queries-config-file-clientid.feature
```diff
@@ -0,0 +1,59 @@
+@sapi3
+Feature: Test the Search API v3 default queries from the config file when using an clientId
+
+  # test_client_with_default_search_query has a default filter with 'regions:nis-24134' in the config file
```
LucWollants, line 4: Only one small remark, maybe we need to change the default query to make sure it is the client ID default query and not the API key default query?

JonasVHG: Good remark will do it like that.

PR #1958, opened 2025-01-14
III-6448 rdf organisation add images

grubolsch:
### Added
- Add images to organisation RDF projection

---

Ticket: https://jira.uitdatabank.be/browse/III-6448

### DIFF
--- a/src/Model/ValueObject/MediaObject/ImagesToMediaObjectReferencesConvertor.php
+++ b/src/Model/ValueObject/MediaObject/ImagesToMediaObjectReferencesConvertor.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Model\ValueObject\MediaObject;
+
+use CultuurNet\UDB3\Media\MediaObject;
+use CultuurNet\UDB3\Media\MediaObjectRepository;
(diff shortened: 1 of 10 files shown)


src/Model/ValueObject/MediaObject/ImagesToMediaObjectReferencesConvertor.php
```diff
@@ -0,0 +1,44 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Model\ValueObject\MediaObject;
+
+use CultuurNet\UDB3\Media\MediaObject;
+use CultuurNet\UDB3\Media\MediaObjectRepository;
+use CultuurNet\UDB3\Model\ValueObject\MediaObject\MediaObject as MediaObjectDto;
+
+class ImagesToMediaObjectReferencesConvertor
```
LucWollants, line 11: As discussed through Slack a better approach could be to work with `Image` inside the `ImageEditor` and use the `ImageNormalizer` to get the URL of the thumbnail. But this can be done in a follow-up pull request.

LucWollants (approved):

LucWollants:
As discussed through Slack there are several issues with the current implementation for media:
- The `Broadway` repository of the `Image` aggregate is used to return a JSON representation, this should be done with a dedicated JSON projector
- The `MediaObject` is only used for images, this can be simplified to use an `Image` as it is also done for `Organizer`

PR #1960, opened 2025-01-16
III-6471 Media object editor works directly with images

grubolsch:
### Changed
- Made the MediaObjectEditor (now renamed to ImageEditor) accept Images directly instead of MediaObjectReferences.
This is because we want to refactor MediaObjectReferences away in the future, this is the first step.

---

Ticket: https://jira.uitdatabank.be/browse/III-6471


grubolsch:
This is just the first part to simplify the code around Images as discussed in the ticket.

app/Media/MediaServiceProvider.php
```diff
@@ -150,10 +150,11 @@ function () use ($container) {
         );
 
         $container->addShared(
-            ImagesToMediaObjectReferencesConvertor::class,
+            ImageNormalizer::class,
```
LucWollants, line 153: If we expose `ImageNormalizer` here I would also change the other usages, or we also inline the constructor in the service providers. Now we have a mix.

LucWollants (approved):
Looks good, only one small remark about the exposure of `ImageNormalizer`

app/Media/MediaServiceProvider.php
LucWollants: Modified with https://github.com/cultuurnet/udb3-backend/pull/1960/commits/cb7543479b4a848bfb92132fb8f8c97bf9a28962

PR #1963, opened 2025-01-22
III-6494 RDF convert blank nodes to named hashes

grubolsch:
### Added
- Added 2 new classes and tests to convert blank nodes in the RDF projection to named nodes with a URI. This URI will contain a unique hash based on the content of the node. Currently this PR does not do anything, it is just for preperation for the work in https://jira.publiq.be/browse/III-6450

---

Ticket: https://jira.uitdatabank.be/browse/III-6494

### DIFF
--- a/tests/RDF/NodeUri/CRC32HashGeneratorTest.php
+++ b/tests/RDF/NodeUri/CRC32HashGeneratorTest.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\RDF\NodeUri;
+
+use CultuurNet\UDB3\Json;
+use PHPUnit\Framework\TestCase;
+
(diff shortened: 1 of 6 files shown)


tests/RDF/NodeUri/CRC32HashGeneratorTest.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $this->hashGenerator = new CRC32HashGenerator();
+    }
+
+    /**
+     * @test
+     */
+    public function it_should_generate_produces_consistent_hash(): void
+    {
+        $data = [
```
LucWollants, line 23: I would take a more realistic example, for example a translated address, or a complex booking information.

Furthermore, this is already a sorted array and it doesn't test the recursive implementation.

JonasVHG (approved):

PR #1968, opened 2025-01-23
III-6471 Remove `MediaObjectReference`

LucWollants:
### Removed
- Removed `MediaObjectReference`

---

Ticket: https://jira.uitdatabank.be/browse/III-6471


src/Model/Import/MediaObject/MediaManagerImageCollectionFactory.php
```diff
@@ -46,20 +45,20 @@ function (MediaObjectReference $mediaObjectReference) {
...
             },
-            $mediaObjectReferences->toArray()
+            $images->toArray()
         );
 
-        $images = array_filter($images);
+        $imageCollection = array_filter($imageCollection);
 
-        return ImageCollection::fromArray($images);
+        return ImageCollection::fromArray($imageCollection);
```
simon-debruijn, line 62:
```suggestion
        return ImageCollection::fromArray($mediaImages);
```

simon-debruijn (approved):

src/Model/Import/MediaObject/MediaManagerImageCollectionFactory.php
LucWollants: That makes indeed much more sense. The final goal is to remove the `Image` object and `ImageCollection` classes from the `CultuurNet\UDB3\Media\Image` namespace and only work with the new classes from the `CultuurNet\UDB3\Model\ValueObject\MediaObject\` namespace. 

But that requires even more work ... to be continued ... 😏

PR #1971, opened 2025-01-24
III-6480 rdf convert blanknode for periodoftime

grubolsch:
### Added
- rdf convert blanknode for periodoftime
- Did not make a new normalizer for this one because it seemed to small with very limited use.

---

Ticket: https://jira.uitdatabank.be/browse/III-6480


src/Model/ValueObject/Calendar/DateRange.php
```diff
@@ -40,4 +41,12 @@ public function compare(DateRange $dateRange): int
 
         return 0;
     }
+
+    public function toArray(): array
```
LucWollants, line 45: Nit: not sure if we need to expose it like this

LucWollants (approved):

src/Model/ValueObject/Calendar/DateRange.php
grubolsch: How would you prefer it? I can make a normalizer if you prefer, but I thought that would be so small, and probably never re-used that I opted for this approach.

LucWollants: Either a new normalizer or inline.

PR #1972, opened 2025-01-27
III-6487 RDF convert blanknode for geometry

grubolsch:
### Added
- Added named nodes for Geometry RDF projection.
- Added to _OrganizerJsonToTurtleConverter_ and _PlaceJsonToTurtleConverter_ the new RdfResourceFactory dependency.

---

Ticket: https://jira.uitdatabank.be/browse/III-6487

### DIFF
--- a/tests/Organizer/ReadModel/RDF/OrganizerJsonToTurtleConverterTest.php
+++ b/tests/Organizer/ReadModel/RDF/OrganizerJsonToTurtleConverterTest.php
@@ -9,6 +9,9 @@ use CultuurNet\UDB3\Address\Parser\ParsedAddress;
 use CultuurNet\UDB3\Iri\CallableIriGenerator;
 use CultuurNet\UDB3\Json;
 use CultuurNet\UDB3\Model\Serializer\Organizer\OrganizerDenormalizer;
+use CultuurNet\UDB3\RDF\NodeUri\CRC32HashGenerator;
+use CultuurNet\UDB3\RDF\NodeUri\NodeUriGenerator;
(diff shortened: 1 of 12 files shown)


tests/Organizer/ReadModel/RDF/OrganizerJsonToTurtleConverterTest.php
```diff
@@ -73,6 +76,7 @@ protected function setUp(): void
             new OrganizerDenormalizer(),
             $addressParser,
             $this->imageNormalizer,
+            new RdfResourceFactoryWithoutBlankNodes(new NodeUriGenerator(new CRC32HashGenerator())),
```
LucWollants, line 79: Could it be that you forgot to use `RdfResourceFactory` from the container which takes into account the `('config')['rdf']['blank_nodes_allowed']` ?

If so, I think that is also wrong on other classes.

grubolsch: This is for the test classes, not the actual implementation. Because I updated the expected outputs, this is required.
The downside of course it that currently the unit tests don't test the old versions anymore.
I want to fix that I will have to duplicate all the unit tests for the RDF projections to the old outputs.

Given that I hope this only lives for a week or 2, and given how much code this would double, and needs to be removed later - I prefer this approach.

LucWollants: Ignore my comment, because it's not an issue to hardcode the flavour for the unit tests.

LucWollants (approved):

PR #1976, opened 2025-01-28
From and to are always required for this class, remove nulls

grubolsch:
### Changed
- From and to are always required for this class, remove nulls


LucWollants (approved):
Looks good! To be sure I also started the acceptance tests on Jenkins.

PR #1981, opened 2025-01-29
III-6486 RDF convert blanknode for contact point

grubolsch:
### Changed
- RDF convert blanknode for contact point

---

Ticket: https://jira.uitdatabank.be/browse/III-6486


LucWollants (approved):
I already approved, because only a minor issue.

PR #1991, opened 2025-02-06
III-6518 RDF Make sure capitalisation of hashes is correct

grubolsch:
Once again this looks like a very big PR, but in fact it only contains 1 extra line of domain code, the rest are test.

### Changed
- Make sure capitalisation of hashes is correct

---

Ticket: https://jira.uitdatabank.be/browse/III-6518

### DIFF
--- a/features/data/organizers/rdf/organizer-with-images.ttl
+++ b/features/data/organizers/rdf/organizer-with-images.ttl
@@ -13,23 +13,23 @@
   a cp:Organisator ;
   dcterms:created ""^^xsd:dateTime ;
   dcterms:modified ""^^xsd:dateTime ;
-  adms:identifier <http://data.uitdatabank.local:80/organizers/%{organizerId}#Identifier-%{identifier}> ;
+  adms:identifier <http://data.uitdatabank.local:80/organizers/%{organizerId}#identifier-%{identifier}> ;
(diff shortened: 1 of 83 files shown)


src/RDF/NodeUri/NodeUriGenerator.php
```diff
@@ -20,6 +20,9 @@ public function generate(string $nodeName, array $fields): string
             $nodeName = substr($nodeName, strrpos($nodeName, ':') + 1);
         }
 
+        // Convert AddressDetail to addressDetail
+        $nodeName = mb_strtolower(substr($nodeName, 0, 1)) . substr($nodeName, 1);
```
LucWollants, line 24: Would https://www.php.net/manual/en/function.lcfirst.php also work?

LucWollants (approved):

features/data/organizers/rdf/organizer-with-images.ttl
```diff
@@ -13,31 +13,31 @@
   a cp:Organisator ;
   dcterms:created ""^^xsd:dateTime ;
```
Stiksels, line 14: is het ok om hier een lege string te hebben staan?

Stiksels (approved):
@grubolsch very impressive list of changes! They all look very good!

PR #1993, opened 2025-02-07
Fixed updating non existing organizers throwing 404 and not 403

Anahkiasen:
### Fixed
- Fixed updating non existing organizers throwing 404 and not 403

---

Uses the same middleware approach as how it was fixed for offers and places

Ticket: https://jira.publiq.be/browse/III-5324


LucWollants (approved):
Looks good!

PR #1994, opened 2025-02-12
III-6366 Set up ownership mailings

grubolsch:
To prepare to sending ownership mails, we want to remember if a mail has been sent already or not.
For this I added an extra database table.

## Notes:
- I tried to make it generic named (hence identifier instead of organiserId) for reuse for other mails later
- I also note down the email / date of the mail, while not technically required this will make it very practical if we ever need to lookup if a specific mail was sent.

### Added
- Mail sent migration + repo + test

---

### DIFF
--- a/app/Migrations/Version20250212071939.php
+++ b/app/Migrations/Version20250212071939.php
@@ -15,7 +15,7 @@ class Version20250212071939 extends AbstractMigration
         $table = $schema->createTable('mails_sent');
 
         $table->addColumn('identifier', Types::GUID)->setLength(36)->setNotnull(true);
-        $table->addColumn('email', Types::STRING)->setLength(100)->setNotnull(true);
+        $table->addColumn('email', Types::STRING)->setLength(320)->setNotnull(true);
         $table->addColumn('type', Types::STRING)->setLength(100)->setNotnull(true);
         $table->addColumn('dateTime', Types::DATETIME_IMMUTABLE)->setNotnull(true);
 
--- a/app/Migrations/Version20250212071939.php
+++ b/app/Migrations/Version20250212071939.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\Migrations\AbstractMigration;
+
+class Version20250212071939 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable('mails_sent');
+
+        $table->addColumn('identifier', Types::GUID)->setLength(36)->setNotnull(true);
+        $table->addColumn('email', Types::STRING)->setLength(100)->setNotnull(true);
+        $table->addColumn('type', Types::STRING)->setLength(100)->setNotnull(true);
+        $table->addColumn('dateTime', Types::DATETIME_IMMUTABLE)->setNotnull(true);
+
+        $table->setPrimaryKey(['identifier', 'type']);
+    }
+
+    public function down(Schema $schema): void
+    {
+        $schema->dropTable('mails_sent');
+    }
+}
--- a/src/Mailer/DBALMailsSentRepository.php
+++ b/src/Mailer/DBALMailsSentRepository.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailer;
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\Uuid;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddress;
+use DateTimeInterface;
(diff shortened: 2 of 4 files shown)


app/Migrations/Version20250212071939.php
```diff
@@ -0,0 +1,29 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+class Version20250212071939 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable('mails_sent');
+
+        $table->addColumn('identifier', Types::GUID)->setLength(36)->setNotnull(true);
+        $table->addColumn('email', Types::STRING)->setLength(100)->setNotnull(true);
```
steftrenson, line 18: Apparently the longest possible email address can be 320 characters. Might never happen off course but my gut feeling said 100 feels very arbitrary.

app/Migrations/Version20250212071939.php
```diff
@@ -0,0 +1,29 @@
...
+        $table->addColumn('type', Types::STRING)->setLength(100)->setNotnull(true);
+        $table->addColumn('dateTime', Types::DATETIME_IMMUTABLE)->setNotnull(true);
+
+        $table->setPrimaryKey(['identifier', 'type']);
```
LucWollants, line 22: I think this can result in issues because for one ownership there can be 2 emails, one for the created and one for the approved.

This also makes me wonder if we need to be more explicit about the column names instead of having these generic column names.

grubolsch: 1) This is why Type is included in the primary key, so a record will read something like
id-of-the-ownership-request - request_approved - koen@publiq.be - some date

2) I can see that, I would really like to try if we can make do with generic names because I don't want to end up in a situation where we start sending more mails from UDB and we end up with multiple tables all tracking sent emails.

LucWollants: Ah, sorry, I missed the type part.

src/Mailer/DBALMailsSentRepository.php
```diff
@@ -0,0 +1,50 @@
...
+    public function addMailSent(Uuid $identifier, EmailAddress $email, string $type, DateTimeInterface $dateTime): void
+    {
+        $this->connection->createQueryBuilder()
+            ->insert('mails_sent')
+            ->setValue('identifier', ':identifier')
+            ->setValue('email', ':email')
+            ->setValue('type', ':type')
+            ->setValue('dateTime', ':dateTime')
+            ->setParameters([
+                ':identifier' => $identifier->toString(),
```
LucWollants, line 43: The `:` is not needed in the `setParameters` method

src/Mailer/DBALMailsSentRepository.php
```diff
@@ -0,0 +1,50 @@
...
+                ->execute()
+                ->fetchOne()
+            !== false;
+    }
+
+    public function addMailSent(Uuid $identifier, EmailAddress $email, string $type, DateTimeInterface $dateTime): void
+    {
+        $this->connection->createQueryBuilder()
+            ->insert('mails_sent')
+            ->setValue('identifier', ':identifier')
```
LucWollants, line 38: You could also use the convenience method `values`

```
->values([
   'identifier' => ':identifier',
   'email'      => ':email',
   'type'       => ':type',
   'dateTime'   => ':dateTime',
])
```

tests/Mailer/DBALMailsSentRepositoryTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->assertEquals($isSent, $this->repository->isMailSent($identifierMatched, $typeMatched));
+    }
+
+    public function mailSentDataProvider(): array
+    {
+        $uuid = Uuid::uuid4();
+        $uuid2 = Uuid::uuid4();
+
+        return [
+            [$uuid, OwnershipApproved::class, $uuid, OwnershipApproved::class, true],
```
LucWollants, line 76: Maybe add a `string` index to explain what every scenario does?

tests/Mailer/DBALMailsSentRepositoryTest.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * @dataProvider mailSentDataProvider
+     * @test
+     */
+    public function handles_is_mail_sent(Uuid $identifier, string $type, Uuid $identifierMatched, string $typeMatched, bool $isSent): void
+    {
+        $email = new EmailAddress('koen@publiq.be');
+        $date = DateTimeImmutable::createFromFormat(DateTimeInterface::ATOM, self::DATE_TIME_VALUE);
+
+        $this->connection->insert('mails_sent', [
```
LucWollants, line 60: I wonder why you didn't use this elegant shorthand inside the real implementation of the `DBALMailsSentRepository`? 😊

grubolsch: good point!

LucWollants:
@grubolsch I don't see any of the proposed fixes on this feature branch. Could it be you made the changes on the wrong branch?

PR #1997, opened 2025-02-13
III-6366 Mailer smtp class

grubolsch:
### Added
- Installed Symfony mailer package
- Provide a mailer class to sent mails by SMTP
- Add the twig templates that will be used in the feature for the mails
- Providing an option to only sent mails to certain domains for testing purposes

---

Related to: https://github.com/cultuurnet/appconfig/pull/956

### DIFF
--- a/src/Mailer/SmtpMailer.php
+++ b/src/Mailer/SmtpMailer.php
@@ -21,13 +21,16 @@ final class SmtpMailer implements Mailer
     private MailerInterface $mailer;
     private LoggerInterface $logger;
     private Address $from;
+    /** @var string[] */
+    private array $whiteListedDomains;
 
-    public function __construct(TwigEnvironment $twig, MailerInterface $mailer, LoggerInterface $logger, Address $from)
+    public function __construct(TwigEnvironment $twig, MailerInterface $mailer, LoggerInterface $logger, Address $from, array $whiteListedDomains)
     {
         $this->twig = $twig;
         $this->mailer = $mailer;
         $this->logger = $logger;
         $this->from = $from;
+        $this->whiteListedDomains = $whiteListedDomains;
     }
 
     public function send(EmailAddress $to, string $subject, string $htmlTemplate, string $textTemplate, array $variables = []): bool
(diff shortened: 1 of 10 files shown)


src/Mailer/SmtpMailer.php
```diff
@@ -0,0 +1,83 @@
...
+        } catch (LoaderError|SyntaxError|RuntimeError $e) {
+            $this->logger->error('[TwigTemplate] ' . $e->getMessage());
+            return false;
+        }
+
+        $email = (new Email())
+            ->from($this->from)
+            ->to($to->toString())
+            ->subject($subject)
+            ->text($text)
```
simon-debruijn, line 50: I think the rendering of the txt template can be done automatically based on the html by symphony

simon-debruijn (approved):

src/Mailer/SmtpMailer.php
grubolsch: I just tested this but this seems not accurate, but maybe I am missing a detail?
- When sending the HTML template instead of the text the HTML is still in the plain text version.
- When not calling the text() method it does not sent a plain text version at all.

grubolsch: Not sure if we want that. I think for links this would be a problem, because then the URL would be lost because it is inside the href property.

JonasVHG (approved):

src/Mailer/SmtpMailer.php
```diff
@@ -0,0 +1,83 @@
...
+            return false;
+        }
+    }
+
+    /*
+     * Checks if whitelisted domains are configured, in which case only mails to those domains can be sent.
+     * This is used on acc / test.
+     * If empty, on production, it will send mails to everybody.
+     * */
+    private function isAllowedToSentEmail(EmailAddress $email): bool
```
LucWollants, line 69: Why would we do something like this if we could just configure a tool like Mailtrap or Mailcatcher?

LucWollants:
@grubolsch I see you removed the Twig parsing from the mail sender, but now the Twig parsing is missing?
I think you also need to update the lock file because you merged changes from master.

src/Mailer/SmtpMailer.php
LucWollants: I guess one of the advantages of managing the templates ourselves was the option to use mail-catching tools. So there is no need for this configuration.

grubolsch:
> @grubolsch I see you removed the Twig parsing from the mail sender, but now the Twig parsing is missing?

Well yes, because the SmtpMailer is not used anywhere concretly yet (except in a unit test).
So there is no place where I can move the twig logic.

I will have to add it to different PRs, where I actually add the logic to sent the emails.

grubolsch:
> @grubolsch I think you also need to update the lock file because you merged changes from master.

Updated the lock file, no longer any failed tests. :-)

LucWollants (approved):

PR #2000, opened 2025-02-13
III-6366 Sent ownership requested email

grubolsch:
### Added
- Sent "ownership requested" emails to the owner.

---


grubolsch:
Might need some more changes later because of how I determine who is an owner (see slack discussion).

grubolsch:
Also an open question is, if we only need to mails in Dutch. Will discuss this with Ben on Monday.

app/EventBus/EventBusServiceProvider.php
```diff
@@ -88,6 +89,7 @@ function () use (&$eventBus, $container): void {
                             OwnershipSearchProjector::class,
                             OwnershipPermissionProjector::class,
                             ImageLDProjector::class,
+                            SendMailsForOwnership::class,
```
LucWollants, line 92: I think there is one big drawback with this approach. The event listeners are also processed synchronously. This means that the call to create, approve or reject an ownership is blocked until the mail is set.

Maybe we need to look into the approach of an export where an asynchronous command is used with a dedicated worker. Check for the usage of `createResqueCommandBus`

LucWollants: Maybe we need an extra step, basically a conversion between the event and an async command to send a mail. So the `OwnershipRequested` is handled by service that create an async `SendMail` command.

app/CommandHandling/CommandBusServiceProvider.php
```diff
@@ -250,6 +253,15 @@ function () use ($container) {
             }
         );
 
+        $container->addShared(
+            MailerCommandBus::class,
+            function () {
+                $commandBus = $this->createResqueCommandBus('event_export', $this->container);
```
steftrenson, line 259: Shouldn't this have it's own specific queue iso reusing the `event_export` queue?

src/Mailer/Event/SentMail.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailer\Event;
+
+use CultuurNet\UDB3\Model\ValueObject\Identity\Uuid;
+use CultuurNet\UDB3\Model\ValueObject\Web\EmailAddress;
+
+final class SentMail
```
steftrenson, line 10: I think this should be `SendMail` and be in a command namespace rather than the event one. Since it's used as a command to actually send an email, and is handled by a commandhandler.

src/Mailer/MailSentCommandHandler.php
```diff
@@ -0,0 +1,54 @@
...
+namespace CultuurNet\UDB3\Mailer;
+
+use Broadway\CommandHandling\CommandHandler;
+use CultuurNet\UDB3\CommandHandling\AsyncCommand;
+use CultuurNet\UDB3\CommandHandling\AsyncCommandTrait;
+use CultuurNet\UDB3\Mailer\Event\SentMail;
+use DateTimeImmutable;
+use Psr\Log\LoggerInterface;
+
+class MailSentCommandHandler implements CommandHandler, AsyncCommand
```
steftrenson, line 14: Since the command is `SendMail` this should probably be `SendMailCommandHandler`

src/Mailer/Ownership/SendMailsForOwnership.php
```diff
@@ -20,37 +20,38 @@
 use CultuurNet\UDB3\User\UserIdentityResolver;
 use DateTimeInterface;
 use Psr\Log\LoggerInterface;
+use Twig\Environment as TwigEnvironment;
 
 final class SendMailsForOwnership implements EventListener
 {
     private const SUBJECT_OWNERSHIP_REQUESTED = 'Beheers aanvraag voor organisatie {{ organisationName }}';
 
+    private ResqueCommandBus $mailerCommandBus;
```
steftrenson, line 29: It's better to typehint on the `CommandBus` interface rather than the `ResqueCommandBus`. This couples the `SendMailsForOwnership` class too tightly to the Resque command bus, and it doesn't actually matter for the class which kind of bus it sends its commands to.

src/Mailer/MailSentCommandHandler.php
```diff
@@ -0,0 +1,54 @@
...
+    public function __construct(Mailer $mailer, MailsSentRepository $mailsSentRepository, LoggerInterface $logger)
+    {
+        $this->mailer = $mailer;
+        $this->mailsSentRepository = $mailsSentRepository;
+        $this->logger = $logger;
+    }
+
+    public function handle($command): void
+    {
+        if (!$command instanceof SentMail) {
```
steftrenson, line 31: This shouldn't be necessary when you add a typehint to the `handle()`. If I look at `EventExportCommandHandler` it looks like the commanbus is auto-wired and selects its commandhandler based on the typehints.

src/Mailer/Ownership/SendMailsForOwnership.php
```diff
@@ -79,12 +80,8 @@ public function handle(DomainMessage $domainMessage): void
         }
     }
 
-    private function handleOwnershipRequested(OwnershipRequested $ownershipRequested, DateTimeInterface $dateTime): void
+    private function handleOwnershipRequested(OwnershipRequested $ownershipRequested): void
     {
-        if ($this->mailsSentRepository->isMailSent(new Uuid($ownershipRequested->getId()), OwnershipRequested::class)) {
```
steftrenson: Actually, I think this does still make sense to do here. If we already know the mail is sent, it's not necessary to put it on a queue and handle it again.

src/Mailer/MailSentCommandHandler.php
grubolsch: The refactored handler will handle 3 types of Commands later on.

src/Mailer/Ownership/SendMailsForOwnership.php
grubolsch: I changed it to a ContextDecoratedCommandBus, because it needs a context to run as a worker in Resque.

src/Mailer/MailSentCommandHandler.php
grubolsch: Changed to `SendOwnershipMailCommandHandler`

app/CommandHandling/CommandBusServiceProvider.php
grubolsch: That part was a WIP, it has completely changed by now.

app/EventBus/EventBusServiceProvider.php
grubolsch: All of this has been implemented, with the necessary headache :-)

src/Mailer/Ownership/SendMailsForOwnership.php
grubolsch: All logic has completly been moved to the redis queue. It is a very fast, lightweight queue in memory, and I would not want to run this check twice.
I would prefer to keep it, together with all the checks, in the SendOwnershipMailCommandHandler.

grubolsch:
Ready for final review, mails are now async sent.
I am aware that I am currently not sending the email to the correct person, a follow up PR will be created to fix this.
If you want we can wait to merge both pr's a the same time, but this PR is already becoming quite big.

LucWollants:
> If you want we can wait to merge both pr's a the same time, but this PR is already becoming quite big.

When using the feature flag we can disable it but still merge and deploy the code.

app/Organizer/OrganizerServiceProvider.php
```diff
@@ -28,13 +30,22 @@ public function register(): void
...
             fn () => new CallableIriGenerator(
                 fn ($cdbid) => $container->get('config')['url'] . '/organizers/' . $cdbid
             )
         );
 
+        // use if you want to sent the user to the user-friendly GUI version
+        $container->addShared(
+            self::ORGANIZER_FRONTEND_IRI_GENERATOR,
+            fn () => new CallableIriGenerator(
+                fn ($cdbid) => $container->get('config')['baseurl'] . '/organizers/' . $cdbid
```
LucWollants, line 45: I think `baseurl` is too general if you want to point to the organizer inside the front end, maybe use something like `frontend_url`

Also I don't think the URL is correct, for example try
https://acc.uitdatabank.be/organizers/a9f9abc0-bb5e-4d03-8470-8e1a6f3a5f1a
which will fail, you need a `preview` path like
https://acc.uitdatabank.be/organizers/a9f9abc0-bb5e-4d03-8470-8e1a6f3a5f1a/preview

src/Mailer/Command/SendOwnershipRequestedMail.php
```diff
@@ -0,0 +1,25 @@
...
+use CultuurNet\UDB3\CommandHandling\AsyncCommand;
+use CultuurNet\UDB3\CommandHandling\AsyncCommandTrait;
+
+final class SendOwnershipRequestedMail implements AsyncCommand
+{
+    use AsyncCommandTrait;
+
+    private string $uuid;
+
+    public function __construct(string $uuid)
```
LucWollants, line 16: All the existing ownership commands are typed on `Uuid` so I think that will also work here?

src/Mailer/Ownership/SendMailsForOwnershipEventHandler.php
```diff
@@ -0,0 +1,52 @@
...
+        DomainMessageSpecificationInterface $domainMessageSpecification
+    ) {
+        $this->mailerCommandBus = $mailerCommandBus;
+        $this->isReplay = $domainMessageSpecification;
+    }
+
+    public function handle(DomainMessage $domainMessage): void
+    {
+        if ($this->isReplay->isSatisfiedBy($domainMessage)) {
+            // This is a replay, don't sent the email
```
LucWollants, line 35: I think that is very clear with the `isReplay` property 😊

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
```diff
@@ -0,0 +1,141 @@
...
+
+        }
+    }
+
+    public function sentMail(SendOwnershipRequestedMail $command, string $rawSubject, string $template): void
+    {
+        $uuid = new Uuid($command->getUuid());
+
+        if ($this->mailsSentRepository->isMailSent($uuid, get_class($command))) {
+            return;
```
LucWollants, line 79: Maybe useful to log this, makes troubleshooting much easier.

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
```diff
@@ -0,0 +1,141 @@
...
+        $ownerId = $ownershipItem->getOwnerId();
+        $ownerDetails = $this->identityResolver->getUserById($ownerId);
+
+        if ($ownerDetails === null) {
+            $this->logger->warning(sprintf('[ownership-mail] Could not load owner details for %s', $ownerId));
+            return;
+        }
+
+        $params = [
+            'organisationName' => $organizer['name']['nl'],
```
LucWollants, line 108: You can't be sure that the `nl` name of the organizer is provided

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
```diff
@@ -0,0 +1,141 @@
...
+            'firstName' => $ownerDetails->getUserName(),//@todo change to be the correct user
+            'organisationUrl' => $this->organizerIriGenerator->iri($ownershipItem->getItemId()),
+        ];
+
+        $subject = $this->parseSubject($rawSubject, $organizer['name']['nl']);
```
LucWollants, line 113: No need to extract the organizer name again, it's already done on line 108.

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
```diff
@@ -0,0 +1,141 @@
...
+                $this->twig->render($template . '.html.twig', $params),
+                $this->twig->render($template . '.txt.twig', $params),
+            );
+        } catch (LoaderError|RuntimeError|SyntaxError $e) {
+            $this->logger->error($e->getMessage());
+            return;
+        }
+
+        if (!$success) {
+            return;
```
LucWollants, line 129: I guess it's worth logging that the mail was not send

src/Mailer/Ownership/SendMailsForOwnershipEventHandler.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailer\Ownership;
```
LucWollants, line 5: Maybe a better namespace is `CultuurNet\UDB3\Mailer\Handlers`

tests/Mailer/Ownership/SendMailsForOwnershipEventHandlerTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->commandBus = $this->createMock(ContextDecoratedCommandBus::class);
+        $this->domainMessageSpecification = $this->createMock(DomainMessageSpecificationInterface::class);
+        $this->sendMailsForOwnership = new SendMailsForOwnershipEventHandler(
+            $this->commandBus,
+            $this->domainMessageSpecification,
+        );
+    }
+
+    /** @test
+     * This is the happy path
```
LucWollants, line 38: 😊

tests/Mailer/Ownership/SendMailsForOwnershipEventHandlerTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $event = new OwnershipRequested(
+            $id,
+            $organizerId,
+            'organizer',
+            $ownerId,
+            'google-oauth2|102486314601596809843'
+        );
+        $domainMessage = (new DomainMessageBuilder())->setRecordedOnFromDateTimeString(self::DATE_TIME_VALUE)->create($event);
+
+        $this->domainMessageSpecification
```
LucWollants, line 55: You can't use `getEvent` here?

tests/Mailer/Ownership/SendMailsForOwnershipEventHandlerTest.php
```diff
@@ -0,0 +1,102 @@
...
+            ->willReturn(true);
+
+        $this->commandBus->expects($this->never())->method('dispatch');
+
+        $this->sendMailsForOwnership->handle(
+            $domainMessage
+        );
+    }
+
+    private function getEvent(string $itemId, string $ownerId): OwnershipRequested
```
LucWollants, line 92: This is a very generic name, we have soo many events

tests/Mailer/Ownership/SendMailsForOwnershipEventHandlerTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+use CultuurNet\UDB3\Broadway\Domain\DomainMessageSpecificationInterface;
+use CultuurNet\UDB3\CommandHandling\ContextDecoratedCommandBus;
+use CultuurNet\UDB3\EventSourcing\DomainMessageBuilder;
+use CultuurNet\UDB3\Mailer\Command\SendOwnershipRequestedMail;
+use CultuurNet\UDB3\Ownership\Events\OwnershipRequested;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class SendMailsForOwnershipEventHandlerTest extends TestCase
```
LucWollants, line 15: Maybe we also need a test to make sure no mails are send for another random event then `OwnershipRequested` ?

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
```diff
@@ -0,0 +1,141 @@
...
+    private Mailer $mailer;
+    private MailsSentRepository $mailsSentRepository;
+    private LoggerInterface $logger;
+    private DocumentRepository $organizerRepository;
+    private UserIdentityResolver $identityResolver;
+    private IriGeneratorInterface $organizerIriGenerator;
+    private TwigEnvironment $twig;
+    private OwnershipSearchRepository $ownershipSearchRepository;
+
+    public function __construct(
```
LucWollants, line 40: This class contains a lot of injected interfaces. Maybe it makes sense to create a class that extracts the required parameters for the twig template. That allows for much more readable code, easier-to-write tests and easier-to-maintain code.

src/Mailer/Ownership/SendMailsForOwnershipEventHandler.php
grubolsch: You really hate comments. :-P

src/Mailer/Command/SendOwnershipRequestedMail.php
grubolsch: I did this because this command will be serialized on a message queue, and I was thought that you want to use primitive types to not use to much memory/storage on the queue system, in this case reddis.
What is your opinion on this?

tests/Mailer/Ownership/SendMailsForOwnershipEventHandlerTest.php
grubolsch: yes, we can!

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
grubolsch: Added extra logs, and also updated the unit tests to verify the log.

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
grubolsch: I hope this is better, I now look at the mainLanguage, let me know if this still needs improvement.

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
grubolsch: Added log. + unittest

app/Organizer/OrganizerServiceProvider.php
grubolsch: Changed the name of the config value, and added the /preview suffix.

src/Mailer/Ownership/SendMailsForOwnershipEventHandler.php
LucWollants: Yes 😆 ... in this case the comment is just repeating the code.

src/Mailer/Ownership/SendOwnershipMailCommandHandler.php
LucWollants: A while ago you created a class for this to solve that issue for Places `ExtractOfferName`. I guess you could create a more general class that extracts a translated property. So you create a method that gets the data and the name of the property to retrieve.

src/Mailer/Handler/Helper/OwnershipMailParamExtractor.php
```diff
@@ -0,0 +1,45 @@
...
+     * @throws DocumentDoesNotExist
+     */
+    public function fetchParams(OwnershipItem $ownershipItem, UserIdentityDetails $identityDetails): array
+    {
+        $organizerProjection = $this->organizerRepository->fetch($ownershipItem->getItemId());
+
+        $organizer = $organizerProjection->getAssocBody();
+
+        return [
+            'organisationName' => $organizer['name'][$organizer['mainLanguage']] ?? $organizer['name']['nl'],
```
LucWollants, line 40: This can still result in issues, check `ExtractOfferName`

LucWollants (approved):
Looks really good! Thanks for all the improvements and refactors!

Only one small remark about extracting the name of an organizer, see https://github.com/cultuurnet/udb3-backend/pull/2000/files#r1973882246

PR #2003, opened 2025-02-19
III-6511 authenticate api key

JonasVHG:
### Added

- `CachedApiKeyAuthenticator`: `ApiKeyAuthenticator` with Caching
- `CachedApiKeyAuthenticatorTest`: Unit tests

### Changed

- `AuthServiceProvider`: Use `CachedApiKeyAuthenticator` instead of `CultureFeedApiKeyAuthenticator `

### Fixed

- Authenticated `apiKeys` are now cached.

---

Ticket: https://jira.publiq.be/browse/III-6511

### DIFF
--- a/src/Cache/CachedApiKeyAuthenticator.php
+++ b/src/Cache/CachedApiKeyAuthenticator.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Cache;
+
+use CultuurNet\UDB3\ApiGuard\ApiKey\ApiKey;
+use CultuurNet\UDB3\ApiGuard\ApiKey\ApiKeyAuthenticationException;
+use CultuurNet\UDB3\ApiGuard\ApiKey\ApiKeyAuthenticator;
+use Symfony\Contracts\Cache\CacheInterface;
+
+final class CachedApiKeyAuthenticator implements ApiKeyAuthenticator
+{
+    private ApiKeyAuthenticator $baseApiKeyAuthenticator;
+
+    private CacheInterface $cache;
+
+    public function __construct(ApiKeyAuthenticator $baseApiKeyAuthenticator, CacheInterface $cache)
+    {
+        $this->baseApiKeyAuthenticator = $baseApiKeyAuthenticator;
+        $this->cache = $cache;
+    }
+    public function authenticate(ApiKey $apiKey): void
+    {
+        $isAuthenticated = $this->cache->get(
+            $apiKey->toString(),
+            function () use ($apiKey) {
+                try {
+                    $this->baseApiKeyAuthenticator->authenticate($apiKey);
+                    return true;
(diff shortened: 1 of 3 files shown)


src/Cache/CachedApiKeyAuthenticator.php
```diff
@@ -0,0 +1,40 @@
...
+    private ApiKeyAuthenticator $baseApiKeyAuthenticator;
+
+    private CacheInterface $cache;
+
+    public function __construct(ApiKeyAuthenticator $baseApiKeyAuthenticator, CacheInterface $cache)
+    {
+        $this->baseApiKeyAuthenticator = $baseApiKeyAuthenticator;
+        $this->cache = $cache;
+    }
+    public function authenticate(ApiKey $apiKey): void
```
LucWollants, line 23: Nit: missing new line

tests/Cache/CachedApiKeyAuthenticatorTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @test
+     */
+    public function it_can_get_a_cached_api_key(): void
+    {
+        $this->fallbackApiKeyAuthenticator->expects($this->never())
+            ->method('authenticate');
+
+        $this->expectException(ApiKeyAuthenticationException::class);
```
LucWollants, line 52: Nice to see all these tests!

But I'm a bit confused by this exception. Does that mean only a key that is not authorised is tested? If so, would it make sense to also test a key that is authorised and doesn't throw an exception?

LucWollants:
Nice work already. I left a few comments.

But now I also wonder if we also need to cache this part https://github.com/cultuurnet/udb3-backend/blob/master/src/Http/Auth/RequestAuthenticatorMiddleware.php#L166

tests/Cache/CachedApiKeyAuthenticatorTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+        $this->cachedApiKeyAuthenticator->authenticate($this->cachedApiKey);
+    }
+
+    /**
+     * @test
+     */
+    public function it_can_get_an_uncached_invalid_api_key_from_the_decoree(): void
```
LucWollants, line 60: Nit: It's not an invalid API key it is an API key that throws `ApiKeyAuthenticationException`. So maybe a better name is `unauthorized_api_key` ?

This makes me wonder if the name of the previous method is correct `it_can_get_a_cached_api_key`

tests/Cache/CachedApiKeyAuthenticatorTest.php
```diff
@@ -0,0 +1,83 @@
...
+            function () {
+                return false;
+            }
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_can_get_a_cached_api_key(): void
```
LucWollants, line 47:
```suggestion
    public function it_can_get_a_cached_unauthorized_api_key(): void
```

LucWollants (approved):
Thanks for the extra tweaks.

PR #2005, opened 2025-02-19
Change KeycloakCommand constructor parameter type

LucWollants:
### Fixed
- Fixed KeycloakCommand constructor by changing parameter type


JonasVHG (approved):

simon-debruijn (approved):
No changes have to be made to any providers of the di container?

LucWollants:
> No changes have to be made to any providers of the di container?

I checked, but I think all types are fine now.

It was a side effect from another refactor which introduced `CachedUserIdentityResolver`, and that is another interface than `KeycloakUserIdentityResolver. But they both derive from `UserIdentityResolver` which is the type to use.

PR #2010, opened 2025-02-24
III-6532 Improve Config for cache

JonasVHG:
### Changed

- `CacheServiceProvider`: Rename `cache\redis` to `cache\persistent_cache`
- `CacheServiceProvider`: Store temporary stuff in non-persistent `cache\temporary_cache`.

### Related PR

- https://github.com/cultuurnet/appconfig/pull/967
- DO NOT MERGE before the `appconfig`-PR has been deployed.

---

Ticket: https://jira.publiq.be/browse/III-6532

### DIFF
--- a/app/Cache/CacheServiceProvider.php
+++ b/app/Cache/CacheServiceProvider.php
@@ -28,7 +28,7 @@ final class CacheServiceProvider extends AbstractServiceProvider
             new CallableArgument(
                 fn ($cacheType) => new PredisCache(
                     new Client(
-                        $container->get('config')['cache']['persistent_cache'],
+                        $container->get('config')['cache']['data_cache'],
                         ['prefix' => $cacheType . '_'],
                     )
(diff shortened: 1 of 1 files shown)


app/Cache/CacheServiceProvider.php
```diff
@@ -38,7 +38,7 @@ public function register(): void
         $container->addShared(
             Client::class,
```
LucWollants, line 39: This approach is not future-proof. If we change the `cache` in the future we need `temporary_cache` and `persistent_cache` as exported services

steftrenson (approved):

LucWollants (approved):

app/Cache/CacheServiceProvider.php
LucWollants: Solved by using `app_cache` and `data_cache`

PR #2011, opened 2025-02-26
III-6538 Sent ownership mails to correct owners

grubolsch:
Added 2 strategies where to sent mails 

### Added
- Send mail to all owners (creator + current users with roles that give access to the organizer)
- Send mails to requester of ownership

### Todo
- I still need to add a lookup within the roles table, to sent mails to all users with roles that give access to the organizer.

---

Ticket: https://jira.uitdatabank.be/browse/III-6538


src/Mailer/Handler/SendOwnershipMailCommandHandler.php
```diff
@@ -57,17 +61,17 @@ public function handle($command): void
...
+                    self::TEMPLATE_OWNERSHIP_REQUESTED,
+                    new SendToRequesterOfOwnership($this->identityResolver, $this->logger)
                 );
                 break;
-
         }
     }
 
-    public function sentMail(SendOwnershipRequestedMail $command, string $rawSubject, string $template): void
+    public function processCommand(SendOwnershipRequestedMail $command, string $rawSubject, string $template, RecipientStrategy $recipientStrategy): void
```
JonasVHG, line 74: This function is only used here, maybe make it private?

tests/Mailer/Ownership/RecipientStrategy/SendToOwnersOfOrganisationTest.php
```diff
@@ -0,0 +1,86 @@
...
+use CultuurNet\UDB3\ReadModel\JsonDocument;
+use CultuurNet\UDB3\User\UserIdentityDetails;
+use CultuurNet\UDB3\User\UserIdentityResolver;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+
+class SendToOwnersOfOrganisationTest extends TestCase
+{
+    private MockObject $identityResolver;
```
JonasVHG, line 19:
```suggestion
    /**
     * @var UserIdentityResolver&MockObject
     */
    private $identityResolver;
```
Always handy to have better typehinting, same for the other `MockObjects`

src/Mailer/Ownership/RecipientStrategy/SendToRequesterOfOwnership.php
```diff
@@ -0,0 +1,35 @@
...
+    public function __construct(UserIdentityResolver $identityResolver, LoggerInterface $logger)
+    {
+        $this->identityResolver = $identityResolver;
+        $this->logger = $logger;
+    }
+
+    /** @return UserIdentityDetails[] */
+    public function getRecipients(OwnershipItem $item): array
+    {
+        $ownerDetails = $this->identityResolver->getUserById($item->getOwnerId());
```
LucWollants, line 26: The call `getOwnerId` gets the owner not the requester. The requester is the user who asks for the ownership, but he/she can ask that for another person. That is basically what happens if an owner of an organizer goes to the ownership page and adds a new owner.

src/Mailer/Ownership/RecipientStrategy/SendToOwnersOfOrganisation.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @return UserIdentityDetails[]
+     * @throws DocumentDoesNotExist
+     */
+    public function getRecipients(OwnershipItem $item): array
+    {
+        $organizer = $this->organizerRepository->fetch($item->getItemId())->getAssocBody();
+
+        //@todo loop over ALL owners of organisation
+        $ownerDetails = $this->identityResolver->getUserById($organizer['creator'] ?? '');
```
LucWollants, line 36: I would make this a separate `SendToCreatorOfOrganisation` implementation of `RecipientStrategy`. If you make a `CombinedRecipientStrategy` you can make the combination you need. This is the same approach as the `AnyVoter`

LucWollants: With this same approach you can make `Role RecipientStrategy` and `OwnershipRecipientStrategy`

grubolsch: Implemented, really liked the idea.

app/Mailer/MailerServiceProvider.php
```diff
@@ -58,19 +62,31 @@ function () use ($container): MailsSentRepository {
...
                         new FilesystemLoader(__DIR__ . '/../../src/Mailer/templates'),
                     ),
                     $this->container->get(OwnershipSearchRepository::class),
                     $this->container->get(OwnershipMailParamExtractor::class),
-                    LoggerFactory::create($this->container, LoggerName::forResqueWorker('mails')),
+                    new SendToOwnerOfOwnership(
+                        $this->container->get(UserIdentityResolver::class),
+                        $logger
+                    ),
+                    new CombinedRecipientStrategy(
```
LucWollants, line 76: Is it not weird to combine a single Strategy?

grubolsch: yeah, but in a later pr I will combined it with "ownership_search". 
But now I think of it, I could already write that class, and then just make a PR later to add the projector to fill the table.

src/Mailer/Ownership/RecipientStrategy/SendToCreatorOfOrganisation.php
```diff
@@ -0,0 +1,44 @@
...
+     * @throws DocumentDoesNotExist
+     */
+    public function getRecipients(OwnershipItem $item): array
+    {
+        $organizer = $this->organizerRepository->fetch($item->getItemId())->getAssocBody();
+
+        $ownerDetails = $this->identityResolver->getUserById($organizer['creator'] ?? '');
+
+        if ($ownerDetails === null) {
+            $this->logger->warning(sprintf('[ownership-mail] Could not load owner details for %s', (empty($organizer['creator']) ? 'unknown' : $organizer['creator'])));
```
LucWollants, line 38: Why is here a reference to mail?

src/Mailer/Ownership/RecipientStrategy/RecipientStrategy.php
```diff
@@ -0,0 +1,18 @@
...
+namespace CultuurNet\UDB3\Mailer\Ownership\RecipientStrategy;
+
+use CultuurNet\UDB3\Ownership\Repositories\OwnershipItem;
+use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
+use CultuurNet\UDB3\User\UserIdentityDetails;
+
+interface RecipientStrategy
+{
+    /**
+     * @throws DocumentDoesNotExist
```
LucWollants, line 14: I'm not sure this throw is generic, because in the `SendToOwnerOfOwnership` implementation it returns an empty array.

app/Mailer/MailerServiceProvider.php
LucWollants: So I need to wait with the review or you remove the use `CombinedRecipientStrategy` here?

grubolsch: I added a new class SendToOwnersOfOrganisation, that fetches all ownerships of an organizer that are approved and sends a mail to them. So feel free to continue the review.

src/Mailer/Handler/SendOwnershipMailCommandHandler.php
```diff
@@ -9,65 +9,67 @@
...
     private LoggerInterface $logger;
 
     public function __construct(
         Mailer $mailer,
         MailsSentRepository $mailsSentRepository,
-        UserIdentityResolver $identityResolver,
         TwigEnvironment $twig,
         OwnershipSearchRepository $ownershipSearchRepository,
         OwnershipMailParamExtractor $paramExtractor,
+        RecipientStrategy $sendToCreatorOfOrganisation,
```
LucWollants, line 46: I guess this parameter name is no longer correct because of using the combined strategy?

src/Mailer/Handler/SendOwnershipMailCommandHandler.php
```diff
@@ -83,24 +85,40 @@ public function sentMail(SendOwnershipRequestedMail $command, string $rawSubject
...
-        //@todo loop over ALL owners of organisation
-        $ownerId = $ownershipItem->getOwnerId();
-        $ownerDetails = $this->identityResolver->getUserById($ownerId);
-
-        if ($ownerDetails === null) {
-            $this->logger->warning(sprintf('[ownership-mail] Could not load owner details for %s', $ownerId));
+        try {
+            $recipients = $recipientStrategy->getRecipients($ownershipItem);
+        } catch (DocumentDoesNotExist $e) {
+            $this->logger->warning(sprintf('[ownership-mail] Could not load organizer: %s', $e->getMessage()));
```
LucWollants, line 91: This is a weird message, how can getting the recipients result in a warning when getting the organizer?

src/Mailer/Ownership/RecipientStrategy/SendToOwnerOfOwnership.php
```diff
@@ -0,0 +1,35 @@
...
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Mailer\Ownership\RecipientStrategy;
+
+use CultuurNet\UDB3\Ownership\Repositories\OwnershipItem;
+use CultuurNet\UDB3\User\UserIdentityDetails;
+use CultuurNet\UDB3\User\UserIdentityResolver;
+use Psr\Log\LoggerInterface;
+
+class SendToOwnerOfOwnership implements RecipientStrategy
```
LucWollants, line 12: I have the impression this is only used in a test?

tests/Mailer/Ownership/RecipientStrategy/CombinedRecipientStrategyTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+        $combinedStrategy = new CombinedRecipientStrategy($strategyReturns2Items, $recipientStrategyReturn1Item);
+        $recipients = $combinedStrategy->getRecipients($this->ownershipItem);
+
+        $this->assertCount(3, $recipients);
+        $this->assertSame([$this->recipient1, $this->recipient2, $this->recipient3], $recipients);
+    }
+
+    /** @test */
+    public function make_sure_nobody_gets_email_double(): void
```
LucWollants, line 64: I wonder if this is the responsibility of the `CombinedRecipientStrategy`? What if another `Strategy` would be used without the `CombinedRecipientStrategy` and returns the same email address multiple times? 

I think this can be easily solved by upgrading the associative array to some kind of smart collection `Recipients` which only stores unique user identities based on email.

tests/Mailer/Ownership/RecipientStrategy/SendToCreatorOfOrganisationTest.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->identityResolver = $this->createMock(UserIdentityResolver::class);
+        $this->logger = $this->createMock(LoggerInterface::class);
+        $this->strategy = new SendToCreatorOfOrganisation($this->identityResolver, $this->organizerRepository, $this->logger);
+    }
+
+    /** @test */
+    public function it_gets_owner_details(): void
+    {
+        $ownerDetails = new UserIdentityDetails(
+            $this->ownerId,
```
LucWollants, line 61: I don't think this is correct. You are testing the `SendToCreatorOfOrganisation` strategy, so I would expect that the creator of the organisation is used?

LucWollants: Even more for this Strategy it's very weird you need an `OwnerShipItem` because for the creator no ownership needs to be created. This seems like a code smell and leads to confusing and difficult to maintain code.

Given those extra insights I think your Strategy interface needs the item id as parameter for it's interface and not the ownership item.

LucWollants (commented):
Nice approach with the `RecipientStrategy` but now that some real use cases are implemented and tested, I think the interface should be
```
public function getRecipients(string $itemId): Recipients;
```

tests/Mailer/Ownership/RecipientStrategy/CombinedRecipientStrategyTest.php
grubolsch: I implemented a `Recipients` class that forces the uniqueness.
I did keep the unit tests because it still does a useful check.

tests/Mailer/Ownership/RecipientStrategy/SendToCreatorOfOrganisationTest.php
grubolsch: Renamed tests and now use creator.

LucWollants (approved):

PR #2012, opened 2025-02-26
III-6541 bugfix nulls

JonasVHG:
### Changed

- `CachedUserIdentityResolver`: prevent caching `Null` by throwing an catching an `InvalidArgumentException` & refactor to improve readability.
- `CachedUserIdentityResolverTest`: Add unit tests to check `Null` isn't stored in the cache.

### Fixed

- `Null` is no longer cached.

---

Ticket: https://jira.publiq.be/browse/III-6541

### DIFF
--- a/src/User/Keycloak/CachedUserIdentityResolver.php
+++ b/src/User/Keycloak/CachedUserIdentityResolver.php
@@ -75,7 +75,13 @@ final class CachedUserIdentityResolver implements UserIdentityResolver
                 $this->cache->get(
                     $this->createCacheKey($nick, 'nick'),
                     function () use ($nick) {
-                        return $this->getUserIdentityDetailsAsArray($this->userIdentityResolver->getUserByNick($nick));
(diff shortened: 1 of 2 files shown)


src/User/Keycloak/CachedUserIdentityResolver.php
```diff
@@ -25,38 +26,68 @@ public function __construct(
...
     {
-        return $this->deserializeUserIdentityDetails(
-            $this->cache->get(
-                $this->createCacheKey($userId, 'user_id'),
-                function () use ($userId) {
-                    return $this->getUserIdentityDetailsAsArray($this->userIdentityResolver->getUserById($userId));
-                }
-            )
-        );
+        try {
```
JonasVHG, line 29: This function is becoming rather complex. This is something that can/should be tackled when we are on PHP 8

LucWollants: Try to throw the `CacheException` inside the `getUserIdentityDetailsAsArray` method and use an arrow function.

LucWollants (approved):

PR #2014, opened 2025-02-26
Enable xdebug in docker

grubolsch:
### Added
Another attempt to get Xdebug inside our docker setup, with an optional docker profile, so you can quickly switch between both versions.
All instructions can be found inside the readme file.

This is how it will look inside docker engine.
Screenshot 2025-02-26 at 16 25 09


docker/php/Dockerfile
```diff
@@ -32,3 +34,15 @@ RUN sed -ri -e 's!/var/www/html!${APACHE_DOCUMENT_ROOT}!g' /etc/apache2/apache2.
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 
 CMD ["/usr/bin/supervisord", "-n", "-c", "/etc/supervisor/conf.d/supervisord.conf"]
+
+# Always copy a file, but choose# Always copy a file, but choose xdebug.ini or an empty one xdebug.ini or an empty one
```
LucWollants, line 38: This is a weird comment 😊

LucWollants:
I wonder how this will behave on Jenkins, because as far as I know Jenkins uses our docker-compose file?

docker/php/Dockerfile
grubolsch: Because 
1) COPY cannot be used inside an if, so I have to write always write the config file 
2) But if the config is there when xdebug is not installed, PHP will give an error because it tries to load a module that does not exists

So solutions was to overwrite the config file with an empty one, in case of no xdebug activated.

docker-compose.yml
```diff
@@ -26,6 +23,31 @@ services:
       rabbitmq:
         condition: service_healthy
 
+  php-cli:
+      container_name: php.uitdatabank
+      extends:
+          service: php
+      profiles:
```
LucWollants, line 30: Is this correct, should this not always run? I think this breaks the existing setup when not using the xdebug profile

bin/start-xdebug.sh
```diff
@@ -0,0 +1,2 @@
+export XDEBUG_MODE=debug XDEBUG_SESSION=1 PHP_IDE_CONFIG="serverName=host.docker.internal"
+# https://github.com/cultuurnet/udb3-backend/pull/1425/files
```
LucWollants, line 2: I guess this doesn't have much added value?

readme.md
```diff
@@ -32,6 +32,41 @@ To run the migrations, you can use the following composer script:
...
+docker-compose up
+```
+
+### Install profile with Xdebug
+
+```
+docker-compose --profile xdebug up
+```
+
+You don't have to rebuild to switch, *you can just switch between versions in your docker engine*.
```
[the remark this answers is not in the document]
grubolsch: Yes, added some extra text to make it more clear.
Personally I use it most when running unit tests, i have a seperate bash version to go into.
Also put this in the readme file.

grubolsch:
> I wonder how this will behave on Jenkins, because as far as I know Jenkins uses our docker-compose file?

Give that the base image is still called "php", nothing has changed in Jenkins.
To be sure, I will ask Kristof to verify before merging.

LucWollants:
> > I wonder how this will behave on Jenkins, because as far as I know Jenkins uses our docker-compose file?
> 
> Give that the base image is still called "php", nothing has changed in Jenkins. To be sure, I will ask Kristof to verify before merging.

Once the other acceptance test run is done you can easily start a run with parameters and change the default branch to `enable-xdebug-in-docker`

LucWollants:
> You mean you want to use xdebug on the server? That was not the goal of this pr.

No, I mean you can do the verification because committing to the main branch

grubolsch:
Ahh. I understand.  I will run the tests. :-)

grubolsch:
feature tests ran without issues https://jenkins.publiq.be/blue/organizations/jenkins/uitdatabank-acceptance-tests/detail/uitdatabank-acceptance-tests/150/pipeline/

bin/start-xdebug.sh
```diff
@@ -0,0 +1 @@
+export XDEBUG_MODE=debug XDEBUG_SESSION=1 PHP_IDE_CONFIG="serverName=host.docker.internal"
```
LucWollants, line 1: This bin folder is only used for the CLI commands, I would place this command inside the `docker` folder, which also contains a `config.sh`

LucWollants (approved):
Nice work! I have only one minor remark left on the location of the `start-xdebug.sh`

PR #2015, opened 2025-02-26
Docker MySQL volume location in is invalid, use correct location

grubolsch:
### Fixed
A recent change has moved the docker mysql volume to `/bitnami/mysql/data`.
On my machine this gives permission issues that it cannot write to this location, but `/opt/bitnami/mysql/data` does work.
I think this location is incorrect, but if somebody else could test both paths on their machine, I would be grateful.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
       - MYSQL_USER=${MYSQL_USER}
       - MYSQL_PASSWORD=${MYSQL_PASSWORD}
     volumes:
-      - 'udb3-mysql:/bitnami/mysql/data'
+      - 'udb3-mysql:/opt/bitnami/mysql/data'
       - ./docker/mysql:/docker-entrypoint-initdb.d
     ports:
       - '3306:3306'


docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
       - MYSQL_USER=${MYSQL_USER}
       - MYSQL_PASSWORD=${MYSQL_PASSWORD}
     volumes:
-      - 'udb3-mysql:/bitnami/mysql/data'
+      - 'udb3-mysql:/opt/bitnami/mysql/data'
```
JonasVHG, line 38: Maybe best to check this with @steftrenson on his machine it didn't work with the old path. And AFAIK you both use Docker instead of Orbstack.(Personally I never had an issue with either solution).

steftrenson: @JonasVHG Orbstack does not do a `down` operation, but just starts and stops the containers. The changes in the `docker-compose` file will be ignored because of this, and only take effect when you do a `down - up` operation. That's why both work. 

For me personally this does not work, when I do `make down` and then `make up` the database is cleared again. I think we need to sit together to figure this out, or change the mysql docker image to the official one from mysql, and not use a bitnami version.

grubolsch:
Can be closed, removing the volume itself fixed the problem.

PR #2016, opened 2025-02-28
III-6527 Remove feature flag blank_nodes_allowed

grubolsch:
### Removed
Remove feature flag blank_nodes_allowed, because it is now always on.
There is a related ticket to remove this option in the config file.

---

Ticket: https://jira.uitdatabank.be/browse/III-6527 
Related: https://github.com/cultuurnet/appconfig/pull/971


app/RDF/RdfServiceProvider.php
```diff
@@ -41,10 +40,6 @@ public function register(): void
         $this->container->addShared(
             RdfResourceFactory::class,
             function (): RdfResourceFactory {
-                if ($this->getContainer()->get('config')['rdf']['blank_nodes_allowed'] ?? true) {
-                    return new RdfResourceFactoryWithBlankNodes();
-                }
-
                 return new RdfResourceFactoryWithoutBlankNodes(new NodeUriGenerator(new CRC32HashGenerator()));
```
JonasVHG, line 43: Is it still relevant to keep the name `RdfResourceFactoryWithoutBlankNodes`, because `RdfResourceFactoryWithBlankNodes` no longer exist. I think the new class should be renamed to `RdfResourceFactory`

JonasVHG (approved):

app/RDF/RdfServiceProvider.php
grubolsch: Good point, I removed the interface and renamed the class.

PR #2019, opened 2025-03-12
III-6538 Send requested mails all owners

grubolsch:
### Added
- Introduced an additional projector (or at least extra handlers within a projector) to populate the ownership_search table. This change ensures that the frontend displays all owners of an organizer, rather than only those added through the ownership flow.
- The projector now listens to the ConstraintAdded, ConstraintUpdated, and ConstraintRemoved events.

### Coming in a next PR
- There is flow where the constraint is not changed but the role permission is, this will be handled by a later PR
- I will also mark ownerships that have multiple users as un-deletable, to prevent accidental mass removal of privileges.

---

Ticket: https://jira.uitdatabank.be/browse/III-6538


src/Ownership/Readmodels/OwnershipSearchProjector.php
```diff
@@ -76,4 +95,74 @@ public function applyOwnershipDeleted(OwnershipDeleted $ownershipDeleted): void
...
+    }
+
+    protected function applyConstraintUpdated(ConstraintUpdated $constraintEvent): void
+    {
+        $roleId = $constraintEvent->getUuid();
+
+        $this->connection->beginTransaction();
+
+        // First clear all existing owners for this role, because if the constraint is changed some users might no longer be valid.
+        $this->ownershipSearchRepository->deleteByRole($roleId);
```
LucWollants, line 111: I wonder if this can cause some side effects if an existing ownership gets deleted? This would happen if a role attached to a "normal" ownership gets updated.

src/Ownership/Repositories/Search/DBALOwnershipSearchRepository.php
```diff
@@ -116,6 +116,16 @@ public function searchTotal(SearchQuery $searchQuery): int
             ->fetchOne();
     }
 
+    public function doesUserForOrganisationExist(Uuid $organizerId, string $ownerId): bool
```
LucWollants, line 119: Why not use the existing `search` method?

grubolsch: I found this convenience method much easier to read, it communicates intent better than the generic search method.
I would prefer to keep the method, but I can move it to a separate class if you prefer?

src/Ownership/Readmodels/OwnershipSearchProjector.php
grubolsch: if a normal role is updated, there will be no ownership attached to it.
I don't immediately see the problem, in which direction are you thinking?

LucWollants: What is a normal role? Because an ownership also has a role, so how do you see the difference?

PR #2029, opened 2025-03-25
III-6300 Improve name ownership user

LucWollants:
### Changed
- Renamed `dev_e2e_test` to `udbtestinvoerder_ownerships` to make it more clear this user is used for ownerships only

---
Ticket: https://jira.uitdatabank.be/browse/III-6300

### DIFF
--- a/features/organizer/get-creator.feature
+++ b/features/organizer/get-creator.feature
@@ -2,7 +2,7 @@ Feature: Test getting creator of organizer
     Background:
         Given I am using the UDB3 base URL
         And I am using an UiTID v1 API key of consumer "uitdatabank"
-        And I am authorized as JWT provider v2 user "dev_e2e_test"
+        And I am authorized as JWT provider v2 user "udbtestinvoerder_ownerships"
         And I send and accept "application/json"
 
     Scenario: Getting the creator of an organizer as creator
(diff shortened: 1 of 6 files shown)


JonasVHG (approved):

features/organizer/get-creator.feature
```diff
@@ -2,7 +2,7 @@ Feature: Test getting creator of organizer
     Background:
         Given I am using the UDB3 base URL
         And I am using an UiTID v1 API key of consumer "uitdatabank"
-        And I am authorized as JWT provider v2 user "dev_e2e_test"
+        And I am authorized as JWT provider v2 user "udbtestinvoerder_ownerships"
```
simon-debruijn, line 5: Is the udbtest prefix necessary?

simon-debruijn (approved):

features/organizer/get-creator.feature
LucWollants: It's not needed, but it needs to be aligned with the appconfig.

For me it's fine to remove the `udbtest` prefix, what do you think @JonasVHG ?
If we remove it here we also need to remove it from the `udbtestinvoerder_1` and `udbtestinvoerder_2`

LucWollants: I already merged, it's easy enough to create a new pull request for removing the prefix.

PR #2039, opened 2025-04-04
III-6591 BUGFIX TypeError: current() expects parameter 1 to be array

grubolsch:
### Fixed
- TypeError: current() expects parameter 1 to be array, string given
- Provided unit test to prove fix

---

Ticket: https://jira.uitdatabank.be/browse/III-6591


LucWollants:
This will only happen in the acceptance environment because the data is not replayed and contains bugs. There is no need to fix this in the code, you could replay the specific event/place/organizer to fix it inside the data.

PR #2040, opened 2025-04-15
III-6570 ownership acceptance tests for Mails

JonasVHG:
### Changed

- `features/Steps/MailSteps.php`: Remove private functions & use `assertStringMatches()` instead of `assertStringEquals()`
- `features/Support/EmailMessage.php`: Use `Text` instead of `Html` to avoid formatting issues.
- `data/mails/*`: Use `txt` instead `html` + add ownership templates.
- `features/ownership/*`: Add check on e-mail+ add wait because e-mails are not always sent chronologically.

### Follow-up PR

- https://github.com/cultuurnet/udb3-backend/pull/2041

---

Ticket: https://jira.publiq.be/browse/III-6570


features/ownership/approve.feature
```diff
@@ -9,6 +9,7 @@ Feature: Test approving ownership
     Given I create a minimal organizer and save the "id" as "organizerId"
     And I am authorized as JWT provider v2 user "invoerder_ownerships"
     And I request ownership for "auth0|64089494e980aedd96740212" on the organizer with organizerId "%{organizerId}" and save the "id" as "ownershipId"
+    And I wait 4 seconds
```
LucWollants, line 12: As discussed in a Huddle, I think the wait needs to happen after the approval. Because in this scenario, there are 2 mails. The first mail from the request and the second mail from the approval.

LucWollants: I think the easiest is just to put the wait before the `"ownership-approved" mail has been sent from`

JonasVHG:
Best to look at this PR with the follow-up PR: https://github.com/cultuurnet/udb3-backend/pull/2041

LucWollants (approved):

PR #2047, opened 2025-04-17
Backend endpoint with list of municipalities and educational levels

grubolsch:
### Added

- Added 2 new endpoints that serve a static file 
       -  list of municipalities and educational levels
       -  list of educational levels

### Related PR
  - Documentation: https://github.com/cultuurnet/apidocs/pull/439
---

Ticket: https://jira.uitdatabank.be/browse/III-6579


app/Http/PsrRouterServiceProvider.php
```diff
@@ -596,4 +600,11 @@ private function bindMailinglist(Router $router): void
     {
         $router->put('mailing-list/{mailingListId}/', SubscribeUserToMailinglistRequestHandler::class);
     }
+
+    private function bindTaxonomyEndpoints(Router $router): void
```
LucWollants, line 604: Also, here we may need to add the `cultuurkuur` group to provide `/cultuurkuur/regions/` and that would change the name of `bindTaxonomyEndpoints` to `bindCultuurkuur`

tests/Taxonomy/GetEducationLevelsRequestHandlerTest.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $request = (new Psr7RequestBuilder())
+            ->withUriFromString('/education-levels/')
+            ->build('GET');
+
+        $response = (new GetEducationLevelsRequestHandler())->handle($request);
+
+        $this->assertSame('application/json', $response->getHeaderLine('Content-Type'));
+
+        $json = json_decode((string)$response->getBody(), true, 512, JSON_THROW_ON_ERROR);
```
LucWollants, line 22: We have the `Json` class for this

LucWollants:
I think we can make these endpoints public, so no need for authentication. This means a call to `addPublicRoute` is needed.

grubolsch:
> I think we can make these endpoints public, so no need for authentication. This means a call to `addPublicRoute` is needed inside the `AuthServiceProvider` provider.

Done, thank you for the really clear instructions! :-)

src/Cultuurkuur/data/regions.json
```diff
@@ -0,0 +1,5710 @@
...
+        "label": "Brussels Hoofdstedelijk Gewest",
+        "value": "nis-01000",
+        "children": [
+            {
+                "label": "Regio Brussel",
+                "value": "reg-brussel",
+                "children": [
+                    {
+                        "label": "Anderlecht + deelgemeenten",
+                        "value": "nis-21001"
```
LucWollants, line 12: This `nis-21001` is wrong and should be `cultuurkuur_werkingsregio_nis-21001` otherwise the wrong label will be added.

LucWollants (changes requested):
I think the region list is started from the old list and not the one with the 2025 mergers included, I updated the linked ticket with the recent list from Slack.

grubolsch:
I now used the 2025 list, I manually checked with the list of merged cities and verified they are all oké.
Also, the format for both files has changed based on our Slack talk.

label -> name
value -> label

Name is now an object with property "nl"

LucWollants (approved):
Nice work, thanks for the extra tweaks.

I wonder if it's worth to write a very basic acceptance test to make sure these endpoints work without authentication? Because that is something not tested with the unit test.

PR #2048, opened 2025-05-09
III-6618 ownership cli

JonasVHG:
### Added

- `AddOwnershipCommand`: CLI Command to add `ownerships`.
- `DomainMessageHasMailsDisabled`: Add new `DomainMessageSpecificationInterface`
- `DisableMailsMiddleware`: MiddleWare to disable sending mails, when changing `Ownership`
- `DisableMailsMiddlewareTest`: Unit test for `DisableMailsMiddleware`

### Changed

- `ConsoleServiceProvider`: Register new Command `AddOwnershipCommand`.
- `EventBusServiceProvider`: register new MiddleWare `DisableMailsMiddleware`
- `SendMailsForOwnershipEventHandler`: Check if mails are disabled in MiddleWare.
- `SendMailsForOwnershipEventHandlerTest`: Add unit test to see if mails disabled in MiddleWare work
- `MailerServiceProvider`: Updated `constructor` of `SendMailsForOwnershipEventHandler`
 
---

Ticket: https://jira.publiq.be/browse/III-6618

### DIFF
--- a/app/Console/Command/AddOwnershipCommand.php
+++ b/app/Console/Command/AddOwnershipCommand.php
@@ -35,7 +35,7 @@ final class AddOwnershipCommand extends AbstractCommand
 
     private const USER = 'user';
 
-    private const SEND_MAILS = 'send-mails';
+    private const ENABLE_MAILS = 'enable-mails';
 
     private UuidFactory $uuidFactory;
 
@@ -78,10 +78,10 @@ final class AddOwnershipCommand extends AbstractCommand
                 InputOption::VALUE_REQUIRED,
                 'The id or email of a user give ownership to.'
             )->AddOption(
-                self::SEND_MAILS,
+                self::ENABLE_MAILS,
                 null,
                 InputOption::VALUE_NONE,
-                'Option to send mails during the CLI Command, default is not sending mails.'
+                'Option to enable mailing during the CLI Command, default is not sending mails.'
             );
     }
 
@@ -90,7 +90,7 @@ final class AddOwnershipCommand extends AbstractCommand
         $itemType = $input->getArgument(self::ITEM_TYPE);
         $itemId = $input->getArgument(self::ITEM_ID);
         $user = $input->getArgument(self::USER);
-        $sendMails = $input->getOption(self::SEND_MAILS);
+        $sendMails = $input->getOption(self::ENABLE_MAILS);
 
         if (!$this->itemExists($itemId)) {
             $output->writeln('Organizer does not exist.');
--- a/app/Console/Command/AddOwnershipCommand.php
+++ b/app/Console/Command/AddOwnershipCommand.php
@@ -27,7 +27,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
-final class OwnershipCommand extends AbstractCommand
+final class AddOwnershipCommand extends AbstractCommand
 {
     private const ITEM_TYPE = 'item-type';
 
--- a/app/Console/Command/OwnershipCommand.php
+++ b/app/Console/Command/OwnershipCommand.php
@@ -133,14 +133,11 @@ final class OwnershipCommand extends AbstractCommand
     {
         try {
             $userEmail = new EmailAddress($user);
-            $user = $this->userIdentityResolver->getUserByEmail($userEmail);
-            if (!$user) {
-                return null;
-            }
-            return new UserId($user->getUserId());
+            $userDetails = $this->userIdentityResolver->getUserByEmail($userEmail);
         } catch (InvalidEmailAddress $exception) {
-            return new UserId($user);
+            $userDetails = $this->userIdentityResolver->getUserById($user);
         }
+        return $userDetails !== null ? new UserId($userDetails->getUserId()) : null;
     }
 
     private function itemExists(string $itemId): bool
--- a/app/Console/Command/OwnershipCommand.php
+++ b/app/Console/Command/OwnershipCommand.php
@@ -1,0 +1,130 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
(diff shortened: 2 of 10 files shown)


app/Console/Command/OwnershipCommand.php
```diff
@@ -0,0 +1,155 @@
...
+    {
+        try {
+            $userEmail = new EmailAddress($user);
+            $user = $this->userIdentityResolver->getUserByEmail($userEmail);
+            if (!$user) {
+                return null;
+            }
+            return new UserId($user->getUserId());
+        } catch (InvalidEmailAddress $exception) {
+            return new UserId($user);
```
LucWollants, line 142: This fallback can be tricky. For example, if someone made a mistake and provided `john.doe@gmail` as a parameter. This will result in an `InvalidEmailAddress` exception, and the value will be cast to `UserId. This will work because it is a string and not empty. Maybe we also need to call `getUserById` to make sure the user is found?

I think the HTTP request handler suffers from the same issue, there is no validation to match a passed user ID to an existing user. So the HTTP layer accepts random user IDs. Maybe we need an extra acceptance test for this?

app/Console/Command/OwnershipCommand.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        $userId = $this->getUserId($user);
+        if ($userId === null) {
+            $output->writeln('No user found for ' . $user);
+            return self::FAILURE;
+        }
+
+        if ($this->ownershipExist($itemId, $userId)) {
+            $output->writeln('Ownership already exists.');
```
LucWollants, line 96: Maybe we also need to display the UUID of the ownership? That makes it easier to search for the existing ownership.

app/Console/Command/OwnershipCommand.php
JonasVHG: Good point, I've fixed this for the CLI Command & made a ticket for the HTTP endpoint => https://jira.publiq.be/browse/III-6622

LucWollants (changes requested):
I think we also need to be very careful when running this command. If we run this command in production all existing owners and all new owners will get an email, that could lead to several mails for one user.

LucWollants:
> > I think we also need to be very careful when running this command. If we run this command in production all existing owners and all new owners will get an email, that could lead to several mails for one user.
> 
> At present it is only required for CultuurKuur, who can be informerd, maybe something to pick up with Eva.

Now we remember this will send a lot of emails and we can coordinate this. But if we run this in the future in batch, it's very likely we won't think about it. So I would try to find a solution to run this with emails disabled by default and add an option to enable emails.

app/Console/Command/AddOwnershipCommand.php
```diff
@@ -0,0 +1,174 @@
...
+                'Option to send mails during the CLI Command, default is not sending mails.'
+            );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $itemType = $input->getArgument(self::ITEM_TYPE);
+        $itemId = $input->getArgument(self::ITEM_ID);
+        $user = $input->getArgument(self::USER);
+        $sendMails = $input->getOption(self::SEND_MAILS);
```
LucWollants, line 93: Should this have a better default value?

src/Mailer/Handler/SendMailsForOwnershipEventHandler.php
```diff
@@ -37,6 +37,10 @@ public function handle(DomainMessage $domainMessage): void
             return;
         }
 
+        if ($domainMessage->getMetadata()->get('disable_mails')) {
```
LucWollants, line 40: I see we already have some code on line 36 to check for a replay, maybe we need to follow the same pattern and implement a new `DomainMessageSpecificationInterface`?

JonasVHG: Done.

src/Mailer/Handler/SendMailsForOwnershipEventHandler.php
```diff
@@ -21,20 +21,26 @@ final class SendMailsForOwnershipEventHandler implements EventListener
...
+
+    /**
+     * @var DomainMessageSpecificationInterface[]
+     */
+    private array $domainMessageSpecifications;
 
     public function __construct(
         ContextDecoratedCommandBus $mailerCommandBus,
-        DomainMessageSpecificationInterface $domainMessageSpecification
+        DomainMessageSpecificationInterface ...$domainMessageSpecification
```
LucWollants, line 32: I'm not sure about this change, because now you could use the `SendMailsForOwnershipEventHandler` without passing the extra `DomainMessageHasMailsDisabled`. This could potentially lead to still sending emails.

tests/Mailer/Handler/SendMailsForOwnershipEventHandlerTest.php
```diff
@@ -125,7 +163,7 @@ public function it_does_not_dispatch_with_an_invalid_event(): void
                 new Description('description-456')
             ));
 
-        $this->domainMessageSpecification
+        $this->domainMessageSpecificationForReplays
```
LucWollants, line 166: I know that this test is not changed, but I think that the dispatch is not called because the replay mock returns `true`. So it doesn't really test the `switch` statement inside the `handle` method.

src/Broadway/Domain/DomainMessageHasMailsDisabled.php
```diff
@@ -0,0 +1,18 @@
...
+
+use Broadway\Domain\DomainMessage;
+
+class DomainMessageHasMailsDisabled implements DomainMessageSpecificationInterface
+{
+    public const METADATA_MAILS_DISABLED_KEY = 'disable_mails';
+
+    public function isSatisfiedBy(DomainMessage $domainMessage): bool
+    {
+        $metadata = $domainMessage->getMetadata()->serialize();
```
LucWollants, line 15: Nit: I wonder if we could just use the `get` method on the `Metadata` class here?

LucWollants (approved):

PR #2051, opened 2025-05-13
III-6616 mm organizers

JonasVHG:
### Fixed

---

### DIFF
--- a/app/Console/Command/UpdatePlaceLocality.php
+++ b/app/Console/Command/UpdatePlaceLocality.php
@@ -1,0 +1,155 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use Broadway\CommandHandling\CommandBus;
+use CultuurNet\UDB3\Json;
+use CultuurNet\UDB3\Model\ValueObject\Geography\Address;
+use CultuurNet\UDB3\Model\ValueObject\Geography\CountryCode;
+use CultuurNet\UDB3\Model\ValueObject\Geography\Locality;
+use CultuurNet\UDB3\Model\ValueObject\Geography\PostalCode;
+use CultuurNet\UDB3\Model\ValueObject\Geography\Street;
+use CultuurNet\UDB3\Model\ValueObject\Translation\Language;
+use CultuurNet\UDB3\Place\Commands\UpdateAddress;
+use CultuurNet\UDB3\ReadModel\DocumentDoesNotExist;
+use CultuurNet\UDB3\ReadModel\DocumentRepository;
+use CultuurNet\UDB3\Search\ResultsGenerator;
+use CultuurNet\UDB3\Search\ResultsGeneratorInterface;
+use CultuurNet\UDB3\Search\SearchServiceInterface;
+use CultuurNet\UDB3\Search\Sorting;
+use Exception;
+use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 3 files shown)


app/Console/ConsoleServiceProvider.php
```diff
@@ -439,13 +441,22 @@ function () use ($container) {
...
                 $container->get('place_jsonld_repository')
             )
         );
 
+        $container->addShared(
+            'console.organizer:update-locality',
+            fn () => new UpdateOrganizerLocality(
+                $container->get('event_command_bus'),
+                $container->get(OrganizersSapi3SearchService::class),
+                $container->get('place_jsonld_repository')
```
LucWollants, line 456: This seems to be the wrong repository. Were you able to test this new CLI command?

LucWollants (commented):
I'm not sure if a straight copy/paste and replacing place with organizer was the best approach. Would it be meaningful to introduce a type argument that accepts either place or organizer and take that into account for creating the feedback messages?

app/Console/Command/UpdatePlaceLocality.php
```diff
@@ -113,26 +144,34 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
...
-                    ),
-                    new Language('nl')
+                $address = new Address(
+                    new Street($jsonLd['address']['nl']['streetAddress']),
+                    new PostalCode($jsonLd['address']['nl']['postalCode']),
+                    new Locality($newLocality),
+                    new CountryCode($jsonLd['address']['nl']['addressCountry'])
                 );
+
+                $command = $itemType->sameAs(ItemType::place()) ? new UpdatePlaceAddress(
```
LucWollants: To increase readability, you can make this a method. Once we are on PHP 8.x, you can even return a union, now the type hint of the return type will need to be comment.

app/Console/Command/UpdatePlaceLocality.php
```diff
@@ -76,33 +98,42 @@ public function configure(): void
...
             return self::SUCCESS;
         }
 
-
         if (!$this->askConfirmation($input, $output, $count)) {
             return self::SUCCESS;
         }
 
-        foreach ($this->searchResultsGenerator->search($query) as $place) {
+        $searchResultsGenerator = $itemType->sameAs(ItemType::place()) ?
```
LucWollants: Just make a helper method to get the correct `searchResultsGenerator`. You can do the same for `documentRepository`

app/Console/Command/UpdatePlaceLocality.php
```diff
@@ -76,33 +98,42 @@ public function configure(): void
...
+        if ($itemType->sameAs($itemType::event())) {
+            $output->writeln('<error>Only itemType place or organizer are accepted.</error>');
+            return self::FAILURE;
+        }
         $query = $input->getArgument(self::QUERY);
         $newLocality = $input->getArgument(self::NEW_LOCALITY);
 
         if ($query === null || $newLocality === null) {
-            $output->writeln('<error>Missing argument, the correct syntax is: place:update-locality "sapi3_query" "new_locality"</error>');
+            $output->writeln('<error>Missing argument, the correct syntax is: place:update-locality "item_type" "sapi3_query" "new_locality"</error>');
```
LucWollants: I don't think the names in the error method match with the actual values of the constants?

LucWollants (approved):
Looks good. Thanks for the extra tweaks!

PR #2052, opened 2025-05-14
III-6634 bulk update labels

LucWollants:
### Added
- Added new endpoint `PUT '/{offerType:events|places}/{offerId}/labels/'` to update all labels on an event or place

### Note
- For now the API docs and the JSON schema validation is still missing

---

Ticket: https://jira.uitdatabank.be/browse/III-6634


JonasVHG (approved):
There is no test for the unhappy path(When a user does not have permission on certain private labels), but since this is an internal endpoint & the Command is already thoroughly in other parts of the code, this is just minor nit.

LucWollants:
> There is no test for the unhappy path(When a user does not have permission on certain private labels), but since this is an internal endpoint & the Command is already thoroughly in other parts of the code, this is just minor nit.

No, it's a valid remark. I will add an extra feature test for this.

JonasVHG (approved):

PR #2054, opened 2025-05-16
III-6579 Updated cultuurkuur endpoints

grubolsch:
### Changes
- Make sure the right municipalities are shown in the regions (now som regions have the same municipalities ex: Antwerpen)
- Add the province label for the municipalities, on province level. So this is the second province label in the excel
- Remove region label (ex. ‘reg-brussel’) because it is never used
- Remove ‘+ deelgemeenten’ on all labels
- For educationlevels: lowest level labels (level 3) are changed. Check if this is the case (see excelscheet with changes).

---

Ticket: https://jira.uitdatabank.be/browse/III-6579

### DIFF
--- a/src/Cultuurkuur/data/regions.json
+++ b/src/Cultuurkuur/data/regions.json
@@ -4,124 +4,124 @@
             "nl": "Brussels Hoofdstedelijk Gewest"
         },
         "label": "cultuurkuur_werkingsregio_provincie_nis-01000",
+        "extraLabel": "cultuurkuur_werkingsregio_nis-01000",
         "children": [
             {
                 "name": {
                     "nl": "Regio Brussel"
                 },
-                "label": "reg-brussel",
                 "children": [
                     {
                         "name": {
-                            "nl": "Anderlecht + deelgemeenten"
+                            "nl": "Anderlecht"
                         },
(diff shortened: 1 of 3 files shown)


src/Cultuurkuur/data/regions.json
```diff
@@ -4,124 +4,124 @@
             "nl": "Brussels Hoofdstedelijk Gewest"
         },
         "label": "cultuurkuur_werkingsregio_provincie_nis-01000",
+        "extraLabel": "cultuurkuur_werkingsregio_nis-01000",
```
JonasVHG, line 7: Looks fine, but maybe also check with @vhande because, this will almost exclusivly be used by the frontEnd.

JonasVHG (approved):
Looks fine to me, but see my remark.

LucWollants:
How are the JSON files created now? Initially, we thought it would be a one-time process, but we have to make changes more often. So we should at least document the process or automate it.

grubolsch:
> How are the JSON files created now? Initially, we thought it would be a one-time process, but we have to make changes more often. So we should at least document the process or automate it.

The regions I did what chatGPT and then manual cleanup, the education file is not that large so I did it manually.

PR #2058, opened 2025-05-28
III-6634 Create labelsReplaced event

grubolsch:
### Added
- Added a `labelsReplaced` event replacing (in some cases)`labelsImported`

### Todo
In src/Label/LabelEventRelationTypeResolver.php::getRelationTypeForReplaceLabel() an extra if is needed to support Organizer LabelsReplaced, Luc is working on this.

---

Ticket: https://jira.uitdatabank.be/browse/III-6634

### DIFF
--- a/src/Label/ReadModels/Relations/Projector.php
+++ b/src/Label/ReadModels/Relations/Projector.php
@@ -83,6 +83,22 @@ class Projector extends AbstractProjector
         }
     }
 
+    public function applyLabelsReplaced(LabelsImportedEventInterface $labelsReplaced, Metadata $metadata): void
+    {
+        foreach ($labelsReplaced->getAllLabelNames() as $labelName) {
+            try {
+                $this->writeRepository->save(
+                    $labelName,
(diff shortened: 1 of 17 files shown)


src/Label/ReadModels/Relations/Projector.php
```diff
@@ -83,6 +83,22 @@ public function applyLabelsImported(LabelsImportedEventInterface $labelsImported
...
 
+    public function applyLabelsReplaced(LabelsImportedEventInterface $labelsReplaced, Metadata $metadata): void
+    {
+        foreach ($labelsReplaced->getAllLabelNames() as $labelName) {
+            try {
+                $this->writeRepository->save(
+                    $labelName,
+                    $this->offerTypeResolver->getRelationTypeForReplaceLabel($labelsReplaced),
+                    $labelsReplaced->getItemId(),
+                    true
```
LucWollants, line 94: I don't think we need the value of `true` here for the replace. In that case, we don't even need to implement this method because the `LabelAdded` and `LabelRemoved` will add the labels to the projection.

PR #2075, opened 2025-07-07
III-6702 current user

JonasVHG:
### Fixed

---

Ticket: https://jira.publiq.be/browse/III-6702


JonasVHG:
After a slack discussion we will not implement it this way, but make some backend changes to minimize frontend changes => https://github.com/cultuurnet/udb3-backend/pull/2076

PR #2077, opened 2025-07-10
III-6705 Replace Duplicate Place with Canonical Place on Create/Update

JonasVHG:
### Changed

- `EventEditingServiceProvider`: Updated `constructor` parameter from `PlaceRepository` to `DuplicatePlaceRepository` 
- `ConsoleServiceProvider`, `PlaceJSONLDServiceProvider` & `PlaceServiceProvider`: Use FQN-name instead of `duplicate_place_repository`.
- `RelocateEventToCanonicalPlace`: Refactor logic &. use `DuplicatePlaceRepository` instead of `CanonicalPlaceRepository`
- `RelocateEventToCanonicalPlaceTest`: Refactored tests

### Fixed

- When one creates or updates an `Event` with a `duplicate` `Place`, it should automatically be updated with the `canonical` `Place`.

### Deleted

- `CanonicalPlaceRepository` & `CanonicalPlaceRepositoryTest`

---

Ticket: https://jira.publiq.be/browse/III-6705


app/Console/ConsoleServiceProvider.php
```diff
@@ -265,7 +266,7 @@ function () use ($container) {
             'console.place:process-duplicates',
             fn () => new ProcessDuplicatePlaces(
                 $container->get('event_command_bus'),
-                $container->get('duplicate_place_repository'),
+                $container->get(DuplicatePlaceRepository::class),
```
grubolsch, line 269: Nice improvement! :-)

src/Place/CanonicalPlaceRepository.php
```diff
@@ -4,24 +4,20 @@
 
 namespace CultuurNet\UDB3\Place;
 
-class CanonicalPlaceRepository
+use CultuurNet\UDB3\Place\Canonical\DuplicatePlaceRepository;
+
+final class CanonicalPlaceRepository
```
grubolsch, line 9: I wonder if we should not remove this entire class and just add **DuplicatePlaceRepository** as a dependency to **RelocateEventToCanonicalPlace**.
This class is now just a single line wrapper for **DuplicatePlaceRepository**, and only used in 1 place.

JonasVHG: Indeed, will remove it.

PR #2081, opened 2025-08-18
III-6300 replace v1 tokens with v2 users

JonasVHG:
### Changed

- `features/*`: Replaced most uses of v1 tokens, userIds with v2 users.

### Note

- `features/{event|place}/auto-approve.feature`: Still use the `v1`-token, because of a bug => https://jira.publiq.be/browse/III-6772 We should also decide if we want to keep this feature, because it does not seem to be used by anyone at the moment.

### Related PR

- https://github.com/cultuurnet/appconfig/pull/1145

---

Ticket: https://jira.publiq.be/browse/III-6300

### DIFF
--- a/features/search/auth.feature
+++ b/features/search/auth.feature
@@ -51,9 +51,3 @@ Feature: Test the Search API v3 authentication
     And I am using the Search API v3 base URL
     When I send a GET request to "/events"
     Then the response status should be "403"
-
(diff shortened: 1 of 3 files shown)


bertramakers:
A bit late but also want to add that this is a very good cleanup, especially with the extra boyscouting of removing the now obsolete "v2" part from the steps that are kept 👏

PR #2085, opened 2025-08-25
III-6772 fix for v2 JWT Provider tokens in DomainMessage

JonasVHG:
### Changed

- `AuthorizationSteps`: use `id_token` instead of `access_token` in `iAmAuthorizedAsJwtProviderV2User()`
- `auth.feature`:  Remove faulty test.
- `RequestAuthenticatorMiddleware`: Add check for `typ` of `token` when deciding to `authenticateApiKey()`
- `JsonWebToken`: Remove check for `auth0`-tokens without `azp`
- `JsonWebTokenTest` & `UitIdV2JwtValidatorTest`: Remove tests for `auth0`-tokens without `azp`

### Fixed

- `DomainMessages` should now contain an `apiKey`, when using a `JWT Provider v2-token`.

---

Ticket: https://jira.publiq.be/browse/III-6772

### DIFF
--- a/src/Http/Auth/RequestAuthenticatorMiddleware.php
+++ b/src/Http/Auth/RequestAuthenticatorMiddleware.php
@@ -88,9 +88,7 @@ final class RequestAuthenticatorMiddleware implements MiddlewareInterface
         $this->authenticateToken($request);
 
         // Requests that use a token from the JWT provider (v1 or v2) require an API key from UiTID v1.
-        // Requests that use a token that they got directly from Auth0 do not require an API key.
-        // The difference can be checked by checking if the token has a client id, which is always missing in tokens
-        // from the JWT providers.
+        // Requests that use a token that they got from a clientId do not require an API key.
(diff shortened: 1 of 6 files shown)


src/Http/Auth/RequestAuthenticatorMiddleware.php
```diff
@@ -91,7 +91,7 @@ private function authenticate(ServerRequestInterface $request): void
         // Requests that use a token that they got directly from Auth0 do not require an API key.
         // The difference can be checked by checking if the token has a client id, which is always missing in tokens
         // from the JWT providers.
-        if ($this->token->getClientId() === null) {
+        if ($this->token->getClientId() === null || $this->token->getType() === JsonWebToken::UIT_ID_V2_JWT_PROVIDER_TOKEN) {
```
bertramakers, line 94: I like how you made the check more readable here by checking the token type. Maybe you can also do that for the first check (empty client id) by checking for V1 JWT provider type instead?

bertramakers (approved):

PR #2089, opened 2025-09-04
III-6777 Change calculation canonical place protect uitpas place

grubolsch:
### Changed
Change the canonical place algoritme in the following way:

Check if there is an UiTPAS location, in which case, pick it as the canonical place.
If there are more then 1 UiTPAS location, throw an error (like we already do for Museumpass)
If there is an UiTPAS location AND an Museumpass location (and they are not the same place), also throw an error.

---

Ticket: https://jira.uitdatabank.be/browse/III-6777

### DIFF
--- a/src/Label/ReadModels/Relations/Repository/Doctrine/DBALReadRepository.php
+++ b/src/Label/ReadModels/Relations/Repository/Doctrine/DBALReadRepository.php
@@ -29,20 +29,20 @@ class DBALReadRepository extends AbstractDBALRepository implements ReadRepositor
         }
     }
 
-    public function getLabelRelationsForTypes(array $labelNames, RelationType $relationType): array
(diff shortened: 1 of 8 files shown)


src/Label/ReadModels/Relations/Repository/Doctrine/DBALReadRepository.php
```diff
@@ -29,6 +29,24 @@ public function getLabelRelations(string $labelName)
         }
     }
 
+    public function getLabelRelationsForTypes(array $labelNames, RelationType $relationType): array
```
LucWollants, line 32: This name is not really correct, it is just one type, but multiple labels

LucWollants (commented):
I was to eager with the approval, some tests are broken now

PR #2104, opened 2025-10-03
Fix playhead mismatch by using correct event namespace

LucWollants:
### Fixed
- Fixed playhead mismatch on `LabelsReplaced` by using correct event namespace


bertramakers:
Doesn't this mean the tests added in the linked commit / PR are too flexible and should also be updated, since they were green while the functionality was bugged?

LucWollants:
> * e6eb4fa#diff-811dcc72a556315653ab31422bbdb292a8fa00357307db395395710141a5629dR87

The tests in the linked pull request use the infamous `Item` aggregate, so they did not use the wrongly bootstrapped `Event` aggregate.

That being said, we should get rid of that `Item` aggregate.

PR #2219, opened 2026-04-09
III-7119 Analyse Instelling API

lucwollants:
### Added
- Uploaded "instelling" YAML describing the API
- Uploaded initial analysis (with background information, proposed flow and open items)

---

Ticket: https://jira.uitdatabank.be/browse/III-7119

### DIFF
--- a/docs/scholen_api/analysis.md
+++ b/docs/scholen_api/analysis.md
@@ -1,0 +1,336 @@
+# Schools API Import - Analysis
+
+## Executive Summary
+
+We want to import all Flemish schools from the government's Instelling API into UDB3 as places, and keep them in sync over time. The data team handles the entire import flow: fetching schools, preprocessing, duplicate detection via BigQuery, and calling the UDB3 API to create/update places. UDB3's role is to expose its existing Place API.
+
+The data team is the right owner of this import because they have the existing infrastructure and expertise for duplicate detection (BigQuery matching, Google geolocation API, cluster analysis) and the data knowledge needed to clean up and normalize school data before it enters UDB3.
+
+### Terminology
+
+The government API uses the term "instelling" (institution), which covers schools but also other structures like CLBs, internaten, and scholengemeenschappen. In this document we use **"school"** when talking about what we import and manage. We use **"instelling"** only when referring to the API's field names (e.g. `instelling_naam_volledig`) or endpoints (e.g. `GET /instelling`).
+
+---
+
+## 1. Architecture: Two Separate Systems
+
+Both systems are internal to publiq, but they operate independently with a clear boundary:
+
+**Data team (Python/BigQuery)**:
+- Fetches schools from the Instelling API
+- Preprocesses and transforms the data
+- Runs duplicate/cluster analysis against existing UDB3 places
+- Calls the UDB3 API over HTTP to create and update places
+- Tracks which schools have been imported (see chapter 7)
+
+**UDB3**:
+- Exposes the `POST /places/` and `PUT /places/{placeId}/` API endpoints
+- Handles duplicate prevention at creation time (see chapter 3)
+- Stores places as event-sourced aggregates
+- Keeps track of which places are imported schools (see chapter 7 for options)
+
+---
+
+## 2. Cluster Processing
+
+UDB3 has an existing system for managing duplicate places:
+
+1. **External detection**: The data team analyzes places in BigQuery, using name matching and the Google geolocation API. When locations match, places are grouped into a cluster.
+2. **Intermediate table**: Cluster data is uploaded to the `duplicate_places_import` table (columns: `cluster_id`, `place_uuid`).
+3. **Import sync**: The command `place:duplicate-places:import` syncs the intermediate table to `duplicate_places`.
+4. **Canonical selection**: The command `place:process-duplicates` determines which place in each cluster is the "canonical" (primary) one, using this priority:
+   - A place with a configured "canonical label" (highest priority)
+   - The place with the most events
+   - The oldest place (fallback)
+5. **Event relocation**: All events at duplicate (non-canonical) places are automatically moved to the canonical place.
+
+This is a **manual, multi-step workflow**. The data team exports UDB3 places, processes them in BigQuery, and uploads results to the intermediate table. Without those manual steps, the cluster commands have nothing to process.
+
+---
+
+## 3. Duplicate Prevention at Runtime
+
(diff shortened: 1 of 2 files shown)


grubolsch:
Very good start, and I am not sure how usefull feedback still is given that the UWP might do the import.
I am left with one big question, the initial import is probably only the start of the real problem. 
The tricky part is the ongoing sync, and that’s still pretty vague in the doc (frequency = TBD, closure handling = TBD).
Before we start building, I think we need clearer answers on the update strategy:

What’s the source of truth when UDB3 data and Instelling API data don’t match? Do we first want to do some lookup to see if some schools already exists in SAPI3? Do we try to enrich those places and mark them as "synced" or just ignore them completly?
What happens if a curator or school admin edits a place manually, after they gained access right, which they will not have as a defauly because no uitid: will the next sync just overwrite that? 

Or is this a one time thing?

docs/scholen_api/analysis.md
```diff
@@ -0,0 +1,336 @@
...
+- The mapping is part of the normal `POST`/`PUT` payload - no extra infrastructure
+- Any UDB3 API consumer can see which places came from the school import
+
+**Cons**:
+- Listing all imported schools requires a search query, not a simple table scan
+- No place to track sync metadata (`last_synced_at`, `instelling_datum_laatste_wijziging`)
+
+### Option B: `school_api_imported` table in UDB3
+
+A dedicated table in the UDB3 database:
```
grubolsch, line 226: I like Option A more, as long as we make sure the label cannot be removed by normal users.

- The data team already uses BigQuery as their source of truth. Tracking instelling_nummer → place_uuid and sync metadata (like last_synced_at) fits naturally there, no real need to push that into UDB3. Option B’s main benefit was that metadata table, but honestly that belongs with the data team anyway.
- No extra table or new endpoint needed, less code, less bugs :-) , no shared schema to align on across teams.

The hidden label already does exactly what we need: “this place came from source X with ID Y”. Option B would basically be reinventing that with a separate table.

One thing Option A still doesn’t solve (but is separate from the label choice):

The overwrite risk from chapter 10. If a curator or school admin edits a place and the next sync just PUTs the full payload, those changes get wiped. Is it a broader question who is the "source of truth".

docs/scholen_api/analysis.md
```diff
@@ -0,0 +1,336 @@
...
+
+---
+
+## 8. Ownership
+
+Schools need to be able to modify their own data in UDB3. However, ownership is tied to registered **UiTID** users.
+
+The Instelling API provides director names but no email or UiTID, so ownership cannot be assigned automatically during import.
+
+### Possible approaches
```
grubolsch, line 256: I would use a combination of these approaches, all are good ideas. ;-) 

1) First we lookup with the email if there is an uitid for it, if yes, use it as creator (can we do that with the current entry API? I know there is a CLI job for changing ownership) -> if yes, send an email to the user as well, maybe group them in case some people have multiple schools? (like kinderschool, highschool, ...)
2) If not an uitid, send an mail for telling them their school got imported. A cool feature (but optional) would be that if they click a certain link in the mail, the automaticly become an owner (using ownership system), but this probably would require extra code

docs/scholen_api/analysis.md
```diff
@@ -0,0 +1,336 @@
...
+
+## 11. Open Items
+
+### Fields and mapping
+- What do the `filter_instelling_status_erkenning` values (`I`, `S`, `W`, `X`) mean? Only `E` (erkend) is clear. Need to verify via API or documentation.
+- Which `instelling_niveau` and `instelling_type` values to include? Enum values are in `instelling.yaml`.
+- Do we need **translations**? The Instelling API only provides Dutch, but some Brussels schools may need French.
+- Fall back to `instelling_naam` when `instelling_naam_volledig` exceeds 90 characters (Place name max)?
+- How to format `instelling_busnummer`? E.g. `"Kerkstraat 10 bus 3"` or `"Kerkstraat 10/3"`?
+- `instelling_gemeente` is uppercase (e.g. `"MECHELEN"`). Title-case it, or does UDB3 normalize?
```
grubolsch, line 315: Keep it as UDB3 normalized as possible I would say, will also work best with the duplicate places script.

docs/scholen_api/analysis.md
```diff
@@ -0,0 +1,336 @@
...
+
+---
+
+## 11. Open Items
+
+### Fields and mapping
+- What do the `filter_instelling_status_erkenning` values (`I`, `S`, `W`, `X`) mean? Only `E` (erkend) is clear. Need to verify via API or documentation.
+- Which `instelling_niveau` and `instelling_type` values to include? Enum values are in `instelling.yaml`.
+- Do we need **translations**? The Instelling API only provides Dutch, but some Brussels schools may need French.
+- Fall back to `instelling_naam` when `instelling_naam_volledig` exceeds 90 characters (Place name max)?
```
grubolsch, line 313: Mark for manual fix? I would be careful with auto truncate, first see how many edge case there are?

docs/scholen_api/analysis.md
```diff
@@ -0,0 +1,336 @@
...
+- Schools no longer in the API: flag for manual review rather?
+
+---
+
+## 11. Open Items
+
+### Fields and mapping
+- What do the `filter_instelling_status_erkenning` values (`I`, `S`, `W`, `X`) mean? Only `E` (erkend) is clear. Need to verify via API or documentation.
+- Which `instelling_niveau` and `instelling_type` values to include? Enum values are in `instelling.yaml`.
+- Do we need **translations**? The Instelling API only provides Dutch, but some Brussels schools may need French.
```
grubolsch, line 312: BOA is just for flanders

docs/scholen_api/analysis.md
```diff
@@ -0,0 +1,336 @@
...
+- How to format `instelling_busnummer`? E.g. `"Kerkstraat 10 bus 3"` or `"Kerkstraat 10/3"`?
+- `instelling_gemeente` is uppercase (e.g. `"MECHELEN"`). Title-case it, or does UDB3 normalize?
+
+### Labels
+- How to map `instelling_niveau` / `instelling_hoofdstructuur` to existing `cultuurkuur_` labels?
+- Should `instelling_nummer` be a visible or hidden label?
+
+### Rate limits
+- Actual rate limits on the Instelling API? The spec mentions HTTP 429 but no numbers.
+- UDB3 API throughput for bulk operations?
```
grubolsch, line 323: One thing I also miss is how to handle failure and partial runs. For example if the script should be able to uimport specific instelling_ids or from a certain id, in case of a failure. So we don't always have to run everything.

grubolsch (approved):

lucwollants:
No longer relevant, because the data team will handle this and store them inside uitwisselingsplatform.

PR #2226, opened 2026-04-21
III-7108 [PR 1/3] Add overnight boolean to subEvent domain model

grubolsch:
## Summary

- Adds `KAMP_OF_VAKANTIE_TERM_ID = '0.57.0.0.0'` constant to `EventTypeResolver` as the single source of truth for this term id
- Adds `overnight` boolean property to `SubEvent` (`isOvernight()` / `withOvernight(bool)`) and `SubEventUpdate` (`getOvernight(): ?bool` / `withOvernight(?bool)`)
- Enforces domain invariants in `Event`:
  - `updateSubEvents()` preserves the existing overnight value when not explicitly provided in the patch; throws `InvalidArgumentException` if any sub-event ends up with `overnight=true` while the event type is not `KAMP_OF_VAKANTIE_TERM_ID`
  - `updateCalendar()` override performs the same term check before delegating to `parent::updateCalendar()`
  - `updateType()` override resets overnight to `false` on all sub-events and emits `CalendarUpdated` when the type changes away from `KAMP_OF_VAKANTIE_TERM_ID` and at least one sub-event has `overnight=true`

## Test plan

- [ ] `SubEventTest` — default false, withOvernight(true), clear to false, immutability
- [ ] `SubEventUpdateTest` — default null, withOvernight(true/false/null), immutability
- [ ] `EventTest` — overnight preserved on unrelated PATCH, throws without term on updateSubEvents, throws without term on updateCalendar, reset on type change (single + multiple calendars), no CalendarUpdated when no overnight to reset

## Related PRs

- PR 2/3: Write path — deserialization, JSON schema, HTTP validation
- PR 3/3: Read model — normalizer, denormalizer, feature tests

---
Ticket https://jira.publiq.be/browse/III-7108

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -443,11 +443,11 @@ final class Event extends Offer
 
     public function updateType(Category $category): void
     {
-        $wasCamp = $this->typeId === EventTypeResolver::CAMP_UUID;
+        $wasCamp = $this->typeId === EventTypeResolver::CAMP_TERM_ID;
 
         parent::updateType($category);
 
-        $isNowCamp = $category->getId()->toString() === EventTypeResolver::CAMP_UUID;
+        $isNowCamp = $category->getId()->toString() === EventTypeResolver::CAMP_TERM_ID;
 
         if ($wasCamp && !$isNowCamp && $this->calendar instanceof CalendarWithSubEvents) {
             $subEvents = $this->calendar->getSubEvents()->toArray();
@@ -484,14 +484,14 @@ final class Event extends Offer
      */
     private function assertOvernightAllowed(array $subEvents): void
     {
-        if ($this->typeId === EventTypeResolver::CAMP_UUID) {
+        if ($this->typeId === EventTypeResolver::CAMP_TERM_ID) {
             return;
         }
 
         foreach ($subEvents as $subEvent) {
             if ($subEvent->isOvernight()) {
                 throw new \InvalidArgumentException(
-                    'overnight is only allowed when the event has term ' . EventTypeResolver::CAMP_UUID
+                    'overnight is only allowed when the event has term ' . EventTypeResolver::CAMP_TERM_ID
                 );
             }
         }
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -444,7 +444,7 @@ final class Event extends Offer
 
     public function updateType(Category $category): void
     {
-        $wasCamp = $this->typeId === EventTypeResolver::CAMP_OR_VACATION_TERM_ID;
+        $wasCamp = EventTypeResolver::isOvernightAllowed($this->typeId);
 
         parent::updateType($category);
 
@@ -483,7 +483,7 @@ final class Event extends Offer
      */
     private function assertOvernightAllowed(array $subEvents): void
     {
-        if ($this->typeId === EventTypeResolver::CAMP_OR_VACATION_TERM_ID) {
+        if (EventTypeResolver::isOvernightAllowed($this->typeId)) {
             return;
         }
 
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -58,7 +58,6 @@ use CultuurNet\UDB3\Event\Events\ThemeUpdated;
 use CultuurNet\UDB3\Event\Events\TitleTranslated;
 use CultuurNet\UDB3\Event\Events\TitleUpdated;
 use CultuurNet\UDB3\Event\Events\TypeUpdated;
-use CultuurNet\UDB3\Event\EventTypeResolver;
 use CultuurNet\UDB3\Event\Events\TypicalAgeRangeDeleted;
 use CultuurNet\UDB3\Event\Events\TypicalAgeRangeUpdated;
 use CultuurNet\UDB3\Event\Events\VideoAdded;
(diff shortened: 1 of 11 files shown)


grubolsch:
I am doubting if I should move the UUID of the kamp-term in config or not.
On the one hand it feels like config, on the other side it is the same on all envs, and will probably never change.

grubolsch:
@claude

grubolsch:
@claude

grubolsch:
@claude

lucwollants:
> I am doubting if I should move the UUID of the kamp-term in config or not. On the one hand it feels like config, on the other side it is the same on all envs, and will probably never change.

I see we also have `'0.3.1.0.0'` inside `EventTypeResolver`, so I think it's fine to hardcode it.

src/Event/Event.php
```diff
@@ -434,6 +433,69 @@ public function updateSubEvents(SubEventUpdate ...$subEventUpdates): void
...
+
+        return $calendar->withBookingAvailability($this->calendar->getBookingAvailability());
+    }
+
+    /**
+     * @param SubEvent[] $subEvents
+     */
+    private function assertOvernightAllowed(array $subEvents): void
+    {
+        if ($this->typeId === EventTypeResolver::CAMP_OR_VACATION_TERM_ID) {
```
JonasVHG: I would have rather created a new function in EventTypeResolver like we already have with `isOnlyAvailableUntilStartDate()`, something like `isOvernightAllowed()`, because we now have 1 type as a const in `EventTypeResolver`, while all the others are not. It would also be easier if we ever decide to allow other terms for overnight stay.

src/Model/ValueObject/Calendar/SubEventUpdate.php
```diff
@@ -18,6 +18,8 @@ final class SubEventUpdate
     private ?BookingInfo $bookingInfo = null;
     private ?TimeImmutableRange $childcareTimeRange = null;
 
+    private ?bool $overnight = null;
```
lucwollants, line 21: Is there a reason why `SubEvent` defaults to `false` and here inside `SubEventUpdate` it defaults to `null`? I would try to keep it consistent?

src/Event/Event.php
```diff
@@ -434,6 +433,69 @@ public function updateSubEvents(SubEventUpdate ...$subEventUpdates): void
...
+    public function updateType(Category $category): void
+    {
+        $wasCamp = $this->typeId === EventTypeResolver::CAMP_OR_VACATION_TERM_ID;
+
+        parent::updateType($category);
+
+        if ($this->calendar instanceof CalendarWithSubEvents && $wasCamp && $this->typeId !== EventTypeResolver::CAMP_OR_VACATION_TERM_ID) {
+            $subEvents = $this->calendar->getSubEvents()->toArray();
+
+            $hasOvernight = false;
```
lucwollants: I think we can create a simple `SubEvents::withoutOvernight` which calls the `SubEvent::withOvernight(false)`.

There is also no reason to track `$hasOvernight` because applying  `CalendarUpdated` does first checks for actual changes with `sameCalendars` before storing it inside the database.

src/Model/ValueObject/Calendar/SubEventUpdate.php
grubolsch: Yes, the difference is intentional because I need to know in the case of a PUT partial update "not provided in this update request, leave the existing value unchanged." false would mean "explicitly set overnight to false."

I got this from otherfields in SubEventUpdate (startDate, status, bookingAvailability, etc.) follow the same pattern: null = "no change".

src/Event/Event.php
grubolsch: Added the method but did not remove the public const because it is used in several error messages.

src/Event/Event.php
grubolsch: When I removed the $hasOvernight guard and just called apply() every time, the test it_does_not_emit_calendar_updated_when_type_changes_away_from_kamp_of_vakantie_but_no_overnight_is_set started failin, so I think the event was still being recorded even though no overnight flags were set.

So still added a guard check but just use `sameCalendars` instead fo checking specific changed fields.

grubolsch:
All feedback has been processed

src/Event/Event.php
lucwollants: I think that is caused by `sameCalendars` which uses `CalendarNormalizer`. I think it's worth checking what is different because they should be the same?

lucwollants: Ok, I now see the full flow. It is correctly used inside the update method 👍

src/Event/Event.php
```diff
@@ -434,6 +433,60 @@ public function updateSubEvents(SubEventUpdate ...$subEventUpdates): void
...
+    public function updateCalendar(Calendar $calendar): void
+    {
+        if ($calendar instanceof CalendarWithSubEvents) {
+            $this->assertOvernightAllowed($calendar->getSubEvents()->toArray());
+        }
+
+        parent::updateCalendar($calendar);
+    }
+
+    public function updateType(Category $category): void
```
lucwollants, line 445: This can now benefit from early return statements with the handy method you added

Something like this might work now
```
    public function updateType(Category $category): void
    {
        parent::updateType($category);

        if (!($this->calendar instanceof CalendarWithSubEvents)) {
            return;
        }

        if (EventTypeResolver::isOvernightAllowed($this->typeId)) {
            return;
        }

        $resetSubEvents = $this->calendar->getSubEvents()->withoutOvernight()->toArray();
        $updatedCalendar = $this->rebuildCalendarFromSubEvents($resetSubEvents);
        if (!$this->sameCalendars($this->calendar, $updatedCalendar)) {
            $this->apply(new CalendarUpdated($this->eventId, $updatedCalendar));
        }
    }
```

lucwollants (approved):
Looks good.

Maybe with the final pull request, we can remove the plan?

PR #2231, opened 2026-04-23
III-7138: Add caching decorator for holidays endpoint

grubolsch:
## Summary

- Wraps `OpenHolidaysApiService` with a `CachedHolidaysService` decorator
- Caches each unique `startDate` + `endDate` combination for **1 month** (2592000 seconds)
- Uses the existing `CacheFactory` + `app_cache` (Redis) pattern, consistent with other cached services in the codebase
- Cache key format: `holidays_{startDate}_{endDate}`

## Test plan

- [ ] Unit tests: `tests/Holidays/CachedHolidaysServiceTest.php`
- [ ] Test that holidays are fetched from cache on repeated calls with the same params
- [ ] Test that different date ranges produce different cache keys
- [ ] Test that the inner service is called on cache miss

## Notes

This is PR 2/2, to be merged after cultuurnet/udb3-backend#2230.


grubolsch:
@claude

grubolsch:
@claude

grubolsch:
Readfy for review

src/Holidays/CachedHolidaysService.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function getHolidays(DateTimeImmutable $startDate, DateTimeImmutable $endDate): array
+    {
+        return $this->cache->get(
+            $this->createCacheKey($startDate, $endDate),
+            fn () => $this->baseHolidaysService->getHolidays($startDate, $endDate)
+        );
+    }
+
+    private function createCacheKey(DateTimeImmutable $startDate, DateTimeImmutable $endDate): string
```
lucwollants, line 26: I wonder about the granularity of the cache key.

This even makes me wonder if we actually need to consume this API instead of creating a fixed list of holidays. This static approach would introduce a lot less of code, does not require a cache and could even go back in time.

bertramakers: We don't want a static list because we'd have to remember to update it every year, as the dates of some holidays shift and we cannot always determine them easily programmatically. For example "Easter is celebrated on the first Sunday after the Paschal full moon (a mathematical approximation of the first astronomical full moon, on or after 21 March – itself a fixed approximation of the March equinox)". https://en.wikipedia.org/wiki/Date_of_Easter

Assuming query parameters with a start and end date are required on this endpoint I think it makes sense to use them as cache keys?

Do we really need the caching though? I would just call the API and do an `array_map()` on the data to transform it into the format that we need and consider it a proxy. We can always add caching later if really needed?

lucwollants: > update it every year

I was thinking about something like 5 years in the past and 20 years in the future 😁

bertramakers: Ah okay I see what you mean. But then we'd need custom code to filter on the start and end date, either here in the backend or in the frontend. So I'm not sure if it'd really be a **lot** less code?

grubolsch: I discussed this with Sarah and for now the API is fine, 5 years in the future is good enough.

lucwollants (approved):

PR #2232, opened 2026-04-27
III-7156 Fix terms caching

lucwollants:
### Added
- Added regression test `it_does_not_call_the_taxonomy_api_when_no_getter_is_invoked`

### Changed
- Refactor the actual terms HTTP call out of the constructor into `fetchTerms`
- Improved existing tests

---

Ticket: https://jira.uitdatabank.be/browse/III-7156

### DIFF
--- a/src/Taxonomy/JsonTaxonomyApiClient.php
+++ b/src/Taxonomy/JsonTaxonomyApiClient.php
@@ -16,33 +16,13 @@ use Psr\Log\LoggerInterface;
 
 final class JsonTaxonomyApiClient implements TaxonomyApiClient
 {
-    private array $terms;
+    private ?array $terms = null;
 
     public function __construct(
         private readonly ClientInterface $client,
         private readonly string $termsEndpoint,
         private readonly LoggerInterface $logger
     ) {
-        $request = new Request(
-            'GET',
-            $this->termsEndpoint,
-        );
-
-        $response = $this->client->sendRequest($request);
-        if ($response->getStatusCode() !== 200) {
-            $this->logger->error('Taxonomy Api returned non-200 status code', [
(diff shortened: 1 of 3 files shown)


lucwollants:
@claude review

src/Taxonomy/JsonTaxonomyApiClient.php
```diff
@@ -72,13 +52,36 @@ public function getEventFacilities(): Categories
 
     public function getNativeTerms(): array
     {
-        return $this->terms;
+        return $this->fetchTerms();
+    }
+
+    private function fetchTerms(): array
```
bertramakers: Nit: I may be overlooking some detail but I don't see a reason to make this a separate method from `getNativeTerms()`? They effectively do the same thing since `getNativeTerms()` only calls `fetchTerms()` without any extra logic. So I think this can be moved to `getNativeTerms()`, and the one other internal usage of `fetchTerms()` can be changed to just use `getNativeTerms()` directly?

(Also I'm not sure what the "native" in `getNativeTerms()` is supposed to mean, but since that was already defined before this PR it's probably too much scope creep to go into that 😄)

bertramakers (approved):
Clean fix, nice test coverage! Just one small optional nit.

src/Taxonomy/JsonTaxonomyApiClient.php
lucwollants: There is indeed no need for the extra `fetchTerms` method. I removed it in 9e40d1d

The `getNativeTerms` follows the pattern we use when we want to convert a VO to its primitive form. In the context of the `TaxonomyApiClient` interface, all methods return `Categories`. But the existing `TermRepository` needed a plain `array`, so the `getNativeTerms` method was added for this.

PR #2236, opened 2026-04-30
Restore handling of invalid argument exception for date range

lucwollants:
### Changed
- Restore handling of invalid argument exception for date range to solve a 500 error


JonasVHG (approved):

bertramakers (approved):
Looks good to me based on the original code before the childcare time changes. Thanks for the extra tests!

PR #2238, opened 2026-05-06
III-7096 Improve boosting test cases

steftrenson:
### Added
- Add acceptance tests for positive boosting of events, places, and offers via the Search API v3 `q` parameter using TF-IDF term scoring (e.g. `(term^10) OR (NOT term)`)
- Add acceptance tests for negative boosting of events, places, and offers via the Search API v3 `q` parameter (e.g. `(term^0.1) OR (NOT term)`)
- Add a test case for the alternative negative boosting pattern: positively boosting the NOT-branch instead of decimating the term branch (e.g. `(term) OR (NOT term)^10`)
- Add a new Behat step `I create an event with name :name and save the :jsonPath as :variableName` to create an event with a specific name in a single step
- Add a new Behat step `I create a place with name :name and save the :jsonPath as :variableName` to create a place with a specific name in a single step
- Add a new Behat step `I wait for :count result(s) at :url with parameters:` that polls the Search API until the expected number of results is indexed, replacing brittle fixed `wait N seconds` delays
- Add `@negativeBoosting` tag and a `@BeforeScenario` hook that skips the tagged scenarios when `skip_negative_boosting` is set to `true` in `config.features.php`, to allow environments running Elasticsearch 8 (where negative boosting is broken) to skip those tests

### Changed
- Rewrite the boosting feature file to use isolated scenarios per resource type (`@testIsolation`) instead of a single shared Background that set up all resources upfront
- Replace fixed `wait 2 seconds` calls in boosting scenarios with the new polling step that waits until the expected document count is actually indexed
- Suffix all search terms with a random string per test run to prevent stale documents from previous runs accumulating on the same Elasticsearch shard and inflating `docFreq`, which would make TF-IDF score ordering unreliable
- Use equal-length document names in negative boosting tests so that `fieldNorm` does not inadvertently tip the score ordering before the boost query is applied

> **Note:** A companion change in the Search repository is also required. The Elasticsearch index must be configured to use a single shard so that IDF scores are computed globally and the score-based ordering is deterministic. Without this, the boosting tests may fail when documents land on different shards with different `docFreq` values.

---
Ticket: https://jira.uitdatabank.be/browse/III-7096


grubolsch (approved):

features/Steps/RequestSteps.php
```diff
@@ -248,4 +248,22 @@ private function waitForItemWithUrlToBeIndex(string $url): void
...
+    /**
+     * @Given I wait for :count result(s) at :url with parameters:
+     */
+    public function iWaitForResultsAtWithParameters(int $count, string $url, TableNode $parameters): void
+    {
+        $params = $this->addScenarioLabelToSearchParameters($url, $parameters->getRows());
+
+        $elapsedTime = 0;
+        do {
+            $response = $this->getHttpClient()->getWithParameters($url, $params, $this->variableState);
```
lucwollants, line 262: I wonder if this is missing the disabling of the default filters, like in `waitForItemWithUrlToBeIndex`

lucwollants (approved):

features/Steps/RequestSteps.php
steftrenson: Yes, it makes sense to add it here as well. Thanks for the feedback!

lucwollants (approved):
Looks good!

PR #2241, opened 2026-05-08
III-7133 Search departure places feature tests

steftrenson:
### Added
- Add `features/search/departure-places.feature` with four Behat scenarios verifying that departure places are correctly indexed and searchable in SAPI: departure places returned in embedded search results, filtering by departure place UUID via the `q` parameter (including OR queries and events with multiple departure places), filtering via the `departurePlaces[]` URL parameter, and AND logic when multiple `departurePlaces[]` values are provided
- Add `I wait for the JSON response at :jsonPath to have :nrOfEntries entries` Behat step that replays the last GET request in a polling loop (up to 5 seconds) until the array at the given JSON path reaches the expected length, avoiding flaky fixed-duration waits for async SAPI indexing
- Add `I wait for the JSON response at :jsonPath to be :value` Behat step that replays the last GET request in a polling loop (up to 5 seconds) until a scalar value at the given JSON path matches the expected value, used to reliably wait for `totalItems` to reflect the correct count before asserting search results
- Add `lastGetUrl` and `lastGetParams` to `RequestState` so the two new polling steps can replay the most recent parameterised GET request without re-applying the `@testIsolation` scenario label filter a second time

---
Ticket: https://jira.uitdatabank.be/browse/III-7133

### DIFF
--- a/features/search/departure-places.feature
+++ b/features/search/departure-places.feature
@@ -39,18 +39,20 @@ Feature: Test departure places in search results
     And I keep the value of the JSON response at "id" as "departurePlaceId2"
     And I create a minimal place and save the "url" as "departurePlaceUrl3"
     And I keep the value of the JSON response at "id" as "departurePlaceId3"
+    And I create a minimal place and save the "url" as "departurePlaceUrl4"
+    And I keep the value of the JSON response at "id" as "departurePlaceId4"
     And I create an event from "events/audience-type/event-audience-type-children-only.json" and save the "id" as "eventId1"
     And I publish the event at "/events/%{eventId1}"
     And I set the JSON request payload to:
     """
-    ["%{departurePlaceUrl1}", "%{departurePlaceUrl3}"]
+    ["%{departurePlaceUrl1}", "%{departurePlaceUrl3}", "%{departurePlaceUrl4}"]
     """
     And I send a PUT request to "/events/%{eventId1}/departurePlaces/"
     And I create an event from "events/audience-type/event-audience-type-children-only.json" and save the "id" as "eventId2"
     And I publish the event at "/events/%{eventId2}"
     And I set the JSON request payload to:
     """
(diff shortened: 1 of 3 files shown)


features/search/departure-places.feature
```diff
@@ -0,0 +1,142 @@
...
+
+  Background:
+    Given I am using the UDB3 base URL
+    And I am using an UiTID v1 API key of consumer "uitdatabank"
+    And I am authorized as JWT provider user "centraal_beheerder"
+    And I send and accept "application/json"
+    And I create a minimal place and save the "url" as "placeUrl"
+
+  @testIsolation
+  Scenario: Departure places are returned in search results
```
lucwollants:
```suggestion
  Scenario: Departure places are embedded in search results
```

features/search/departure-places.feature
```diff
@@ -0,0 +1,142 @@
...
+    Then I wait for the JSON response at "totalItems" to be 1
+    And the JSON response at "member/0/@id" should include "%{eventId1}"
+    And I send a GET request to "/events" with parameters:
+      | disableDefaultFilters | true                                 |
+      | q                     | departurePlaces:%{departurePlaceId2} |
+    Then I wait for the JSON response at "totalItems" to be 1
+    And the JSON response at "member/0/@id" should include "%{eventId2}"
+    And I send a GET request to "/events" with parameters:
+      | disableDefaultFilters | true                                                                        |
+      | q                     | departurePlaces:%{departurePlaceId1} OR departurePlaces:%{departurePlaceId2} |
```
lucwollants, line 76: After reading the API docs, it's clear that the URL parameters only support an AND. The q parameter does support both AND and OR.

Should we add an extra feature test for AND with the advanced parameter? I know it almost feels like testing AND/OR behaviour of Elasticsearch. But we heavily rely on our acceptance tests for doing releases, so better to cover as much as possible.

PR #2240, opened 2026-05-07
[Low prio] Replace abandoned mathiasverraes/money with moneyphp/money

grubolsch:
I wanted to experiment with Claude agents and git worktree to run multiple agents at the same time, so I just picked some low hanging maintance work.
This is a low prio PR, only pick up if you want.

## Summary

- Replaces the abandoned `mathiasverraes/money ^v1.3.0` with the maintained `moneyphp/money ^4.0` (v4.9.0 installed)
- `Currency::getName()` renamed to `Currency::getCode()` across 6 source files
- `Money` constructor now requires `string` amount under `strict_types=1` — wrapped integer expressions with `(string)((int) ...)` in `MoneyFactory` and `ParsedPriceForATheater`
- `MoneyNormalizer` and `PriceInfo::serialize()` cast `getAmount()` to `int` to preserve backward-compatible JSON and event-store output (cents are always whole numbers)
- All 139 `new Money(int, ...)` literal calls in test files updated to `new Money('int', ...)`

## Test plan

- [x] All 5416 tests pass in Docker (`make test`)
- [x] No changes to JSON-LD output, RDF URIs, or event-store serialization format

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2242, opened 2026-05-08
Replace mathiasverraes/money with an internal Money DTO

grubolsch:
## Why

`mathiasverraes/money` is abandoned. The obvious next step would be migrating to `moneyphp/money`, but looking at the actual usage first revealed that the whole package was being used as a 2-field struct: store an integer amount in cents and a currency code string.

**Nothing else from the package was used — no arithmetic, no formatting, no currency validation, no comparison.**

Migrating to `moneyphp/money` would mean dealing with its v4 API quirks (string-typed constructor, `getCode()` instead of `getName()`), adding `(string)((int) $amount)` casts in `MoneyFactory`, and `(int)` casts in serializers — all to keep using a library as a plain data container.

## What

Two small value objects in `src/Model/ValueObject/Price/`:

```php
class Money {
    public function __construct(int $amount, Currency $currency) {}
    public function getAmount(): int {}
    public function getCurrency(): Currency {}
}

class Currency {
    public function __construct(string $code) {}
    public function getCode(): string {}
}
```

- `mathiasverraes/money` removed from `composer.json`
- All `use Money\Money` / `use Money\Currency` imports updated across 11 source files and 23 test files
- `getName()` → `getCode()` in the three projector/converter files that read the currency code
- The one test using `->multiply()` was testing `Tariff::withPrice()`, not multiplication — replaced with a direct `new Money(2000, ...)` 

Closes #2240 (the moneyphp/money migration PR, superseded by this).

## Test plan

- [x] All 5416 tests pass (`make test`)
- [x] No changes to serialized output or RDF URIs


grubolsch:
Jenkins test: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/648/

tests/Model/Offer/ImmutableOfferTest.php
```diff
@@ -331,8 +330,7 @@ public function it_should_return_no_price_info_by_default(): void
 
     /**
      * @test
-     * @throws UnknownCurrencyException
-     */
+         */
```
JonasVHG: There's too much whitespace here, same for line 353.

lucwollants (approved):
Looks good! Two simple classes, but all functionality is still working.

JonasVHG (approved):

PR #2243, opened 2026-05-11
III-7188: Fix childcare time validation failing for UTC-offset event dates

grubolsch:
## Summary

- When frontends (and integrators) send `startDate`/`endDate` as UTC timestamps, the `childcare.start`/`childcare.end` validation was comparing against raw UTC hours instead of local CET/CEST clock time, causing false validation errors.
- Fix: in `TimeImmutableRange::dateTimeToMinutes()`, normalise the incoming `DateTimeImmutable` to `Europe/Brussels` before extracting hours and minutes.
- Updated the fixture and all affected unit/feature tests to use explicit timezone offsets so the test intent stays clear and correct under the new behaviour. Added a new feature scenario and unit test that directly reproduce the reported bug.

### Ticket
Ticket: https://jira.uitdatabank.be/browse/III-7188

### DIFF
--- a/features/event/sub-event-childcare-time.feature
+++ b/features/event/sub-event-childcare-time.feature
@@ -7,106 +7,112 @@ Feature: Test SubEvent childcare times
     And I send and accept "application/json"
     And I create a place from "places/place.json" and save the "url" as "placeUrl"
 
-  Scenario: Create an event with childcare times on a single calendar subEvent
-    When I set the JSON request payload to:
-    """
-    {
-      "mainLanguage": "nl",
-      "name": {"nl": "Event met kinderopvang"},
-      "terms": [{"id": "0.50.4.0.0", "label": "Concert", "domain": "eventtype"}],
-      "location": {"@id": "%{placeUrl}"},
-      "calendarType": "single",
-      "startDate": "2021-05-17T16:00:00+00:00",
-      "endDate": "2021-05-17T22:00:00+00:00",
-      "subEvent": [
-        {
-          "startDate": "2021-05-17T16:00:00+00:00",
-          "endDate": "2021-05-17T22:00:00+00:00",
-          "childcare": {
-            "start": "15:00",
-            "end": "23:00"
-          }
-        }
-      ]
-    }
-    """
+  Scenario: Create a single-calendar event with childcare times in CET
+    Given I set the variable "startDate" to "2021-05-17T16:00:00+02:00"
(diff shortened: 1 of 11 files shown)


grubolsch:
@claude

grubolsch:
@claude

grubolsch:
@claude

grubolsch:
Jenkins build: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/653/

features/event/sub-event-childcare-time.feature
```diff
@@ -234,6 +234,36 @@ Feature: Test SubEvent childcare times
     And the JSON response at "subEvent/0/childcare/end" should be "23:00"
     And the JSON response should not have "subEvent/0/childcare/start"
 
+  Scenario: Create an event where childcare times are valid in CET but startDate and endDate are sent as UTC
+    When I set the JSON request payload to:
+    """
+    {
```
JonasVHG: I would not put an entire event-json hardcoded in a test, I think it would be best to create a separate test file.
Also some other tests have also use UTC, it think we need tests for both create/update bothin UTC & CET.

grubolsch: More tests with both timezone.
Added more data files.

src/Model/ValueObject/TimeImmutableRange.php
```diff
@@ -70,6 +71,8 @@ public function endIsAfterTime(Time $time): bool
 
     private function dateTimeToMinutes(DateTimeImmutable $dateTime): int
     {
-        return (int) $dateTime->format('H') * 60 + (int) $dateTime->format('i');
+        // Childcare applies exclusively to Flanders; HH:mm times are always entered in Brussels local time.
```
bertramakers: Small nit:

```suggestion
        // We always handle HH:mm time information as Brussels local time. To compare it to a datetime, the datetime should also be converted to Europe/Brussels
```

I know this valueobject is only used for childcare at the moment, but since it doesn't have childcare in its namespace or class name I would refrain from directly referencing it. Since we might want to re-use this in other places later. We'll probably want the same conversion then, but the childcare comment might become confusing when we use this class in other features and it directly references childcare here.  Because people might wonder if the conversion should also be done for their feature or not etc.

bertramakers (approved):
Thanks for the comments in the tests to see the timezone conversions at a glance! I just have a small suggestion for the comment in `TimeImmutableRange` but overall this looks good and well tested!

PR #2244, opened 2026-05-12
Add feature tests for adjusted opening hours on places

grubolsch:
## Summary

- Adjusted opening hours (`openingHoursAdjustedDays`) already worked for places at the API level — the shared `UpdateCalendarRequestHandler`, `UpdateCalendarValidatingRequestBodyParser`, and `AdjustedDaysValidator` handle both events and places, and `place-calendar-put.json` already references `place-openingHoursAdjustedDays.json`. Feature test coverage was missing.
- Added `features/place/opening-hours-adjusted.feature` mirroring the event equivalent with 10 scenarios: create periodic, create permanent, add via PUT, replace via PUT, clear via PUT, and 5 validation error cases (startDate after endDate, window before calendar start, window after calendar end, two overlap cases).
- Childcare scenarios are excluded — places do not support childcare times.
- Uses `Given I create a random name of 6 characters` + `%{name}` in all fixtures to avoid duplicate place detection (same pattern as `duplicate.feature`).

## Ticket
https://jira.publiq.be/browse/III-7136

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -6090,12 +6090,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/cultuurnet/udb3-json-schemas.git",
-                "reference": "3c4b1ec0c463250972d04c85cb64e5c669d08a7d"
+                "reference": "4df3bd56d2ef1f97f535becc9bd8c8fb8af1844d"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/cultuurnet/udb3-json-schemas/zipball/3c4b1ec0c463250972d04c85cb64e5c669d08a7d",
-                "reference": "3c4b1ec0c463250972d04c85cb64e5c669d08a7d",
+                "url": "https://api.github.com/repos/cultuurnet/udb3-json-schemas/zipball/4df3bd56d2ef1f97f535becc9bd8c8fb8af1844d",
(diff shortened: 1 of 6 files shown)


grubolsch:
Turned out I did not need any code changes, after the JSON schema change, so I just added some feature tests.

2 things that I am still thinking about, please share your opinion:
- I copied all feature tests from events/ to places/, but maybe that is overkill here? 
- I did not add any feature tests because all the feature tests already test it end-to-end.  There's no new logic to unit-test: the handler, validator, and schema are all shared with events and already tested there.

bertramakers:
> Turned out I did not need any code changes, after the JSON schema change, so I just added some feature tests.
> 
> 2 things that I am still thinking about, please share your opinion:
> 
>     * I copied all feature tests from events/ to places/, but maybe that is overkill here?
> 
>     * I did not add any feature tests because all the feature tests already test it end-to-end.  There's no new logic to unit-test: the handler, validator, and schema are all shared with events and already tested there.

Looks good @grubolsch , thanks! I don't think the feature tests are overkill. I agree there is no need for extra unit tests since most of the internal code is re-used and already unit tested.

From my side I only wonder if the fact that childcare is not supported on place should also be covered with a feature test? Having a test that explicitly checks that it's not supported can give some clarity in the future if we ever wonder if it was intentional or not that it's unsupported.

bertramakers (approved):

PR #2245, opened 2026-05-12
III-7137 Throw error when remainingCapacity sent on top level

grubolsch:
## Summary
- `remainingCapacity` is only valid inside `subEvent.bookingAvailability` (see `event-subEvent-bookingAvailability.json` in `publiq/udb3-json-schemas`). At the top-level `bookingAvailability` of an Event or Place, it has no effect — but until now it was silently accepted.
- New `RemainingCapacityValidatingRequestBodyParser` (wired into `OfferValidatingRequestBodyParser`, so it applies to both Event and Place) throws a `bodyInvalidData` ApiProblem with a SchemaError at `/bookingAvailability/remainingCapacity` and a user-friendly message.
- Added unit tests in `ImportEventRequestHandlerTest` and `ImportPlaceRequestHandlerTest` covering the new rejection.

## Ticket

https://jira.publiq.be/browse/III-7137

### DIFF
--- a/src/Http/Offer/RemainingCapacityValidatingRequestBodyParser.php
+++ b/src/Http/Offer/RemainingCapacityValidatingRequestBodyParser.php
@@ -23,7 +23,7 @@ final class RemainingCapacityValidatingRequestBodyParser implements RequestBodyP
             throw ApiProblem::bodyInvalidData(
                 new SchemaError(
                     '/bookingAvailability/remainingCapacity',
-                    'remainingCapacity can only be set on a subEvent, not on the top-level bookingAvailability.'
+                    'remainingCapacity can only be set on a sub-event entry, not on the top-level bookingAvailability. Set it under /subEvent/{index}/bookingAvailability instead.'
                 )
             );
         }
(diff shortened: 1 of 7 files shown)


grubolsch:
@claude

grubolsch:
@claude

grubolsch:
Jenkins: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/655/

src/Http/Offer/RemainingCapacityValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,33 @@
...
+
+        if (!is_object($data) || !isset($data->bookingAvailability) || !is_object($data->bookingAvailability)) {
+            return $request;
+        }
+
+        if (property_exists($data->bookingAvailability, 'remainingCapacity')) {
+            throw ApiProblem::bodyInvalidData(
+                new SchemaError(
+                    '/bookingAvailability/remainingCapacity',
+                    'remainingCapacity is not valid on the top-level bookingAvailability.'
```
bertramakers: Nit:
```suggestion
                    'remainingCapacity is not supported on the top-level bookingAvailability.'
```
("Not valid" could imply the data that's sent is not valid, while "not supported" is a bit more clear that it will never work)

bertramakers: (You'll also need to update the tests if you accept this suggestion of course)

bertramakers (approved):
Looks good to me, seems consistent with other custom validation logic and well covered with tests. Just one minor suggestion regarding the exact error message.

PR #2247, opened 2026-05-20
III-7189 birthdateRange acceptance tests

JonasVHG:
### Added

- `event-with-birthdate-range-in-2022.json`: New data file

### Changed

- `event-with-birthdate-range.json`: Renamed to `event-with-birthdate-range-in-2022.json`.
- `event/create.feature`: Use new name in existing acceptance test.
- `Makefile`:  Exclude `wip` tests.
- `search/url-parameters-offers.feature`: Add `wip` test for new common filter for `birthdateRange`.
- `search/advanced-queries-offers.feature`: Add `wip` test for advanced query for `birthdateRange`.
- `search/audience-type.feature`: Use `wip` instead of `external`

### Note

The aceptance tests for matching the typical-age-range will be added in a next PR, the added once are marked as `work-in-progress` because, the search-api PR is not finished yet.

---

Ticket: https://jira.publiq.be/browse/III-7189


JonasVHG:
@claude, review.

lucwollants (commented):
I think these acceptance tests should be extended with more complex combinations for the typical age range and the birthdate range.

Now we always default to an empty index for typical age range, but we also want to verify various combinations. Even combinations that are correct and combinations that don't make sense.

For example

```
Event: typicalAgeRange=6-12, birthdateRange=2021-01-31..2021-12-31

Search birthdateRange=2013-01-01..2019-12-31 (people born 2013-2019, aged 7-13 today in 2026):
- SHOULD 1: birthdateRange 2021 overlaps 2013-2019? No
- SHOULD 2: typicalAgeRange [6, 12] overlaps [7, 13]? Yes → match
```

features/search/advanced-queries-offers.feature
```diff
@@ -215,6 +215,36 @@ Feature: Test the Search API v3 advanced queries on offers
...
+  @wip
+  Scenario: Search for birthdate ranges using an advanced query
+    When I create a minimal place and save the "url" as "placeUrl"
+    And I create an event from "events/event-with-birthdate-range-in-2020.json" and save the "id" as "eventId2020"
+    And I wait for the event with url "/events/%{eventId2020}" to be indexed
+    And I create an event from "events/event-with-birthdate-range-in-2022.json" and save the "id" as "eventId2022"
+    And I wait for the event with url "/events/%{eventId2022}" to be indexed
+    And I am using the Search API v3 base URL
+    When I send a GET request to "/events" with parameters:
+      | q | id:(%{eventId2020} OR %{eventId2022}) AND birthdateRange:[2019-01-01 TO 2019-12-31] |
```
steftrenson: Can't the `@testIsolation` tag be used so the ids don't need to be explicitly passed anymore? Would make the test a bit more readable.

Same for the other test btw.

features/search/advanced-queries-offers.feature
```diff
@@ -215,6 +215,36 @@ Feature: Test the Search API v3 advanced queries on offers
     %{eventId}
     """
 
+  @wip
+  Scenario: Search for birthdate ranges using an advanced query
+    When I create a minimal place and save the "url" as "placeUrl"
+    And I create an event from "events/event-with-birthdate-range-in-2020.json" and save the "id" as "eventId2020"
+    And I wait for the event with url "/events/%{eventId2020}" to be indexed
+    And I create an event from "events/event-with-birthdate-range-in-2022.json" and save the "id" as "eventId2022"
+    And I wait for the event with url "/events/%{eventId2022}" to be indexed
```
steftrenson, line 225: 👍 on using the I wait for step, it'll make the test stable.

JonasVHG:
> I think these acceptance tests should be extended with more complex combinations for the typical age range and the birthdate range.
> 
> Now we always default to an empty index for typical age range, but we also want to verify various combinations. Even combinations that are correct and combinations that don't make sense.
> 
> For example
> 
> ```
> Event: typicalAgeRange=6-12, birthdateRange=2021-01-31..2021-12-31
> 
> Search birthdateRange=2013-01-01..2019-12-31 (people born 2013-2019, aged 7-13 today in 2026):
> - SHOULD 1: birthdateRange 2021 overlaps 2013-2019? No
> - SHOULD 2: typicalAgeRange [6, 12] overlaps [7, 13]? Yes → match
> ```

This is done in the follow up PR with intersection. This PR is only for the basic tests of `birthdateRange`.

steftrenson (approved):
Not entirely a fan of the @wip tag, but I hope this gets cleaned up afterwards.

JonasVHG:
> Not entirely a fan of the @wip tag, but I hope this gets cleaned up afterwards.

Yes, but that was decided because we try do to first write the apidocs, then the tests & then the actual implementation. This way we can write tests, that do not fail while the actual code is still not in the main branch. Once the actual code is on the main/master branch the @wip tags should go.

lucwollants (approved):

PR #2248, opened 2026-05-21
III-7133 Search departure places feature tests

steftrenson:
### Added
- Add `features/search/departure-places.feature` with four Behat scenarios verifying that departure places are correctly indexed and searchable in SAPI: departure places returned in embedded search results, filtering by departure place UUID via the `q` parameter (including OR queries and events with multiple departure places), filtering via the `departurePlaces[]` URL parameter, and AND logic when multiple `departurePlaces[]` values are provided
- Add `I wait for the JSON response at :jsonPath to have :nrOfEntries entries` Behat step that replays the last GET request in a polling loop (up to 5 seconds) until the array at the given JSON path reaches the expected length, avoiding flaky fixed-duration waits for async SAPI indexing
- Add `I wait for the JSON response at :jsonPath to be :value` Behat step that replays the last GET request in a polling loop (up to 5 seconds) until a scalar value at the given JSON path matches the expected value, used to reliably wait for `totalItems` to reflect the correct count before asserting search results
- Add `lastGetUrl` and `lastGetParams` to `RequestState` so the two new polling steps can replay the most recent parameterised GET request without re-applying the `@testIsolation` scenario label filter a second time

---
Ticket: https://jira.uitdatabank.be/browse/III-7133

### DIFF
--- a/features/State/RequestState.php
+++ b/features/State/RequestState.php
@@ -18,6 +18,9 @@ final class RequestState
     private string $json = '';
     private array $form = [];
 
+    private string $lastGetUrl = '';
+    private array $lastGetParams = [];
+
     public function getBaseUrl(): string
     {
         return $this->baseUrl;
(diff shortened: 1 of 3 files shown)


steftrenson:
Based on #2241

lucwollants (approved):

PR #2256, opened 2026-05-28
Remove swiftmailer, migrate event-export mails to symfony/mailer

grubolsch:
I got the idea looking at a dependabot request (https://github.com/cultuurnet/udb3-backend/pull/2253) that needed to update the symfony mailer. I remembered that we also had another package that sents mails in the code base, so I wondered if it would be difficult to remove one of them.
Turns out swiftmailer/swiftmailer is abandoned anyway, so I chose the symfony mailer to upgrade to.

## AI generated Summary
- `swiftmailer/swiftmailer` is abandoned; its replacement `symfony/mailer` is already a dependency (used by `Mailer/SmtpMailer` for ownership mails).
- The only remaining consumer was the event-export notification mail. Rewrote `NotificationMailer` and `DefaultMessageFactory` (now in `EventExport\Notification\Symfony\`) to build a `Symfony\Component\Mime\Email` and dispatch via `Symfony\Component\Mailer\MailerInterface`.
- `SwiftMailerServiceProvider` deleted; `ExportServiceProvider` now constructs its own `SymfonyMailer` from `config['mail']['smtp']` (the same DSN that `MailerServiceProvider` already uses).
- `swiftmailer/swiftmailer` removed from `composer.json` / `composer.lock`. `symfony/mailer` got an incidental patch bump (5.4.45 → 5.4.52, within the existing `^5.4`).

## Notes
- The legacy `'swiftmailer.options'` config key is no longer read. `config.php` is gitignored, so deployment/ops configs should drop that key separately (no runtime impact if left in — it just becomes dead config).
- The old `NotificationMailer::sendNotificationMail` printed `"sent N e-mails"` to stdout, derived from Swiftmailer's send-count return value. Symfony's `MailerInterface::send()` returns `void`, so the print has been removed. Errors still propagate as exceptions.

## Jenkins tests
https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/676/


grubolsch:
I will make the config PR for this if I get the green light from the team to do this change.

features/Steps/MailSteps.php
```diff
@@ -27,8 +27,8 @@ public function aMailHasBeenSentFromToWith(string $messageType, string $from, st
         assertEquals($to, $mailobject->getTo()->getByIndex(0)->toString());
         assertEquals($subject, $mailobject->getSubject());
         assertStringMatchesFormat(
-            $this->fixtures->loadMail($messageType, $this->variableState),
-            $mailobject->getContent()
+            rtrim($this->fixtures->loadMail($messageType, $this->variableState)),
```
lucwollants, line 30: I do wonder if we should "fix" this inside the `loadMail` method? Now it feels like something we might forget in the future.

src/EventExport/Notification/Symfony/DefaultMessageFactory.php
```diff
@@ -35,27 +37,16 @@ public function __construct(
         $this->subjectFactory = $subjectFactory;
     }
 
-    public function createMessageFor(EmailAddress $address, EventExportResult $eventExportResult): \Swift_Message
+    public function createMessageFor(EmailAddress $address, EventExportResult $eventExportResult): Email
```
lucwollants, line 40: This seems to be some reusable Factory for all things related to mails, but it is only used for `EventExportResult`. Should we create a follow-up pull request to align the event exports with the ownership mail creation?

lucwollants (approved):
Looks good!

I added two remarks, but they are not related to the removal of the SwiftMailer package and can be handled in another pull request.

PR #2265, opened 2026-06-03
III-7134 Add capacity to place bookingAvailability

grubolsch:
## Summary

- Add `capacity` support to the top-level `bookingAvailability` of places, mirroring the [apidocs PR](https://github.com/cultuurnet/apidocs/pull/509) and the [udb3-json-schemas branch](https://github.com/cultuurnet/udb3-json-schemas/tree/III-7134/add-capacity-to-places).
- Fix `CalendarDenormalizer` so the parsed top-level `bookingAvailability` is actually applied to `PermanentCalendar` and `PeriodicCalendar`. Previously it was parsed but silently dropped for these calendar types.
- Capacity now flows through `POST /places`, `PUT /places/{placeId}` and `PUT /places/{placeId}/calendar`. The dedicated `PUT /places/{placeId}/booking-availability` endpoint still returns 400 (unchanged), matching the documented behavior.

## Side effect on events (BC-safe)

`CalendarDenormalizer` is shared with events, so the fix also affects events with `permanent` or `periodic` calendars: an explicitly provided top-level `bookingAvailability` is now honored on those calendars instead of being silently dropped. This is BC-safe — when no value is sent, the denormalizer resolves to `BookingAvailability::Available()`, which equals the constructor default on both calendar types, so `sameCalendars` produces no diff and no spurious `CalendarUpdated` is emitted. The contract change is asserted by new unit tests (`it_honors_explicit_unavailable_booking_availability_on_a_{permanent,periodic}_calendar`) and reflected in the adjusted `UpdateCalendarRequestHandlerTest` / `CopyEventRequestHandlerTest` data sets.

## Ticket
https://jira.publiq.be/browse/III-7134


grubolsch:
@claude

grubolsch:
@claude

grubolsch:
One thing I am not sure of, should we make a /booking-availability endpoint work on places?
  
  Right now `PUT /places/{placeId}/booking-availability `always returns 400 (it only supports calendars with sub-events). This is not documented in apidocs, only the events variant is. With capacity now being a shared field, it could make sense to enable it for places too.
  
**Pros**
  - Consistency with events/
  - Gives a dedicated way to update just the capacity without touching the rest of the place or its calendar. (not sure the frontend needs this)
  - Future-proof if we ever expand more bookingAvailability fields for places.

**Cons**
  - For places type can only ever be Available, so the endpoint is effectively just "set capacity".
  - We'd end up with 3 ways to set the same field (PUT /places/{id}, PUT /places/{id}/calendar, and the dedicated endpoint).
  - The event endpoint earns its keep because it fans out to all sub-events, so it does a lot more then places would. So maybe overkill?

features/place/booking-availability.feature
```diff
@@ -22,4 +22,146 @@ Feature: Test place bookingAvailability property
...
+  Scenario: Update a place calendar with top-level booking availability capacity
+    Given I set the JSON request payload to:
+    """
+    {
+      "calendarType": "permanent",
+      "bookingAvailability": {
+        "type": "Available",
+        "capacity": 150
+      },
+      "openingHours": [
```
lucwollants: I guess there is no need to pass this data that doesn't get asserted?

features/place/booking-availability.feature
```diff
@@ -22,4 +22,146 @@ Feature: Test place bookingAvailability property
...
+    And I get the place at "%{placeUrl}"
+    And the JSON response at "bookingAvailability" should be:
+    """
+    {
+      "type": "Available",
+      "capacity": 150
+    }
+    """
+
+  Scenario: Update a periodic place calendar with top-level booking availability capacity
```
lucwollants, line 63: This scenario updates an existing place and doesn't do what it says. It modifies the existing permanent place from the previous scenario and changes both the calendar type and modifies the capacity of 150 to 75

tests/Model/Serializer/ValueObject/Calendar/CalendarDenormalizerTest.php
```diff
@@ -708,6 +708,130 @@ public function it_only_inherits_capacity_when_sub_event_has_no_explicit_capacit
...
+
+    /**
+     * @test
+     */
+    public function it_honors_explicit_unavailable_booking_availability_on_a_permanent_calendar(): void
+    {
+        $data = [
+            'calendarType' => 'permanent',
+            'openingHours' => [],
+            'bookingAvailability' => ['type' => 'Unavailable'],
```
lucwollants, line 805: This only makes sense in the context of an event, because in the context of a place, `Unavailable` doesn't exist.

tests/Http/Offer/UpdateCalendarRequestHandlerTest.php
```diff
@@ -405,7 +405,7 @@ public function validEventDataProvider(): array
                                 new StatusReason('Covid')
                             )
                         )
-                    )
+                    )->withBookingAvailability(BookingAvailability::Unavailable())
```
lucwollants: This almost feels like we introduced an unwanted side effect?

src/Model/Serializer/ValueObject/Calendar/CalendarDenormalizer.php
```diff
@@ -147,6 +143,8 @@ public function denormalize($data, $class, $format = null, array $context = [])
             $calendar = $calendar->withStatus($topLevelStatus);
         }
 
+        $calendar = $calendar->withBookingAvailability($topLevelBookingAvailability);
```
lucwollants: I now wonder how strictly we prevented on the HTTP validation layer a permanent or periodic event from having a top level booking availability?

lucwollants:
> One thing I am not sure of, should we make a /booking-availability endpoint work on places?
> 
> Right now `PUT /places/{placeId}/booking-availability `always returns 400 (it only supports calendars with sub-events). This is not documented in apidocs, only the events variant is. With capacity now being a shared field, it could make sense to enable it for places too.
> 
> **Pros**
> 
> * Consistency with events/
> * Gives a dedicated way to update just the capacity without touching the rest of the place or its calendar. (not sure the frontend needs this)
> * Future-proof if we ever expand more bookingAvailability fields for places.
> 
> **Cons**
> 
> * For places type can only ever be Available, so the endpoint is effectively just "set capacity".
> * We'd end up with 3 ways to set the same field (PUT /places/{id}, PUT /places/{id}/calendar, and the dedicated endpoint).
> * The event endpoint earns its keep because it fans out to all sub-events, so it does a lot more then places would. So maybe overkill?

I think being able to set the capacity with the calendar PUT endpoint is fine.

features/place/booking-availability.feature
lucwollants: The calendar type of `place.json` is permanent and not periodic. So this tests updating capacity on a permanent place and not a periodic place.

features/place/booking-availability.feature
```diff
@@ -22,4 +22,146 @@ Feature: Test place bookingAvailability property
...
+          "dayOfWeek": ["monday"],
+          "opens": "09:00",
+          "closes": "17:00"
+        }
+      ]
+    }
+    """
+    When I send a PUT request to "%{placeUrl}/calendar"
+    Then the response status should be "400"
+    And the JSON response at "type" should be "https://api.publiq.be/probs/body/invalid-data"
```
lucwollants: We also assert the content of the error message

src/Model/Serializer/ValueObject/Calendar/CalendarDenormalizer.php
grubolsch: HTTP validation doesn't prevent it at all. Both event-calendar-put.json and place-calendar-put.json declare bookingAvailability as a plain top-level property with no if-tying it to a specific calendarType. 

The only existing guard is Offer::updateBookingAvailability → CalendarTypeNotSupported ("Updating booking availability on calendar type: PERMANENT is not supported. Only single and multiple..."), and that runs only on the dedicated PUT /booking-availability endpoint — not on PUT /calendar.

That inconsitency already existed before this branch, I am not sure we can add such conditional logic in the json schema, or I could add a guard class to /calendar as well.

tests/Http/Offer/UpdateCalendarRequestHandlerTest.php
grubolsch: You are right, there should not be Unavailable in this case, the first step here is to add a rule in the json schema I think: https://github.com/cultuurnet/apidocs/pull/536

src/Model/Serializer/ValueObject/Calendar/CalendarDenormalizer.php
grubolsch: I made a PR for JSON schema changes https://github.com/cultuurnet/apidocs/pull/536

lucwollants:
As discussed and posted in Slack we need to figure out the inconsistency on top level booking availability and calendar types permanent and periodic.

src/Model/Serializer/ValueObject/Calendar/CalendarDenormalizer.php
grubolsch: Decided not to do this in the end because it would break backwards compatible.

grubolsch:
Jenkins run: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/786/

grubolsch:
See confluence page: https://confluence.publiq.be/spaces/UPW/pages/228524321/Capacity+for+periodic+permanent+activities

The current decision is to NOT support capacity on top level for permanent / periodic types.

src/Model/Serializer/ValueObject/Calendar/CalendarDenormalizer.php
```diff
@@ -147,6 +143,14 @@ public function denormalize($data, $class, $format = null, array $context = [])
             $calendar = $calendar->withStatus($topLevelStatus);
         }
 
+        if (in_array($data['calendarType'], ['single', 'multiple'], true)) {
+            $calendar = $calendar->withBookingAvailability($topLevelBookingAvailability);
+        } elseif ($topLevelBookingAvailability->getCapacity() !== null) {
+            $calendar = $calendar->withBookingAvailability(
```
lucwollants: I think you exposed the `capacity` to `permanent` and `periodic` events now and according to Confluence and your comment this is not allowed.

features/event/capacity.feature
```diff
@@ -313,3 +313,81 @@ Feature: Test capacity and remainingCapacity on sub-events
...
+          "jsonPointer": "/bookingAvailability/capacity"
+        }
+      ],
+      "status": 400,
+      "title": "Invalid body data",
+      "type": "https://api.publiq.be/probs/body/invalid-data"
+    }
+    """
+
+  Scenario: Allow a permanent calendar without capacity on an event
```
lucwollants: I think there is no added value in maintaining this test.

src/Http/Offer/EventCalendarCapacityValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,49 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Rejects a top-level bookingAvailability.capacity for events with a permanent or periodic calendar.
+ *
+ * A top-level capacity is only meaningful for events that have concrete sub-events, i.e. single or
+ * multiple calendars. Places may have a capacity on their permanent/periodic calendars, so this parser
+ * must only be wired into event request handlers, never into the shared place handling.
+ */
+final class EventCalendarCapacityValidatingRequestBodyParser implements RequestBodyParser
```
lucwollants, line 19: Not sure if this is specific to Event, I wonder if we can just name it `CapacityValidatingRequestBodyParser`?

src/Http/Offer/EventCalendarCapacityValidatingRequestBodyParser.php
```diff
@@ -0,0 +1,49 @@
...
+    public function parse(ServerRequestInterface $request): ServerRequestInterface
+    {
+        $data = $request->getParsedBody();
+
+        if (!is_object($data)) {
+            return $request;
+        }
+
+        $calendarType = $data->calendarType ?? null;
+        if ($calendarType !== 'permanent' && $calendarType !== 'periodic') {
```
lucwollants, line 30: Can we compare with the `CalendarType` enum?

lucwollants (approved):
Looks good!

lucwollants:
@grubolsch maybe I approved to early, I wonder if we also need to use the new request validator on https://publiq.stoplight.io/docs/uitdatabank/entry-api/reference/operations/update-a-event-booking-availability

PR #2270, opened 2026-06-11
Hotfix/broken upload media test

grubolsch:
### Fixed
- When pulling in the new json schema, the URL validation in upload media fails.  (If I understand correctly the w regexp mod just forces at least one char to be after the //, I can see the improvement)


lucwollants (approved):
I approved but we need to investigate why this is different from `common-string-uri-readOnly`

PR #2271, opened 2026-06-11
Remove abandoned twig/extensions package

grubolsch:
## Summary
- `twig/extensions` is abandoned and was used in a single place: the `truncate(75)` filter on `event.description` in `export.map.activities.html.twig`.
- Replaced that one call with twig's built-in `length` + `slice` filters, then removed the extension registration in `ExportServiceProvider` and dropped the package from `composer.json`/`composer.lock`.

## Test
https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/701/


grubolsch:
added some test to prove it works

lucwollants (approved):
Looks good

PR #2273, opened 2026-06-12
Calendar put should not be to clear bookinginfo

grubolsch:
### Fixed
-   A PUT /events/{eventId}/calendar request silently cleared any bookingInfo that was previously set on individual subEvents via PATCH /events/{eventId}/subEvents. Updating only the dates of an existing event should not wipe per-subEvent contact/booking data.
- Added a feature test to prove the fix works

---

Ticket: https://jira.uitdatabank.be/browse/III-7236

### DIFF
--- a/features/event/sub-event-booking-info.feature
+++ b/features/event/sub-event-booking-info.feature
@@ -187,6 +187,109 @@ Feature: Test SubEvent bookingInfo
     }
     """
 
+  Scenario: bookingInfo on subEvents is preserved when updating the calendar with new dates
+    Given I set the JSON request payload to:
+    """
+    {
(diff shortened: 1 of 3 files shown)


lucwollants (commented):
As discussed through Slack, this will change the replace behaviour as described here https://publiq.stoplight.io/docs/uitdatabank/entry-api/reference/operations/update-a-event-calendar

lucwollants:
We discussed this briefly during the Monday meeting, and @brampauwelyn will adapt the frontend code to pass a full body so we don't need to introduce a different behaviour.

@grubolsch, I think we can close this pull request?

PR #2274, opened 2026-06-16
III-7244 Extend Keycloak token generator with realm

lucwollants:
### Changed
- Extended Keycloak token generator with realm so this can also be used to create a token for UiTPAS on the `uitid` realm

---

Ticket: https://jira.uitdatabank.be/browse/III-7244

### DIFF
--- a/app/User/UserServiceProvider.php
+++ b/app/User/UserServiceProvider.php
@@ -86,6 +86,7 @@ final class UserServiceProvider extends AbstractServiceProvider
             new KeycloakManagementTokenGenerator(
                 new Client(),
                 $container->get('config')['keycloak']['domain'],
+                'master',
                 $container->get('config')['keycloak']['client_id'],
                 $container->get('config')['keycloak']['client_secret']
             ),
--- a/app/User/UserServiceProvider.php
+++ b/app/User/UserServiceProvider.php
@@ -86,7 +86,7 @@ final class UserServiceProvider extends AbstractServiceProvider
             new KeycloakManagementTokenGenerator(
                 new Client(),
                 $container->get('config')['keycloak']['domain'],
-                'master',
+                $container->get('config')['keycloak']['master_realm'],
                 $container->get('config')['keycloak']['client_id'],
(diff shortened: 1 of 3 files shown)


app/User/UserServiceProvider.php
```diff
@@ -86,6 +86,7 @@ private function getManagementTokenProvider(DefinitionContainerInterface $contai
             new KeycloakManagementTokenGenerator(
                 new Client(),
                 $container->get('config')['keycloak']['domain'],
+                'master',
```
grubolsch: There is a `App\Keycloak\Realm` class to use, that bundles all info about realms, I would reuse it  (not a fan of hardcoded strings here).
I would either pass the full Realm object, or if you think that too heavy handed, just access the internal name there, as in:
`App\Keycloak\Realm::getMasterRealm()->internalName`

grubolsch (approved):

grubolsch:
Approved but with one minor remark

app/User/UserServiceProvider.php
lucwollants: Could it be you refer to `publiq-platform` because I couldn't find a `Realm` class inside this codebase?

lucwollants: @grubolsch, fine if I add the config `keycloak.master_realm` here https://github.com/cultuurnet/appconfig/blob/main/templates/docker/uitdatabank/udb3-backend/config.php#L169

grubolsch: Good for me!

lucwollants: See https://github.com/cultuurnet/appconfig/pull/1395/commits/a7cd51db5162e37b0bc15dc05d1dd26439d4518c

lucwollants: It should have been added to the Keycloak block, moved [f2b8c28](https://github.com/cultuurnet/appconfig/pull/1395/commits/f2b8c281635dca500ce34ce7e5f9f1a6a632345c)

PR #2281, opened 2026-06-18
III-7214 improve test stability

steftrenson:
### Changed
- Refactor `waitForItemWithUrlToBeIndex` to be playhead-aware: reads the entity's current playhead from UDB3 before polling SAPI3, then waits until the indexed document's playhead meets or exceeds the expected value, preventing assertions from running against a stale index
- Replace `iWaitForTheJsonResponseAtToBe` with a simpler `iWaitUntilTheResponseContains(int $count)` that checks `totalItems` directly, aligning with the new stable-test pattern
- Migrate all SAPI3 search feature files to use entity-level `I wait for the X with url "..." to be indexed` steps placed before each search block, replacing inline response polling after the query
- Convert all post-query `And I wait for the JSON response at "totalItems" to be "N"` steps to direct `Then the JSON response at "totalItems" should be N` assertions now that waits happen upfront
- Add scenario label filter to `waitForItemWithUrlToBeIndex` polling query so concurrent test runs do not interfere with each other's indexed results
- Restructure departure-places scenarios to set departure places at event creation time, eliminating a second indexing wait round

### Removed
- Remove `iWaitForTheJsonResponseAtToHaveEntries` Behat step (replaced by entity-level wait + direct assertion)
- Remove `iWaitForResultsAtWithParameters` Behat step (replaced by entity-level wait + direct `GET` assertion)
- Remove `iWaitSeconds` utility step — all waits now use playhead-aware entity polling instead of fixed sleeps

### Added
- Add "Writing Stable Tests" section to `docs/features/search.md` documenting the required pattern: batch all entity modifications, wait for each modified entity to be indexed at its current playhead, then assert directly against SAPI3 without further polling
- Add `features/data/events/departure-places/event-with-2-departure-places.json` test fixture for scenarios that require two departure places set at creation time

---
Ticket: https://jira.uitdatabank.be/browse/III-7214


steftrenson:
@claude pls review.

lucwollants (approved):
Looks good and is a real improvement towards stability.

PR #2283, opened 2026-06-18
III-7233 Refactor childrenOnly to a separate boolean

grubolsch:
## Summary

Refactors `childrenOnly` away from the `audienceType` enum into its own dedicated boolean property on the event, mirroring the apidocs change in [cultuurnet/apidocs#534](https://github.com/cultuurnet/apidocs/pull/534).

- New boolean `childrenOnly` on the event (default `false`)
- New domain event `ChildrenOnlyUpdated`, command `UpdateChildrenOnly`, and command handler
- `departurePlaces` validation now requires `childrenOnly === true` instead of `audienceType === childrenOnly`
- `ImportEventRequestHandler` dispatches `UpdateChildrenOnly` so the field can be set during a full import
- Read-model projector emits `childrenOnly: true` on the JSON-LD body, and omits the property when `false`
- `EventDenormalizer` bridges the legacy `audience.audienceType: "childrenOnly"` payload to the new boolean so existing clients (and acceptance-test fixtures) keep working

## Out of scope (separate PRs)

- **#2286** — Adds the dedicated `PUT /events/{eventId}/children-only` HTTP endpoint (built on top of this PR). The aggregate-level `updateChildrenOnly()`/command/handler land here; the HTTP layer is in that PR.
- A follow-up PR will remove `childrenOnly` from the `AudienceType` enum and the BC bridge in `EventDenormalizer`, once historical data and clients are migrated.


grubolsch:
@claude

grubolsch:
@claude

grubolsch:
I don't think we need the bridge suggest by Claude because we are not proving long term backwards compatible support.

features/data/events/event-children-only.json
lucwollants, line 1: This combination of `childrenOnly` and `audienceType` doesn't make sense in the context of the boolean refactor.

src/Model/Serializer/Event/EventDenormalizer.php
```diff
@@ -130,10 +130,27 @@ public function denormalize($data, $class, $format = null, array $context = []):
...
+    private function denormalizeChildrenOnly(array $data, ImmutableEvent $event): ImmutableEvent
+    {
+        if (isset($data['childrenOnly'])) {
+            return $event->withChildrenOnly((bool) $data['childrenOnly']);
+        }
+
+        // Backwards compatibility: childrenOnly used to live on the audienceType
+        // enum. Treat the legacy enum value as the new boolean being true so
+        // existing import payloads keep working. Removed together with the enum value.
+        if (($data['audience']['audienceType'] ?? null) === 'childrenOnly') {
```
lucwollants: Why should we still support this? Because in the final implementation `audienceType` enum will no longer have a value `childrenOnly` and the schema will also not allow it.

That being said, we need to be careful and make sure we can still deserialise the audience type that is inside the event store. But that is done by working with a string on the events instead of the enum.

features/data/events/event-children-only.json
lucwollants: Also the name of the file should change

tests/Model/Serializer/Event/EventDenormalizerTest.php
```diff
@@ -1863,4 +1863,75 @@ public function it_should_support_denormalization_to_immutable_event(): void
...
+        /** @var ImmutableEvent $event */
+        $event = $this->denormalizer->denormalize($eventData, ImmutableEvent::class);
+
+        $this->assertTrue($event->getChildrenOnly());
+    }
+
+    /**
+     * @test
+     */
+    public function it_bridges_legacy_audience_type_children_only_to_the_new_boolean(): void
```
lucwollants: As previously stated, I think this will become irrelevant

tests/Model/Serializer/Event/EventDenormalizerTest.php
```diff
@@ -1863,4 +1863,75 @@ public function it_should_support_denormalization_to_immutable_event(): void
...
+        /** @var ImmutableEvent $event */
+        $event = $this->denormalizer->denormalize($eventData, ImmutableEvent::class);
+
+        $this->assertTrue($event->getChildrenOnly());
+    }
+
+    /**
+     * @test
+     */
+    public function it_prefers_the_top_level_children_only_field_over_the_legacy_audience_type(): void
```
lucwollants: Not an issue if we remove childrenOnly as an audience type

lucwollants (commented):
Looks good, my only main concern is that `childrenOnly` is still accepted as an `audienceType`. It's fine to tweak this in a follow-up pull request, but I would not add extra code to support both and "prefer" the boolean. That seems all code that will become obsolete.

features/data/events/event-children-only.json
grubolsch: I wanted to keep the removal of the audienceType in a seperate PR, to keep it smaller. But I will remove it in the test.

grubolsch:
I wanted to remove all references to the audianceType in a follow up PR, so the frontend has some time to do the changes.

lucwollants:
> I wanted to remove all references to the audianceType in a follow up PR, so the frontend has some time to do the changes.

My main concern was about the extra code to support both, but I see you already removed that code.

lucwollants (approved):
Looks good!

PR #2286, opened 2026-06-18
III-7233 Add PUT /events/{eventId}/children-only endpoint

grubolsch:
## Summary

Adds the dedicated HTTP endpoint for the new `childrenOnly` boolean field introduced in #2283.

- `PUT /events/{eventId}/children-only` — accepts `{"childrenOnly": true|false}`, dispatches `UpdateChildrenOnly`, returns 204
- Includes feature tests as proof

## Depends on

- #2283 (introduces the `childrenOnly` field, command, command handler, and validation switch). Merge that first; this PR is based on it.

## Jenkins run
- https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/730/


grubolsch:
@claude

features/event/children-only.feature
```diff
@@ -0,0 +1,78 @@
...
+  Scenario: Disabling childrenOnly removes it from the GET response
+    When I create a minimal permanent event and save the "url" as "eventUrl"
+    And I set the JSON request payload to:
+    """
+    {
+      "childrenOnly": true
+    }
+    """
+    And I send a PUT request to "%{eventUrl}/children-only"
+    Then the response status should be "204"
```
lucwollants, line 37: Nit: should we also test that `childrenOnly` is `true`

src/Http/LegacyPathRewriter.php
```diff
@@ -23,6 +23,7 @@ final class LegacyPathRewriter
         '/subEvents/' => 'sub-events',
         '/typicalAgeRange/' => 'typical-age-range',
         '/departurePlaces/' => 'departure-places',
+        '/childrenOnly/' => 'children-only',
```
lucwollants: I don't think we need it here because it was never exposed with the wrong format?

features/event/children-only.feature
```diff
@@ -0,0 +1,78 @@
+Feature: Test event childrenOnly property
```
lucwollants, line 1: Do we need feature tests to make sure the bool validation works?

grubolsch: Seems overkill to me? That is a basic type check enforced by the json schema?

lucwollants: If someone removes it by mistake or uses the wrong schema, we would not know it. We have plenty of these kinds of asserts inside the feature tests. The goal of the feature tests is to be 100% "reliable" and "coverable", even without running the unit tests.

lucwollants (approved):

PR #2290, opened 2026-06-19
III-7233 Drop legacy 'childrenOnly' audienceType value

grubolsch:
## Summary
- Removes `childrenOnly` from the `AudienceType` enum and every reference that still produced or asserted it (feature scenarios, JSON fixtures, `TabularDataEventFormatter`, `EventDenormalizer` legacy fallback, schema-mirroring tests).
- `AudienceUpdated::deserialize` now catches an `InvalidArgumentException` thrown by the enum constructor and falls back to `AudienceType::everyone()`, so any historical event that still carries the legacy value can be replayed without crashing the projector. I chose to not convert it to the new childrenOnly field to keep the code simple, given the old audianceType will only be used for test events.

## Ticket
https://jira.publiq.be/browse/III-7233


grubolsch:
@claude

grubolsch:
Will only be merged after the frontend is ready with its changes.

grubolsch:
I am running a jenkins test combining SAPI3 branch III-7261/refactor-children-only with ENTRY API branch III-7233/drop-legacy-childrenonly, combined the tests should be green: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/743/

grubolsch:
Succesfull Jenkins run: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/738/

features/event/departure-places.feature
```diff
@@ -118,7 +118,7 @@ Feature: Test event departure places
     Then the response status should be "400"
     And the JSON response at "schemaErrors/0/error" should be "Array should have at most 20 items, 21 found"
 
-  Scenario: Changing audienceType away from childrenOnly keeps departure places
+  Scenario: Changing audienceType keeps departure places on a childrenOnly event
```
lucwollants: I wonder if this test is still relevant because departure places are not controlled by audience type but by children only. Do we need a children-only equivalent?

features/search/children-only.feature
```diff
@@ -12,15 +12,14 @@ Feature: Test the Search API v3 boa feature
     And I am not authorized
     And I am not using an UiTID v1 API key
 
-  @external
+    # This requires feature flag enableBoaPermission=true
```
lucwollants: Is this a comment that still needs to be fixed?

features/search/children-only.feature
```diff
@@ -87,16 +86,14 @@ Feature: Test the Search API v3 boa feature
     %{otherChildrenOnlyEventId}
     """
 
-  # Todo: this test cannot work till PK-476 is done, so in the meantime it is marked as external
-  @external
   Scenario: When I have the boa scope I can search for all children only events
-    When I am authorized with an OAuth client access token for "boa_client"
+    When I am authorized with an OAuth client access token for "test_client"
```
lucwollants: I think this should stay the boa_client because that client has the new boa scope

features/search/children-only.feature
```diff
@@ -44,7 +43,7 @@ Feature: Test the Search API v3 boa feature
     %{otherChildrenOnlyEventId}
     """
 
-  @external
+
```
lucwollants: Does this test still make sense because BOA is no longer linked to audience types

features/search/children-only.feature
lucwollants, line 1: I have the feeling we are very light on tests here because when we started working on this, the new boa scope was not yet defined. But now we should try to mimic this table https://jira.publiq.be/browse/III-7121?focusedId=146391&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-146391

features/search/children-only.feature
grubolsch: Not sure what you are getting at, this is not a todo, just a warning that when it is false the test fails.
Of course, if we remove the feature flag, the comment can go as well.

lucwollants: Ok, I understand now. But that is a feature flag on the `udb3-search-service` repository, and it's controlled by the Docker appconfig. So I'm not sure it has any added value than?

grubolsch:
jenkins run: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/770/

features/search/children-only.feature
```diff
@@ -1,6 +1,22 @@
 @sapi3
 Feature: Test the Search API v3 boa feature
 
+  # Visibility matrix for childrenOnly events. Each scenario covers two rows at
```
lucwollants: Nit: I'd recommend removing this table because the scenario comments are clear. Otherwise, you'll need to maintain the comment in two places.

lucwollants (approved):

PR #2291, opened 2026-06-19
Fix feature tests for childrenOnly

grubolsch:
The change for childrenOnly broke some feature tests.
I had to mark the SAPI3 tests as external for now, because the SAPI3 implementation no longer matches with the Entry API Code. Will be fixed soon.


grubolsch:
Will run the jenkins test first, once the queue is free

lucwollants (approved):

PR #2292, opened 2026-06-19
III-7259 Fix TypeError on status update for events without typeId

grubolsch:
## Summary

- Sentry issue [UDB3-BACKEND-3G0](https://publiq-vzw.sentry.io/issues/7445394039/): `EventTypeResolver::isOvernightAllowed(): Argument #1 ($eventTermId) must be of type string, null given` when running an export that issues status updates.
- Root cause: `$this->typeId` on the `Offer` aggregate is `?string` and is only populated by `applyEventCreated` and `applyTypeUpdated`. Events sourced through `EventImportedFromUDB2` or `EventCopied` (and never re-typed) leave `typeId === null`, so any status update flowing through `Offer::updateAllStatuses` → `Event::updateCalendar` → `assertOvernightAllowed` crashed with a `TypeError`.
- Fix: widen `EventTypeResolver::isOvernightAllowed` to accept `?string`. A null typeId returns `false` (overnight not allowed); since sub-events default to `overnight = false`, legacy data passes the assertion cleanly while genuine overnight attempts on a typeless event still get blocked.


lucwollants (commented):
Although this fixes the crash, I think we also need to make sure the `typeId` is a valid value after the events `EventImportedFromUDB2` and `EventCopied`

JonasVHG:
We have a `Type onbepaald` `0.51.0.0.0`, see https://taxonomy.uitdatabank.be/terms. I kinda thought we used this `term` if a legacy term from UDB2 could not be matched. I guess I was wrong, I may be a solution to avoid nullable typeIds, but I also think we should not change the UDB2 Import code (too much).

lucwollants:
To avoid side effects, let us stick with nullable for now. We have some more context and could still enhance it later.

lucwollants (approved):

PR #2293, opened 2026-06-21
III-7260 Bootstrap getCardSystemsForEvent on UiTPAS REST API

lucwollants:
### Added
- Added `DistributionKey` with a nullable name, as this field is not required
- Added optional `distributionKeys` to `CardSystem`
- Added named constructor `fromCardSystems` to `CardSystemsJsonResponse`

### Changed
- Refactored `getEventCardSystems` to take into account `manualDistributionKeys`
- Renamed existing `GetCardSystemsFromEventRequestHandler` to `LegacyGetCardSystemsFromEventRequestHandler`
- Refactored `GetCardSystemsFromEventRequestHandler` to interact with the REST API
- Bootstrapped `GetCardSystemsFromEventRequestHandler`based on config `uitpas.rest_api.enabled`

---

Ticket: https://jira.uitdatabank.be/browse/III-7260

### DIFF
--- a/src/UiTPAS/Client/RestUiTPASClient.php
+++ b/src/UiTPAS/Client/RestUiTPASClient.php
@@ -28,6 +28,13 @@ final class RestUiTPASClient implements UiTPASClient
     {
         $cardSystems = [];
         foreach ($this->getEventCardSystemsData($eventId) as $cardSystemData) {
+            // The REST API returns the full master list (enabled and disabled), but consumers
+            // should only see the card systems that are active for the event, matching the
+            // legacy XML endpoint. The full list is kept in getEventCardSystemsData() for writes.
+            if (($cardSystemData['enabled'] ?? false) !== true) {
+                continue;
+            }
+
             $cardSystems[] = (new CardSystem(
                 new Id((string) $cardSystemData['id']),
                 $cardSystemData['name']
@@ -47,6 +54,11 @@ final class RestUiTPASClient implements UiTPASClient
     {
         $distributionKeys = [];
         foreach ($distributionKeysData as $distributionKeyData) {
+            // Same as card systems: only expose the distribution keys that are enabled for the event.
+            if (($distributionKeyData['enabled'] ?? false) !== true) {
+                continue;
+            }
+
             $distributionKeys[] = new DistributionKey(
                 new Id((string) $distributionKeyData['id']),
                 $distributionKeyData['name'] ?? null
--- a/src/UiTPAS/Client/RestUiTPASClient.php
+++ b/src/UiTPAS/Client/RestUiTPASClient.php
@@ -6,6 +6,7 @@ namespace CultuurNet\UDB3\UiTPAS\Client;
 
 use CultuurNet\UDB3\Json;
 use CultuurNet\UDB3\UiTPAS\CardSystem\CardSystem;
+use CultuurNet\UDB3\UiTPAS\CardSystem\DistributionKey;
 use CultuurNet\UDB3\UiTPAS\ValueObject\Id;
 use CultuurNet\UDB3\User\ManagementToken\ManagementTokenProvider;
 use GuzzleHttp\Psr7\Request;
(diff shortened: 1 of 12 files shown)


src/UiTPAS/Client/RestUiTPASClient.php
```diff
@@ -27,15 +28,34 @@ public function getEventCardSystems(string $eventId): array
     {
         $cardSystems = [];
         foreach ($this->getEventCardSystemsData($eventId) as $cardSystemData) {
```
grubolsch, line 30: I see that the https://docs.publiq.be/docs/uitpas/uitpas-api/reference/operations/list-event-card-systems endpoint returns a field "enabled". I wonder if we should filter on this? I am worried that consumers of this endpoint will suddenly start seeing card systems and distribution keys that were previously hidden from them, since the rest of RestUiTPASClient treats the GET response as a master list and flips enabled in place on add/delete/set. I am not sure how the the legacy XML endpoint handled this.

src/UiTPASService/Controller/Response/CardSystemsJsonResponse.php
```diff
@@ -27,6 +28,40 @@ public function __construct(array $cardSystems, int $status = 200, ?HeadersInter
...
+     * @param CardSystem[] $cardSystems
+     */
+    public static function fromCardSystems(
+        array $cardSystems,
+        int $status = 200,
+        ?HeadersInterface $headers = null
+    ): JsonResponse {
+        $data = [];
+        foreach ($cardSystems as $cardSystem) {
+            $id = (int) $cardSystem->getId()->toNative();
```
grubolsch: This cast might be dangerous, the `Id` class allows strings as ids, and if ever there are strings in these ids (because the api changes and we don't know about it for example), all these card systems are going to be cast to id 0, leading to all collapse into each other.
Do we even need the cast given that we just return a JSON response?
And if we do, maybe add a logging if such a type conversion happens?

grubolsch (approved):

grubolsch:
This was already missing before, but I looked into the coverage of feature test for the /card-systems endpoint, and saw that we only test the "failure" path of all UiTPAS endpoints.
See features/uitpas/events.feature:9

I wonder if this is because it is not possible to test with the UiTPAS api, is there a technical limitation?

No blocker here, just some food for thought.

src/UiTPAS/Client/RestUiTPASClient.php
```diff
@@ -27,15 +28,34 @@ public function getEventCardSystems(string $eventId): array
...
+    /**
+     * @param array<int, array<string, mixed>> $distributionKeysData
+     * @return DistributionKey[]
+     */
+    private function mapDistributionKeys(array $distributionKeysData): array
+    {
+        $distributionKeys = [];
+        foreach ($distributionKeysData as $distributionKeyData) {
+            $distributionKeys[] = new DistributionKey(
+                new Id((string) $distributionKeyData['id']),
```
steftrenson, line 63: Are these id and name properties always set? Though I see that in other places in this Client there aren't checks as well.

steftrenson (approved):

src/UiTPAS/Client/RestUiTPASClient.php
lucwollants: Good catch. The old XML endpoint only returned the active card systems/keys, but the new REST one returns the full list (enabled + disabled).

I added filtering for enabled in `getEventCardSystems()` and `mapDistributionKeys()`. The full list is still used internally through `getEventCardSystemsData`.

lucwollants: Done in 892c445

src/UiTPASService/Controller/Response/CardSystemsJsonResponse.php
lucwollants: Fair point!

The `id` now is defined as an integer in the REST spec and was also exposed as an integer in the XML spec. Because changing the type of `id` would be a breaking change, I added a helper method that will throw when the conversion to an integer is not possible.

lucwollants: Done in 335f178

lucwollants:
> This was already missing before, but I looked into the coverage of feature test for the /card-systems endpoint, and saw that we only test the "failure" path of all UiTPAS endpoints. See features/uitpas/events.feature:9
> 
> I wonder if this is because it is not possible to test with the UiTPAS api, is there a technical limitation?
> 
> No blocker here, just some food for thought.

We only test failures because so far we can not rely on the data of acceptance, and we don't have fixed organisers and events to 100% rely on. This fits in the project of cleaning up the ACC data and making sure there is a fixed set of test data available.

src/UiTPAS/Client/RestUiTPASClient.php
lucwollants: - `id` is required, so accessing `$distributionKeyData['id']` directly is fine.
- `name` is optional, so `??` null handles it correctly.

See spec https://docs.publiq.be/docs/uitpas/uitpas-api/reference/operations/list-event-card-systems

PR #2294, opened 2026-06-22
[Aikido] Bump mtdowling/jmespath.php to 2.9.1

grubolsch:
Upgrade jmespath.php to fix critical RCE vulnerabilities in CompilerRuntime where unescaped function names in generated PHP code allow arbitrary code execution via crafted JMESPath expressions.

<details>

<summary>✅ 2 CVEs resolved by this upgrade, including 2 critical 🚨 CVEs</summary>

<br>

This PR will resolve the following CVEs:

| Issue | Severity&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Description |
| --- | --- | --- |
| <pre>[CVE-2026-54133](https://app.aikido.dev/issues/32060743/detail?groupId=9312#CVE-2026-54133)</pre> | <pre>🚨 CRITICAL</pre> | [mtdowling/jmespath.php] A vulnerability in the JMESPath compiler allows remote code execution when processing attacker-controlled JMESPath expressions due to insufficient escaping of function names in generated PHP code. An attacker can craft malicious expressions to inject executable PHP into cache files that are subsequently loaded and executed. |
| <pre>[AIKIDO-2026-928289](https://app.aikido.dev/issues/32060743/detail?groupId=9312#AIKIDO-2026-928289)</pre> | <pre>🚨 CRITICAL</pre> | [mtdowling/jmespath.php] A crafted JMESPath expression can break out of string literals in generated PHP code, allowing arbitrary PHP code execution when using the compiler runtime. An attacker who can control the expression string can inject malicious code into the compiled-expression cache file. |

</details>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
     "league/uri-components": "^2.4",
     "mailjet/mailjet-apiv3-php": "^1.6",
     "monolog/monolog": "~1.11",
+    "mtdowling/jmespath.php": "^2.9.1",
     "opis/json-schema": "^2.1",
     "php-amqplib/php-amqplib": "^3.6.0",
     "php-http/guzzle7-adapter": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,6 @@
     "league/uri-components": "^2.4",
     "mailjet/mailjet-apiv3-php": "^1.6",
     "monolog/monolog": "~1.11",
-    "mtdowling/jmespath.php": "^2.9.1",
     "opis/json-schema": "^2.1",
     "php-amqplib/php-amqplib": "^3.6.0",
     "php-http/guzzle7-adapter": "^1.0",
@@ -87,6 +86,9 @@
     "phpunit/phpunit": "^9.5",
     "publiq/php-cs-fixer-config": "^2.0",
     "rector/rector": "^0.14.5"
+  },
+  "conflict": {
+    "mtdowling/jmespath.php": "<2.9.1"
   },
   "prefer-stable": true,
   "autoload": {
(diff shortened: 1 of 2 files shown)


lucwollants (commented):
This is not inside our composer.json file, so how are we now sure another package would not downgrade it?

grubolsch:
### Note on why this fix is shaped the way it is

`mtdowling/jmespath.php` is transitive — pulled in via `league/flysystem-aws-s3-v3` → `aws/aws-sdk-php` → `mtdowling/jmespath.php (^2.6)`. We never `use` it directly, so a normal `require` entry trips `icanhazstring/composer-unused` (rightly so). The `conflict: <2.9.1` block expresses *our* floor without claiming a direct dependency, so the unused-deps check passes and the version still cannot slip below 2.9.1 under `--prefer-lowest` or a future stricter upper bound from another package.

### Follow-up worth scheduling
Upgrading `aws/aws-sdk-php` is a separate, larger win that this PR intentionally doesn't take on:
- Currently locked at `3.337.3` (Jan 2025); aws-sdk ships several releases per week, so we're ~17 months behind.
- It's pulled in by `league/flysystem-aws-s3-v3 ^2.1` (also from 2021; flysystem v3 has been out for years).
- A real upgrade would touch both packages and needs its own compatibility/API check pass.

Doing that upgrade later would also (likely) eventually let us drop the `conflict` block here — once aws-sdk-php's own constraint moves above `^2.6`, the floor becomes implicit.

lucwollants (approved):

PR #2299, opened 2026-06-23
II-7270 Polyfil the missing top-level booking info from identical sub-events

lucwollants:
### Added
- Polyfil the missing top-level booking info from identical sub-events. I did not write acceptance tests because this is temporary code to make sure booking info is shown on UiV.

### Note 
- This still needs a reindex of the events on SAPI3

---

Ticket: https://jira.uitdatabank.be/browse/III-7270

### DIFF
--- a/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
+++ b/src/Offer/ReadModel/JSONLD/PropertyPolyfillOfferRepository.php
@@ -54,6 +54,7 @@ final class PropertyPolyfillOfferRepository extends DocumentRepositoryDecorator
                 $json = $this->polyfillStatus($json);
                 $json = $this->polyfillBookingAvailability($json);
                 $json = $this->polyfillSubEventProperties($json);
+                $json = $this->polyfillTopLevelBookingInfoFromSubEvents($json);
                 $json = $this->polyfillEmbeddedPlaceStatus($json);
                 $json = $this->polyfillEmbeddedPlaceBookingAvailability($json);
                 $json = $this->polyfillTypicalAgeRange($json);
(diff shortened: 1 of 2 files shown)


bertramakers (commented):
Technically this fixes the current issue.

However I wonder if it wouldn't be better to make this part of the actual projector and implement a real "sync" of the bookingInfo between top-level and subEvents.

For example, take the following scenario's:

1. An integrator creates an event with calendarType single, and sets the bookingInfo on the subEvent > We probably also want the same bookingInfo on the top level because not having it there is technically a breaking change for output channels (like UiV now)
2. An integrator creates an event with calendarType single, and sets the bookingInfo on the top level > While the goal is to have it on the top-level, it will become confusing if some "single" events _also_ have it on their subEvent (scenario above) and others don't (this scenario). So we probably also want to sync it from the top-level to the subEvent
3. An integrator creates an event with calendarType multiple, and sets the same bookingInfo on every subEvent instead of the top-level (due to confusion or because it's easier for them to just always set it on the subEvent instead of having the conditional logic) > It would be useful for our UiV users if this repeated bookingInfo was also set on the top-level, because top-level bookingInfo will probably just be 1 button to click vs a list of dates with a button each.
4. An integrator creates an event with calendarType multiple and top-level bookingInfo > As with scenario 2 this is okay, but might be confusing that in some cases you can have the same bookingInfo on top-level and subEvents, and in others you don't. So we could propogate the top-level bookingInfo on the subEvents to make it more predictable

lucwollants:
Given all the new insights, I will close this ticket because:
- We will use a script to retroactively fix single and multiple
- Implement a sync for single and multiple

PR #2301, opened 2026-06-24
III-7271 sync single events booking info

JonasVHG:
### Changed

- `EventLDProjector`: Cascade top-level booking info to the sub-event for single events & Propagate sub-event booking info to the top level for single events.
- `EventLDProjectorTest`: Add Unit Tests.
- `sub-event-booking-info.feature`: Add Acceptance Tests.

---

Ticket: https://jira.publiq.be/browse/III-7271


JonasVHG:
@claude, review.

lucwollants (commented):
What if we fixed this inside the JSON-LD projector?
A replay of the affected events would solve the issues on UiV?

JonasVHG:
> What if we fixed this inside the JSON-LD projector? A replay of the affected events would solve the issues on UiV?

OK, this a little bit less new code.

JonasVHG:
I will put this back in DRAFT until we decide what to do with multiples.

lucwollants:
> I will put this back in DRAFT until we decide what to do with multiples.

I think it can be a small and focused pull request on single events.
It only needs to be tweaked with URL and URL label instead of full booking info.

src/Event/ReadModel/JSONLD/EventLDProjector.php
```diff
@@ -384,9 +385,70 @@ protected function applyCalendarUpdated(AbstractCalendarUpdated $calendarUpdated
...
+     * both levels stay in sync.
+     */
+    private function syncSingleEventBookingInfoFromSubEvent(\stdClass $offerLd): \stdClass
+    {
+        if (!$this->hasSingleSubEvent($offerLd)) {
+            return $offerLd;
+        }
+
+        if (isset($offerLd->subEvent[0]->bookingInfo)) {
+            $offerLd->bookingInfo = $offerLd->subEvent[0]->bookingInfo;
```
grubolsch, line 433: I would be more precise in overwriting specific fields, we are mainly concered about URL and urlLabel.
Phone and email I would NOT do because we want to remove those fields from sublevel anyway.

src/Event/ReadModel/JSONLD/EventLDProjector.php
```diff
@@ -384,9 +385,70 @@ protected function applyCalendarUpdated(AbstractCalendarUpdated $calendarUpdated
...
+     * BookingInfoUpdated) down onto the subEvent.
+     */
+    private function syncSingleEventBookingInfoFromTopLevel(\stdClass $offerLd): \stdClass
+    {
+        if (!$this->hasSingleSubEvent($offerLd)) {
+            return $offerLd;
+        }
+
+        if (isset($offerLd->bookingInfo)) {
+            $offerLd->subEvent[0]->bookingInfo = $offerLd->bookingInfo;
```
grubolsch, line 412: See remark below

src/Event/ReadModel/JSONLD/EventLDProjector.php
lucwollants: That is what I added here https://github.com/cultuurnet/udb3-backend/pull/2301#issuecomment-4789505486 😏

JonasVHG:
@claude review.

lucwollants:
@JonasVHG and @grubolsch, we should align on the exact requirements here.

I interpret ticket https://jira.publiq.be/browse/III-7271 as:
- If booking info is missing on top level, but filled in on sub event, take the info from the sub event
- If the booking info is missing from the sub event, but present on the top level, take the info from the top level

I have the feeling that the current implementation does:
- When the top level is updated, override the sub event
- When the sub event is updated, override the top level

We should also document this inside the API docs because this can be misleading for integrators, especially if we only sync the url and url label

In the ticket, there is even this requirement about an error when they would be different
> If booking info on sub and top level are contradictory, we should give an error message.

bertramakers (commented):
I think we are missing some validation in the HTTP layer for this requirement:

> If booking info on sub and top level are contradictory, we should give an error message.

(This only matters in `POST /events` and `PUT /events/:id`. When updating the bookinginfo of a subevent specifically on a single event, it's okay to overwrite the top level bookinginfo even if it contradicts. And vice versa. So the code in the projector is okay to keep I think)

grubolsch:
Is this on hold for the meeting on wednesday?

JonasVHG:
We will no longer do this.

PR #2302, opened 2026-06-24
Script to sync toplevel bookingInfo url from subEvent when empty

grubolsch:
### Added
- Script to sync toplevel bookingInfo url from subEvent when empty, if all sub events are the same.

### DIFF
--- a/app/Console/Command/FixMultipleEventBookingInfo.php
+++ b/app/Console/Command/FixMultipleEventBookingInfo.php
@@ -1,0 +1,254 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Console\Command;
+
+use CultuurNet\UDB3\Json;
+use CultuurNet\UDB3\User\ManagementToken\ManagementTokenGenerator;
+use GuzzleHttp\Psr7\Request;
+use Psr\Http\Client\ClientInterface;
+use Psr\Log\LoggerInterface;
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Throwable;
+
+/**
+ * TEMPORARY CODE - CAN BE REMOVED AFTER RUNNING FIX
+ *
+ * Copies the booking-info URL of the sub-events up to the top level for events that have an empty top-level URL but
+ * identical sub-event URLs.
+ *
+ * For every id in the given file the command does a GET on the Entry API. An event is only updated when:
+ *  - the top-level bookingInfo URL is empty (null/absent), AND
+ *  - every sub-event has a non-empty bookingInfo URL, AND
+ *  - all those sub-event URLs are equal.
+ *
+ * When updating, the existing top-level phone and email are preserved so they can never be overwritten by accident.
+ * Rejected ids are always written to the log file together with the reason they were skipped.
+ */
+final class FixMultipleEventBookingInfo extends Command
+{
+    private ClientInterface $httpClient;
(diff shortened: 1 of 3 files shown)


app/Console/Command/FixMultipleEventBookingInfo.php
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * TEMPORARY CODE - CAN BE REMOVED AFTER RUNNING FIX
+ *
+ * Copies the booking-info URL of the sub-events up to the top level for events that have an empty top-level URL but
+ * identical sub-event URLs.
+ *
+ * For every id in the given file the command does a GET on the Entry API. An event is only updated when:
+ *  - the top-level bookingInfo URL is empty (null/absent), AND
+ *  - every sub-event has a non-empty bookingInfo URL, AND
+ *  - all those sub-event URLs are equal.
```
bertramakers, line 30: Have you considered the `urlLabel`, `availabilityStarts` and `availabilityEnds`? Those properties all directly influence how and when the `url` is displayed on UiV, widgets and other output channels. So even if you remove the `email` and `phone` in the long run, and we can ignore those for now, the other properties in `bookingInfo` still remain relevant I think?

For example, the event that was originally included in the bug report also has a `urlLabel`: https://io.uitdatabank.be/events/938fbcf1-7790-4a61-bc7d-e9e523b64bab 

In fact, I think that every event created through the FE is always created with a `urlLabel`, and the value is always 1 of 4 options (so it's not always the same). This determines the copy of the button on UiV.

The `availabilityStarts` and `availabilityEnds` are optional but can also be set in the frontend if I'm not mistaken. This determines whether the button is display and clickable on UiV, or a message is shown saying that you cannot register yet. It's possible that some events have been created with one of these properties on their subEvent(s) via the frontend.

app/Console/Command/FixMultipleEventBookingInfo.php
```diff
@@ -0,0 +1,254 @@
...
+    {
+        $dryRun = (bool) $input->getOption('dry-run');
+
+        $ids = $this->readIds((string) $input->getArgument('file'));
+        if ($ids === []) {
+            $output->writeln('<error>No event ids found in the given file.</error>');
+            return self::FAILURE;
+        }
+
+        $token = $this->tokenGenerator->newToken()->getToken();
```
JonasVHG: Since you just fetch the events from io, a token is not needed.

lucwollants: I think it's needed to do the PUT. But I wonder if we should not better use the repos and commands directly and not use the API?

lucwollants: Also, we should be careful when using a PUT method that the client has access to all events.

app/Console/Command/FixMultipleEventBookingInfo.php
lucwollants: I think the comment is not 100% in sync with the actual implementation

app/Console/Command/FixMultipleEventBookingInfo.php
```diff
@@ -0,0 +1,254 @@
...
+    /**
+     * @param array<string, mixed> $bookingInfo
+     */
+    private function updateBookingInfo(string $token, string $id, array $bookingInfo): void
+    {
+        $request = new Request(
+            'PUT',
+            $this->baseUrl . '/events/' . $id . '/booking-info/',
+            [
+                'Authorization' => 'Bearer ' . $token,
```
JonasVHG: This will update it with a management token instead of a user token, do not even know if this will work.
Can you just use a Command and dispatch it.

app/Console/Command/FixMultipleEventBookingInfo.php
grubolsch: I now take all the fields into consideration, but I never overwrite a field on toplevel if it already has a value.

JonasVHG (approved):

lucwollants (approved):
Looks good! 

I also did a quick check locally to make sure it runs and dry-run is accepted.

PR #2307, opened 2026-07-03
III-7307 Rework feature tests to make them ES8 ready

steftrenson:
### Added
- Add a `behat.php` configuration defining `default` and `sapi3` Behat suites under a `default` profile matching the current feature-ci tag set, plus an `es8` profile that overrides both suites to exclude `@negativeBoosting` scenarios
- Add `feature-ci-es8`, `feature-sapi3` and `feature-sapi3-es8` make targets to run the new suite/profile combinations
- Document the new suite-based feature test commands in `docs/commands.md`

### Changed
- Replace inline tag expressions in the `feature-ci`, `feature` and `feature-random` make targets with `--suite=default`, moving tag-filtering knowledge out of the Makefile and into `behat.php`

### Removed
- Remove the dead `skipNegativeBoostingIfConfigured` before-scenario hook from `FeatureContext.php`, which threw a `PendingException` instead of cleanly skipping `@negativeBoosting` scenarios and therefore never produced a green ES8 run; this is now handled by tag filtering in the `es8` Behat profile

---
Ticket: https://jira.uitdatabank.be/browse/III-7307

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ test-group:
 	docker compose exec php composer test -- --group=$(group)
 
 feature-init:
-	docker compose exec php composer feature -- --tags @init
+	docker compose exec php composer feature -- --suite=init
 
 feature-tag:
 	docker compose exec php composer feature -- --tags $(tag)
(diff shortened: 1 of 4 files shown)


steftrenson:
@claude pls review

grubolsch (approved):

PR #2310, opened 2026-07-06
7262 Bootstrap `AddCardSystemToEventRequestHandler`

lucwollants:
### Changed
- Bootstrapped `AddCardSystemToEventRequestHandler`

---

Ticket: https://jira.uitdatabank.be/browse/7262

### DIFF
--- a/src/UiTPASService/Controller/LegacyAddCardSystemToEventRequestHandler.php
+++ b/src/UiTPASService/Controller/LegacyAddCardSystemToEventRequestHandler.php
@@ -11,7 +11,7 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use Slim\Psr7\Response;
 
-final class AddCardSystemToEventRequestHandler implements RequestHandlerInterface
+final class LegacyAddCardSystemToEventRequestHandler implements RequestHandlerInterface
 {
     private CultureFeed_Uitpas $uitpas;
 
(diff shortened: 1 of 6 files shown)


src/UiTPASService/Controller/LegacyAddCardSystemToEventRequestHandler.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $eventId = $routeParameters->getEventId();
+        $cardSystemId = $routeParameters->get('cardSystemId');
+        $distributionKeyId = $routeParameters->has('distributionKeyId') ? $routeParameters->get('distributionKeyId') : null;
+
+        $this->uitpas->addCardSystemToEvent($eventId, $cardSystemId, $distributionKeyId);
```
grubolsch, line 30: I would not make this many variables for one call, I would inline them and give them each one line of code - but that is personal style. Good either way.

src/UiTPASService/Controller/LegacyAddCardSystemToEventRequestHandler.php
```diff
@@ -0,0 +1,34 @@
...
+
+        return new Response(200);
```
grubolsch, line 32: Same remark regarding status code 200 - technicly not accurate but maybe difficult to change without breaking frontend code.

grubolsch (approved):

src/UiTPASService/Controller/LegacyAddCardSystemToEventRequestHandler.php
lucwollants: This will be a breaking change.

src/UiTPASService/Controller/LegacyAddCardSystemToEventRequestHandler.php
lucwollants: Ternary operators can make the method call rather bulky, so that's why we used variables like in most other endpoints.

PR #2311, opened 2026-07-06
7263 Bootstrap `DeleteCardSystemFromEventRequestHandler`

lucwollants:
### Changed
- Bootstrapped `DeleteCardSystemFromEventRequestHandler`

---

Ticket: https://jira.uitdatabank.be/browse/7263

### DIFF
--- a/src/UiTPASService/Controller/LegacyDeleteCardSystemFromEventRequestHandler.php
+++ b/src/UiTPASService/Controller/LegacyDeleteCardSystemFromEventRequestHandler.php
@@ -11,7 +11,7 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use Slim\Psr7\Response;
 
-final class DeleteCardSystemFromEventRequestHandler implements RequestHandlerInterface
+final class LegacyDeleteCardSystemFromEventRequestHandler implements RequestHandlerInterface
 {
     private CultureFeed_Uitpas $uitpas;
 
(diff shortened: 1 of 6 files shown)


src/UiTPASService/Controller/LegacyDeleteCardSystemFromEventRequestHandler.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $routeParameters = new RouteParameters($request);
+        $eventId = $routeParameters->getEventId();
+        $cardSystemId = $routeParameters->get('cardSystemId');
+
+        $this->uitpas->deleteCardSystemFromEvent($eventId, $cardSystemId);
+
+        return new Response(200);
```
grubolsch, line 31: We might not be able to change it to not break the frontend, but technicly this should return a 201, no content returen

grubolsch (approved):

src/UiTPASService/Controller/LegacyDeleteCardSystemFromEventRequestHandler.php
lucwollants: That is the legacy controller and will be removed, but also in the new controller it will stay as a `200`, see https://github.com/cultuurnet/udb3-backend/pull/2311/changes#diff-937eba329c3869046748c6439559ec4638837a82058d9356df87da3f0c6c29a2L31

PR #2317, opened 2026-07-08
III-7216/overnight-filter: feature tests for hasOvernight search filter

grubolsch:
Add feature test for hasOvernight

## Related PR
https://github.com/cultuurnet/udb3-search-service/pull/495

## Ticket
https://jira.publiq.be/browse/III-7216


grubolsch:
@claude

features/search/overnight.feature
```diff
@@ -0,0 +1,225 @@
...
+    And I create a minimal place and save the "url" as "placeUrl"
+    And I wait for the place with url "%{placeUrl}" to be indexed
+
+  @testIsolation
+  Scenario: A single event with an overnight sub-event matches hasOvernight=true
+    When I create an event from "events/overnight/event-single-with-overnight.json" and save the "url" as "eventUrl"
+    And I wait for the event with url "%{eventUrl}" to be indexed
+    And I am using the Search API v3 base URL
+    When I send a GET request to "/events" with parameters:
+      | q                     | %{eventUrl} |
```
lucwollants: There is no need to add the advanced filter with `eventUrl` because you are using `testIsolation`. This is something we already discussed here https://github.com/cultuurnet/udb3-backend/pull/2314#discussion_r3543237773 and it does not match the example you added here https://github.com/cultuurnet/udb3-backend/pull/2332

features/search/overnight.feature
```diff
@@ -0,0 +1,225 @@
...
+      | hasOvernight          | false       |
+      | disableDefaultFilters | true        |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 0
+
+  @testIsolation
+  Scenario: A multiple event with overnight on only one sub-event matches hasOvernight=true
+    When I create a minimal event with overrides and save the "url" as "eventUrl"
```
lucwollants, line 32: Would it make sense to also create a JSON file for this?

lucwollants (commented):
After a review of https://github.com/cultuurnet/udb3-search-service/pull/495, I really wonder how the acceptance tests with `dateFrom` and `dateTo` could work? I think they are wrong because they miss the `hasOvernight` query parameter?

features/search/overnight.feature
grubolsch: It is a little "Fingerspitzengefühl" when to start using fixtures and when not.
I try to apply the rule to only use it if I can use it for more then 1 scenario.

In this case we have a multiple event, overnight on only one sub-event which is only used for this specific test.

grubolsch:
I added a test for the daterange window.

lucwollants:
> I added a test for the daterange window.

So this test seems irrelevant https://github.com/cultuurnet/udb3-backend/pull/2317/changes/9486ea7523ab3217b8d2b22b89e04ae77c4f7394#diff-18886cc4688bf43a4fd2a362026aff6548a63f5986e2bc2cb87e3d482751c339R171

grubolsch:
I removed one test I think was not needed, but you are pointing to the test I added last to check for date ranges? I do think this is a unique test because it tests an event with subevents that are not all equal in hasOvernight

lucwollants:
> I removed one test I think was not needed, but you are pointing to the test I added last to check for date ranges? I do think this is a unique test because it tests an event with subevents that are not all equal in hasOvernight

It seems that the line numbers got mixed up after merging. But I was referring to the feature `Scenario: Overnight does not widen the date range of an event`, which has no `hasOvernight` parameter in its query.

It seems to verify if the start and end date of an event are correctly indexed when `overnight` is set, but that has no impact on the start and end of a sub-event and is something that is unit-tested in `udb3-search-service`.

So it feels like an artificial test, but lacking the intelligence of the code base 😁

grubolsch:
removed the test

lucwollants (approved):

PR #2319, opened 2026-07-10
III-7252 Update completeness scores after BOA changes

JonasVHG:
### Added

- `CompletenessTestConfig`: Add 1 CompletenessTestConfig so we do not have to copy it al around the unit tests.
- `CompletenessFromWeightsTest`: Unit tests for various scenarios`event/place/organizer` & `childrenOnly/other`.
-  `event-everybody-complete.json`, `event-children-only-complete.json`, `place-complete.json` & `organizer-complete.json`: New data files for acceptance tests.
- `{event|place|organizer}/completeness.feature`: new Acceptance tests for `completeness`.

### Changed

- `Weights`: Add `guardNotEmpty()`.
- `CompletenessFromWeights`: Add `ItemType` in constructor, needed for difference for `capacity` between `places` & `events`. Add logic for difference between `childrenOnly` & other events.
- `EventJSONLDServiceProvider`, `PlaceJSONLDServiceProvider` & `OrganizerJSONLDServiceProvider`: Updated constructor of `CompletenessFromWeights`.
- `EventLDProjectorTest`, `PlaceLDProjectorTest` & `OfferLDProjectorTestBase`: Moved `it_projects_the_adding_of_an_image()` & `it_projects_the_editing_of_an_image()` to child classes because weights are now different between `places` & `events`. Also used `CompletenessTestConfig` instead of hard coded values. Updated completeness values.
- `OrganizerLDProjectorTest`: used `CompletenessTestConfig` instead of hard coded values. Updated completeness values
- `event/create.feature` & `place/create.feature`: Update `completeness` scores in acceptance tests.

### Related PR

- https://github.com/cultuurnet/appconfig/pull/1418

---

Ticket: https://jira.publiq.be/browse/III-7252


JonasVHG:
@claude, review.

tests/Completeness/CompletenessFromWeightsTest.php
```diff
@@ -0,0 +1,246 @@
...
+            ],
+            'calendarType' => 'single',
+            'location' => ['@id' => 'https://io.uitdatabank.dev/places/2fdb2225-58b4-4f5f-a01d-9528cd9e7f36'],
+            'typicalAgeRange' => '6-12',
+            'mediaObject' => [['@id' => 'https://io.uitdatabank.dev/images/25025aba-ca07-4d24-8b3a-c389683d05a1']],
+            'description' => ['nl' => str_repeat('a', 201)],
+            'priceInfo' => [['category' => 'base', 'price' => 10]],
+            'contactPoint' => ['phone' => ['016 10 20 30'], 'email' => [], 'url' => []],
+            'bookingInfo' => ['url' => 'https://www.publiq.be'],
+            'faq' => [['question' => 'Is er parking?', 'answer' => 'Ja']],
```
grubolsch: I think that "faq" is wrong, it needs to be "faqs", so this point will never be assigned.
I cannot find a test that actually verifies FAQ to work in the calculation, because then you would have noticed this.
https://publiq.stoplight.io/docs/uitdatabank/06d62a411a983-event-faqs

grubolsch:
Given that this config changes very rarely I wonder if we can add some feature test for this? 2 tests, one for a normal event and one for a BOA event, would be nice.

grubolsch (changes requested):

tests/Completeness/CompletenessFromWeightsTest.php
JonasVHG: The problem is that was also incorrectly `faq` instead of `faqs` in the appconfig PR, so this tests failed succesfully. All 3 are now correctly `faqs`.

grubolsch (approved):

PR #2322, opened 2026-07-16
put stoplight back on main

JonasVHG:
### Changed

- `composer`: Put `udb3-json-schemas` back on `main`

---


grubolsch (approved):

grubolsch:
Good catch!

PR #2326, opened 2026-07-17
III-7246 Add acceptance tests for the dayOfWeek event search filter

steftrenson:
### Added

- Add `features/search/recurring-on-day-of-week.feature`, a new `@sapi3` acceptance test suite covering the `recurringOnDayOfWeek` search filter for both events and places
- Cover permanent events and places, asserting that they are matched on every weekday listed in their opening hours and not on the weekdays that are absent from it
- Cover a permanent place without any opening hours, asserting that it stays searchable but is never matched by the `recurringOnDayOfWeek` filter
- Cover periodic events and places, asserting that a weekday is only matched once it reaches the minimum of four occurrences within the period and stays unmatched when it falls one occurrence short
- Cover `openingHoursClosedDays` for periodic events and places, asserting that closed days are subtracted from the occurrences of a weekday and can push a weekday below the minimum threshold while leaving another weekday above it
- Cover multiple calendar events, asserting that single day sub-events map to the weekday they take place on, that multi-day sub-events are expanded to every weekday in their range including both edges, and that a weekday covered fewer than four times is not matched
- Cover single calendar events, asserting that they remain searchable but are out of scope for the `recurringOnDayOfWeek` filter
- Assert that weekday matching is case-insensitive, so `Wednesday` returns the same result as `wednesday`
- Assert that the comma-separated url parameter syntax OR-combines the weekdays, while the `q` parameter exposes the same field for explicit AND logic
- Assert that an invalid weekday value is rejected with a 404 on both the `/events` and the `/places` endpoint, and that the rejected value is echoed back in the error detail

---
Ticket: https://jira.uitdatabank.be/browse/III-7246

### DIFF
--- a/features/search/recurring-on-day-of-week.feature
+++ b/features/search/recurring-on-day-of-week.feature
@@ -356,6 +356,37 @@ Feature: Test the recurringOnDayOfWeek event search filter
     And the JSON response at "totalItems" should be 0
 
   @testIsolation
+  Scenario: Multiple calendar event with fewer than four occurrences on a weekday is not matched by recurringOnDayOfWeek
+    # Only three sub-events, all on a Wednesday (2026-08-05, 2026-08-12 and 2026-08-19), so the
+    # weekday occurs fewer than the required minimum number of times (4).
+    When I create a minimal event with overrides and save the "url" as "eventUrl"
+    """
+    {
+      "calendarType": "multiple",
+      "startDate": "2026-08-05T10:00:00+02:00",
+      "endDate": "2026-08-19T18:00:00+02:00",
+      "subEvent": [
+        {"startDate": "2026-08-05T10:00:00+02:00", "endDate": "2026-08-05T18:00:00+02:00"},
+        {"startDate": "2026-08-12T10:00:00+02:00", "endDate": "2026-08-12T18:00:00+02:00"},
+        {"startDate": "2026-08-19T10:00:00+02:00", "endDate": "2026-08-19T18:00:00+02:00"}
+      ]
+    }
+    """
+    And I wait for the event with url "%{eventUrl}" to be indexed
+    And I am using the Search API v3 base URL
+    # Without the recurringOnDayOfWeek filter the event is returned, so it is searchable.
+    When I send a GET request to "/events" with parameters:
+      | disableDefaultFilters | true |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 1
+    # With the recurringOnDayOfWeek filter it is not returned, because Wednesday occurs fewer than four times.
+    When I send a GET request to "/events" with parameters:
+      | recurringOnDayOfWeek  | wednesday |
+      | disableDefaultFilters | true      |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 0
+
+  @testIsolation
   Scenario: Single calendar event is out of scope and not matched by recurringOnDayOfWeek
     When I create a minimal event with overrides and save the "url" as "eventUrl"
     """
--- a/features/search/recurring-on-day-of-week.feature
+++ b/features/search/recurring-on-day-of-week.feature
@@ -356,3 +356,4 @@ Feature: Test the recurringOnDayOfWeek event search filter
       | recurringOnDayOfWeek  | someday |
       | disableDefaultFilters | true    |
     Then the response status should be "404"
+    And the JSON response at "detail" should include "someday"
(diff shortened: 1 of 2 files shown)


steftrenson:
To be discussed: https://jira.publiq.be/browse/III-7246?focusedId=147784&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-147784

steftrenson:
@claude pls review

features/search/recurring-on-day-of-week.feature
```diff
@@ -0,0 +1,431 @@
...
+      | recurringOnDayOfWeek  | saturday,friday |
+      | disableDefaultFilters | true            |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 1
+    When I send a GET request to "/events" with parameters:
+      | recurringOnDayOfWeek  | tuesday,thursday |
+      | disableDefaultFilters | true             |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 0
+    # The recurringOnDayOfWeek field is also exposed in the q parameter, where AND logic can be
```
lucwollants, line 56: This is maybe worth another scenario? So one for the URL parameters and one for the advanced parameters (even the advanced one can be split into an AND and OR flavour). That makes it all easier to process for a human brain.

features/search/recurring-on-day-of-week.feature
```diff
@@ -0,0 +1,431 @@
...
+          "opens": "09:00",
+          "closes": "17:00",
+          "dayOfWeek": ["wednesday", "thursday"]
+        }
+      ]
+    }
+    """
+    And I wait for the event with url "%{eventUrl}" to be indexed
+    And I am using the Search API v3 base URL
+    # Without the recurringOnDayOfWeek filter the event is returned, so it is indexed and searchable.
```
lucwollants, line 93: This feels like an unneeded step because we have the step "wait for indexed" on line 91?

features/search/recurring-on-day-of-week.feature
```diff
@@ -0,0 +1,431 @@
...
+          "startDate": "2026-08-05",
+          "endDate": "2026-08-12"
+        }
+      ]
+    }
+    """
+    And I wait for the event with url "%{eventUrl}" to be indexed
+    And I am using the Search API v3 base URL
+    # Without the recurringOnDayOfWeek filter the event is returned, so it is indexed and searchable.
+    When I send a GET request to "/events" with parameters:
```
lucwollants, line 142: Redundant?

features/search/recurring-on-day-of-week.feature
```diff
@@ -0,0 +1,431 @@
...
+
+  @testIsolation
+  Scenario: Single calendar event is out of scope and not matched by recurringOnDayOfWeek
+    Given I create a minimal place and save the "url" as "placeUrl"
+    When I create a minimal event with overrides and save the "url" as "eventUrl"
+    """
+    {
+      "calendarType": "single",
+      "startDate": "2026-08-05T10:00:00+02:00",
+      "endDate": "2026-08-05T18:00:00+02:00",
```
lucwollants, line 232: I would also add a test for a single event that spans a longer period so that the threshold is reached

features/search/recurring-on-day-of-week.feature
```diff
@@ -0,0 +1,431 @@
...
+    }
+    """
+    And I wait for the place with url "%{placeUrl}" to be indexed
+    And I am using the Search API v3 base URL
+    When I send a GET request to "/places" with parameters:
+      | recurringOnDayOfWeek  | wednesday |
+      | disableDefaultFilters | true      |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 1
+    # Matching is case-insensitive, so the capitalised weekday returns the same result.
```
lucwollants, line 274: Now that I see this again here, I would also make the case-insensitive feature a separate scenario

features/search/recurring-on-day-of-week.feature
```diff
@@ -0,0 +1,431 @@
...
+    When I send a GET request to "/places" with parameters:
+      | recurringOnDayOfWeek  | tuesday,thursday |
+      | disableDefaultFilters | true             |
+    Then the response status should be "200"
+    And the JSON response at "totalItems" should be 0
+    # The recurringOnDayOfWeek field is also exposed in the q parameter, where AND logic can be
+    # expressed explicitly instead of the OR-combining of the url parameter.
+    # Tuesday is not part of the opening hours, so requiring both weekdays excludes the place.
+    When I send a GET request to "/places" with parameters:
+      | q                     | recurringOnDayOfWeek:(monday AND tuesday) |
```
lucwollants, line 301: I think we miss a scenario the OR filter in the advanced queries

lucwollants (commented):
Already approved, remarks can be handled in follow up pull request if needed

lucwollants (approved):

PR #2335, opened 2026-08-03
III-7347 Drop matchingBirthdateRanges assertions from the search features

lucwollants:
## Summary
- Removes the two `matchingBirthdateRanges` assertion blocks, since the field is removed from the Search API in cultuurnet/udb3-search-service#516.
- Both scenarios keep their `totalItems` and "response should include" assertions, so birthdate range filtering itself stays covered — only the assertion on the removed response field goes.

## Related
- https://github.com/cultuurnet/udb3-search-service/pull/516

Needs to land together with the search-service PR: these scenarios run against a deployed Search API, so they only pass once the field is gone there.

## Ticket
https://jira.publiq.be/browse/III-7347

### DIFF
--- a/features/search/advanced-queries-offers.feature
+++ b/features/search/advanced-queries-offers.feature
@@ -255,25 +255,6 @@ Feature: Test the Search API v3 advanced queries on offers
     """
     %{eventId2022}
     """
-    And the JSON response at "matchingBirthdateRanges" should be:
-    """
-    [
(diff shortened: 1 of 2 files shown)


lucwollants:
Superseded by #2336 — reopened under the correct ticket (III-7190). Closed automatically by a branch rename, not intentionally.

PR #2338, opened 2026-08-06
Add search feature tests for filtering across age and birthdate

lucwollants:
### Added

Feature tests for the search filtering we get once both ranges are indexed and the query-time conversion is gone.

- A one open bound birthdate range. Only a from, or only a to, matches an event whose birthdate range overlaps it.
- A `typicalAgeRange` advanced query matches an event that was entered with a birthdate range.

Ticket: https://jira.publiq.be/browse/III-7367

### DIFF
--- a/features/search/advanced-queries-offers.feature
+++ b/features/search/advanced-queries-offers.feature
@@ -300,7 +300,9 @@ Feature: Test the Search API v3 advanced queries on offers
     And I wait for the event with url "/events/%{eventId2020}" to be indexed
     And I am using the Search API v3 base URL
     When I send a GET request to "/events" with parameters:
-      | q | id:%{eventId2020} AND typicalAgeRange:[0 TO 200] |
+      | q             | typicalAgeRange:[0 TO 5] |
+      | availableFrom | *                        |
+      | availableTo   | *                        |
     Then the JSON response at "totalItems" should be 1
     And the JSON response should include:
     """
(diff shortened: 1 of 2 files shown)


steftrenson (approved):

JonasVHG (approved):

features/search/advanced-queries-offers.feature
```diff
@@ -293,6 +293,22 @@ Feature: Test the Search API v3 advanced queries on offers
...
     """
 
+  @testIsolation
+  Scenario: Search by typical age range using an advanced query also matches events entered with a birthdate range
+    When I create a minimal place and save the "url" as "placeUrl"
+    And I create an event from "events/event-with-birthdate-range-in-2020.json" and save the "id" as "eventId2020"
+    And I wait for the event with url "/events/%{eventId2020}" to be indexed
+    And I am using the Search API v3 base URL
+    When I send a GET request to "/events" with parameters:
+      | q             | typicalAgeRange:[0 TO 5] |
```
bertramakers, line 303: Nit: It might be useful to also test that you don't get this event back when searching for example on `typicalAgeRange:[6 TO 10]`, to make sure the converted "to" is exactly 5

bertramakers (approved):

features/search/advanced-queries-offers.feature
lucwollants: Added safeguard in e1b9cc3 but the start date of the event is 2021 so the lower boundary is 2 years

PR #2344, opened 2026-08-13
Cover the conflicting bookingAvailability validation

lucwollants:
### Context

`POST /events` accepts a subEvent carrying both a `type` and a `remainingCapacity`; `PATCH /events/{eventId}/sub-events` rejects the very same object. Two schemas describe it and they disagree, and neither side had a test. Tests only, no behaviour change.

### Added

- `tests/Http/Event/ImportEventRequestHandlerTest.php`: The combination is accepted on import (201).
- `tests/Http/Event/UpdateSubEventsRequestHandlerTest.php`: The same object is rejected on the sub events PATCH, with `SchemaError('/0/bookingAvailability', 'The data should match exactly one schema')`.

### Why they disagree

| path | schema | rule |
| --- | --- | --- |
| `POST`/`PUT /events` | `event.json` → `event-subEvent.json` → `event-subEvent-bookingAvailability.json` | `anyOf` — at least one, both allowed. Its description says "or both". |
| `PATCH .../sub-events` | `event-subEvent-patch.json`, inlined | `oneOf` — exactly one, both rejected |

`event.json` also describes the GET representation, where a subEvent always carries a `type` derived from its `remainingCapacity`, so it cannot reject the combination without invalidating its own responses. The patch schema is write-only and can.

See https://jira.publiq.be/browse/III-7385

---


lucwollants:
No longer relevant after the removal of `capacity` and `remainingCapacity`

